PR #20, opened 2021-08-18
[Zhu Tianqi] ip

Tianqi-Zhu:
* text-based
* easy to learn
* <del>FAST</del> ***SUPER*** **FAST** to use

All you need to do is
1. download it from [here](https://github.com/Tianqi-Zhu/ip/releases/tag/v0.1)
2. double-click it.
3. add your tasks.
3. let it manage your tasks for you :laughing:

And it is __FREE__!

Features:
 - [x] Managing tasks
 - [x] Managing deadlines
 - [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,119 @@
...
+    
+    /**
+     * Parse a string of time typed by user of format hhmm.
+     * 
+     * @param timeString String of hhmm format to parse.
+     * @return time parsed
+     * @throws DukeExcpetion if the format is not correct
+     */
+    private static LocalTime parseTime(String timeString) throws DukeExcpetion {
+        String expanded = timeString.substring(0, 2) + ":" + timeString.substring(2) + ":00";
```
janjanchen: Perhaps the variable name `expanded` could be changed to make it more descriptive? (e.g. expandedTimeString)

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public Event(String name, LocalDate startDate, LocalTime startTime) {
+        super(name);
+        this.startDate = startDate;
+        this.startTime = startTime;
+        this.isTimeString = false; 
+    }
+
+    private String getTime() {
```
janjanchen: Perhaps `getTime()` could be named as `getDateAndTime()` since the method will always return date, while time is only included if the user added more details to `Event`

sw-boi (commented):
LGTM! Aside from some nitpicks mentioned, your code looks good.

src/main/java/duke/DukeExcpetion.java
```diff
@@ -0,0 +1,14 @@
+package duke;
+
+/**
+ * Exception when users type invalid input.
+ */
+public class DukeExcpetion extends Exception {
```
DonnyS57: This is something minor but Is it better if the Class is named DukeException? I think it will increase the readability of the code!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,119 @@
...
+    
+    /**
+     * Parse a line of texts typed by user. 
+     * 
+     * @param newInput User's new line of input. 
+     * @param taskList Current taskList.
+     * @return Commands to execute or tasks to put into list.
+     * @throws DukeExcpetion If the input is valid.
+     */
+    public static Executable parse(String newInput, TaskList taskList) throws DukeExcpetion {
```
DonnyS57: Perhaps the if-else conditions can be separated into different methods? I think this will greatly increase the readability of your code as the parse method is currently quite verbose.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,130 @@
...
+        Storage.save(this);
+        return out;
+    }
+    
+    /**
+     * Delete a task by its index. Print the task name and total task amount in the list after adding. For Gui.
+     * 
+     * @param index Index of the task to be deleted. 
+     */
+    public String deteteTask(int index) {
```
DonnyS57: Perhaps the name can be changed to deleteTask? (Another very minor change that can enhance readability).

DonnyS57 (commented):
Overall, I think most of the code adheres to the Coding Standard aside from the minor problems that I highlighted. Keep up the good work!

PR #22, opened 2021-08-19
[Keith Tan Shi Jun] iP

KT27Learn:
# DukePro

DukePro frees your mind of having to remember things you need to do. It's :

* text-based
* easy to learn
* ~~FAST~~ SUPER FAST to use

All you need to do is:

1. download it from [here](https://github.com/KT27Learn/ip)
2. double-click it.
3. add your tasks.
5. And it is FREE!

Features:

- [x]  Managing tasks
- [ ] Reminders (coming soon)
 
## For Developers:
If you want to clone the repository and run it on your CLI instead of using the jar file, you can follow the instructions here.

From the root of the repository:

    cd src/main/java

Compile all the classes in the package:

    javac -classpath . duke/Duke.java
  

Finally, run the program:

    java duke/Duke


src/main/java/duke/Main.java
```diff
@@ -0,0 +1,32 @@
+package duke;
+
+import java.io.IOException;
+
+import duke.ui.MainWindow;
```
yeo-yiheng, line 5: Perhaps you could separate the different import statements based on their packages?

src/main/java/duke/Main.java
```diff
@@ -0,0 +1,32 @@
...
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private Duke duke = new Duke("/Users/keithtan/Desktop/NUS/CS2103 IP/ip/data/duke.txt");
+
+    @Override
+    public void start(Stage stage) {
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
```
yeo-yiheng, line 22: I may be wrong on this but the class `FXMLLoader` seems like it should be named `FxmlLoader`?

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Description:
+ * Class that encapsulates the delete task command inputted by the user
+ *
+ * @author Keith Tan
+ */
+public class DeleteCommand extends Command {
+    public static final String COMMAND_WORD = "delete";
+    private Tasklist list;
+    private final int toDelete;
```
yeo-yiheng: Perhaps your final variable should be in all uppercase?

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Description:
+ * Class that encapsulates the mark task command inputted by the user
+ *
+ * @author Keith Tan
+ */
+public class MarkCommand extends Command {
+
+    public static final String COMMAND_WORD = "add";
+    private final int toMark;
```
yeo-yiheng: Perhaps the final variable count be all uppercase? I have noticed this in some other areas as well.

src/main/java/duke/ui/MainWindow.java
```diff
@@ -0,0 +1,74 @@
+package duke.ui;
+
+import duke.Duke;
```
yeo-yiheng, line 3: Perhaps the import statements can be grouped by their packages? I noticed this in other areas as well.

yeo-yiheng (commented):
LGTM. I like that your program is very neatly arranged and written. Just some really minor styling to change and we're good to go!

src/main/java/duke/Main.java
IceWizard4902: Better separation between the import should be nice

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @author Keith Tan
+ */
+public abstract class Command {
+
+    /**
+     * Returns a boolean of whether current command is the 'bye' command
+     *
+     * @return boolean returns boolean of whether current command is 'bye' command
+     */
```
IceWizard4902, line 21: Perhaps the name of the method can be changed to isByeCommand to improve readability

src/main/java/duke/command/DeleteCommand.java
IceWizard4902: Should the toDelete be in all caps (TODELETE)?

src/main/java/duke/command/MarkCommand.java
IceWizard4902: I think you may forget to change the final variable to all uppercase.

IceWizard4902 (approved):
Everything looks great, and I think that it is an overall job well done! There are no major issues, and the code base was easy to understand. There are just some small pointers that we have both brought up.

src/main/java/duke/Main.java
KT27Learn: Hi I have checked and believed that FXMLLoader is the correct one in this case as upon checking the class name is indeed declared as FXMLLoader in the javafx package

src/main/java/duke/command/DeleteCommand.java
KT27Learn: Thanks for raising this up, I believe I had missed out on this. I will rectify it in a future commit.

src/main/java/duke/command/MarkCommand.java
KT27Learn: Thanks for raising this up, I believe I had missed out on this. I will rectify it in a future commit.

PR #25, opened 2021-08-19
[Liu Chen En] iP

twothicc:
# Duke
> "MMMM so good and tasty" - Merril Howard Kalin ([source](https://www.youtube.com/watch?v=-FrpuPLYnvY))

If you aren't using Duke, then why are you even **ALIVE**? It's,

- text-based.
- easy-to-learn.
- ~~FAST~~ ULTRA FAST to use.

See how an average non-Duke user compares to the average Duke user.
Average non-Duke user vs Average Duke enjoyer

Stoked? Wait no more!
All you need to do is,

1. Download it from [here](https://github.com/twothicc/ip/releases/tag/v0.1).
2. Double-click it.
3. Add your tasks.
4. Let *ol* Duke take care of managing your tasks. :100:

**OH** did I mention that Duke is absolutely **FREE**?

That's right. We don't need you to sell your *soul* for our application. No one should have to do so to manage their tasks.

Features:

- [x] Storing general tasks and tasks with a certain date.
- [x] Marking tasks as done.
- [x] Deleting tasks.
- [x] Finding tasks by date and keyword.
- [ ] Reminders  (coming soon)

* * *

```Java
        Duke duke = new Duke(SAVE_FILENAME);
        duke.run();
}
```


src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,73 @@
...
+        DialogBox db = new DialogBox("Duke: Hello! I'm Duke\nWhat can I do for you?", img);
+        db.flip();
+        return db;
+    }
+
+    public static DialogBox getDukeBye(Image img) {
+        DialogBox db = new DialogBox("Duke: Bye. Hope to see you again soon!", img);
+        db.flip();
+        return db;
+    }
```
leezhixuan: Did you forget to add javadocs for these public methods?

src/main/java/duke/DukeLauncher.java
```diff
@@ -0,0 +1,9 @@
+package duke;
+
+import javafx.application.Application;
+
+public class DukeLauncher {
+    public static void main(String[] args) {
+        Application.launch(Main.class, args);
+    }
```
leezhixuan: Are javadocs missing for this class and its public methods?

leezhixuan (commented):
I think your work is incredibly neat and well documented. Not so sure if it matters but I thought that it might be good to add in some descriptions for your test cases. Good work!

src/main/java/duke/DialogBox.java
twothicc: Thank you for pointing the missing javadocs out to me!
I'll add it in in the next push.

src/main/java/duke/DukeLauncher.java
twothicc: It was a very recent addition. I'll add javadocs for the class and its public methods in the next push.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,73 @@
...
+
+import duke.Ui.Commands;
+import duke.command.Command;
+
+/**
+ * Handles initialization of storage and tasks and running of Duke chatbot.
+ */
+public class Duke {
+    /** Variables related to initializing Duke */
+    private static final String SAVE_FILENAME = "dukeSave.txt";
```
eezj35, line 12: I like that you cached the filename as a constant variable!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,73 @@
...
+     * Storage will read tasks from specified file to TaskList.
+     * If there are errors reading file, TaskList will be empty by default.
+     *
+     * @param fileName Filename that Storage will save tasks to and read tasks from.
+     */
+    public Duke(String fileName) {
+        this.storage = new Storage(fileName);
+        this.ui = new Ui();
+
+        // Read tasks from save file.
```
eezj35, line 37: I like the comments, very helpful in making the code more readable 👍

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,274 @@
...
+        } else if (original.length() > expected) {
+            throw new DukeException(Ui.exceptionInvalidDateTimeFormat());
+        }
+
+        return output;
+    }
+
+    /**
+     * Parses a LocalDate into a string format "d MMMM yyyy" to be printed to the user.
+     * For instance, the LocalDate equivalent of 2020-01-01 will be parsed into "1 January 2020".
```
eezj35, line 115: I like the detailed javadocs summary!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,274 @@
...
+     * @param dateString Formatted string to be parsed.
+     * @return LocalDate equivalent of the date represented by dateString.
+     * @throws DukeException If dateString is of an invalid date format.
+     * @throws DukeException If dateString represents a date later than "+999999999-12-31" or
+     *                       earlier than "-999999999-01-01".
+     */
+    public static LocalDate toLocalDate(String dateString) throws DukeException {
+        String[] split = dateString.split("[/\\s]");
+
+        // Check if split has 3 elements. If not, this is already an invalid date format.
```
eezj35, line 142: Perhaps a comment here might no be as necessary? I think the code is simple enough!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,274 @@
...
+     * @return Command that can be executed to perform a set of actions.
+     */
+    public static Command parse(String userInput) {
+        // Only exactly "list" will be accepted for ListCommand.
+        if (userInput.equals(Commands.LIST.getCommand())) {
+            // Print tasks
+            return new ListCommand(userInput);
+        } else {
+            if (userInput.startsWith(Commands.DONE.getCommand())) {
+                // Mark task as done.
```
eezj35: Perhaps a comment might not be so necessary here? I think the way you have written your code is quite understandeable without comments!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,274 @@
...
+                return new MarkCommand(userInput);
+            } else if (userInput.startsWith(Commands.DELETE.getCommand())) {
+                // Delete a task.
```
eezj35: Same comment here!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,274 @@
...
+     * @return An int representing an index.
+     * @throws DukeException If user input for index cannot be parsed into Integer.
+     */
+    public static int parseUserNumInput(String userInput, Commands command) throws DukeException {
+        // Parses integer in user input. Invalid user input could throw NumberFormatException.
+        try {
+            // Add 1 as user's number input is separated from command by 1 space.
+            return Integer.parseInt(userInput.substring(command.getLength() + 1));
+        } catch (NumberFormatException nfe) {
+            // Invalid user input cannot be parsed into Integer.
```
eezj35: Noticed the same issue about comments here! Maybe you could check your other methods to see if the comments are necessary?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Deals with loading tasks from save file and saving tasks in the file.
+ */
+public class Storage implements Storable {
+
+    /** Tasks saved are separated by SAVE_SEPARATOR */
+    private static final String SAVE_SEPARATOR = " ~ ";
+
+    /** Filename of file to be saved to */
+    private final String saveFileName;
```
eezj35, line 38: This variable sounds a little like a verb, maybe you could change it to be a noun?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,243 @@
...
+     *
+     * @param save The read string.
+     * @return Task corresponding to the read string.
+     * @throws DukeException if string cannot be parsed to date.
+     * @throws DukeException if string cannot be parsed to integer.
+     * @throws DukeException if string does not have sufficient parts for the task it represents.
+     */
+    private static Task parseSaveFormat(String save) throws DukeException {
+        // Split save string by the save separator.
+        String[] saveSplit = save.split(SAVE_SEPARATOR);
```
eezj35: Perhaps this variable name could be made into a noun too?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Removes Task from TaskList.
+     *
+     * @param idx Index of Task to be removed.
+     * @return The removed Task.
+     */
+    public Task remove(int idx) {
+        Task tempTask = this.tasks.get(idx);
```
eezj35, line 69: Maybe could rename this variable to taskToBeRemoved?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return this.descriptor;
+        }
+    }
+
+    /** Enums for Duke chatbot commands */
+    public enum Commands {
```
eezj35: I like your use of Enums here

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,364 @@
...
+    public String getWelcome() {
+        return "Hello! I'm Duke\nWhat can I do for you?";
+    }
+
+    /**
+     * Returns the standard goodbye response.
+     *
+     * @return Standard goodbye response.
+     */
+    public String getGoodbye() {
```
eezj35: Maybe for this method (and similar methods in this class), you could rename the methods to this format: 
"<command>Message". So for this case, maybe something like "goodbyeMessage". Feel free to use or not use my suggesstion though!

eezj35 (commented):
Hi Great job on your code! I think your methods are all of suitable length and your code is generally quite readable, although i think some comments can be removed if the line of code it is referencing is easy to understand. Taking away some comments can reduce the clutter too! I like your OOP approach as well. Overall really good job!

src/main/java/duke/Parser.java
twothicc: I agree that the code is simple, but I would still prefer to keep the comment for clarity purposes.

src/main/java/duke/Parser.java
twothicc: Thanks for pointing this out! This comment does seem to be stating the obvious and could instead contribute to clutter in the code. I will remove it in the next push.

src/main/java/duke/TaskList.java
twothicc: Thanks for the suggestion! That is indeed a more descriptive name. I will make the changes in the next push.

src/main/java/duke/Ui.java
twothicc: Thanks for the suggestion! I agree that this will provide a better description for the function's purpose and I will make relevant changes in the next push.

src/main/java/duke/Storage.java
twothicc: Thanks for the suggestion! However since "separator" is a noun and that it sufficiently describes the purpose of this constant, I do not feel that it is necessary to make changes to this name.

src/main/java/duke/Storage.java
twothicc: Thanks for the suggestion! I agree that it would be more appropriate to use a noun as the name of this variable and I will make the changes in the next push.

PR #32, opened 2021-08-20
[Darren Hoon] iP

darrenhoon:
### DukePro

“Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes/))

- text-based
- easy to learn
- ~FAST~ **SUPER FAST** to use

All you need to do is,

1. download it from [here](https://nus-cs2103-ay2122s1.github.io/website/schedule/week4/project.html)
2. double click it.
3. add your tasks,
4. let it manage your tasks for you!

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```
public class Main {
        Application.launch(MainWindow.class, args);
    }
}
```


src/main/java/exception/DukeException.java
```diff
@@ -0,0 +1,19 @@
...
+ * The DukeException Exception is the superclass of all exceptions to be thrown that are specific to the features
+ * of the Duke Application.
+ * @author  Hoon Darren
+ * @version 1.0
+ * @since   2021-08-21
+ */
+public class DukeException extends Exception {
+
+    /**
+     * Superclass of all exceptions that could be thrown by the Duke app.
```
g4ryy: Perhaps it would be better if the description describes the constructor rather than the class, since you have already done that above. I also noticed this issue in InvalidCommandException.java and TaskNumberException.java.

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * The Parser middleware provides dependency injection for classes that require common methods.
+ * @author  Hoon Darren
+ * @version 1.0
+ * @since   2021-08-21
+ */
+public class Parser {
+
+    /**
+     * Method to split a string into an array, removing spaces.
```
g4ryy: Perhaps it would be better if the sentence could start in the form "Splits a string into ...".

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * Creates a directory if not present in local storage.
+     */
+    public void createDirectory() {
+
+        try {
+            if(!Files.exists(this.dir)) {
+                messages.displayText("You currently do not have a directory to save the database. " +
```
g4ryy: Perhaps breaking before the '+' operator would be better.

src/main/java/ui/FurtherAction.java
```diff
@@ -0,0 +1,5 @@
+package ui;
+
```
g4ryy: Perhaps it would be good if you could include a Javadoc comment to describe what this enum is for.

src/main/java/ui/Ui.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+public class Ui {
+    private static final String WELCOME_MESSAGE = "Hello! I'm Duke\n" + "What can I do for you?";
+    private static final String BORDERLINE = "\n____________________________________________________________\n";
+    private static final String GOODBYE_MESSAGE = "Bye. Hope to see you again soon!";
+
+
+    /**
+     * Displays message input with horizontal borderlines on top and bottom.
+     * @param input String message to be displayed to user
```
g4ryy, line 21: Perhaps you could end the description with a full stop.

g4ryy (commented):
Overall, I found your code really easy to read! There are only a few minor coding standard violations that needs to be fixed, but other than that everything looks good! 🙂

src/main/java/parser/Parser.java
darrenhoon: thanks for the feedback Gary!

PR #33, opened 2021-08-20
[Yeo Jun Wei] iP

JunWei3112:
Duke helps to keep track of the things you need to do, so that you don't have to! It's
* Text-based
* Easy to learn
* ~~Fast~~ SUPER FAST to use!!
* Most importantly, ***FREE!!!!***

To use Duke, you will have to do the following steps:
1. Download the file from [here](https://github.com/JunWei3112/ip/tree/master).
2. Double-click it.
3. Add your tasks.
4. Let Duke manage your tasks! :smile:

Features:
- [x] Managing tasks, deadlines and events
- [x] Loads and saves tasks in your hard disk
- [ ] Find deadlines/events occurring on a specific date (coming soon!)

If you are a Java programmer, you can use it to practice Java too. (wow!)
Here's the main method: 
```java
public class Main  {
    }
}
```


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                    date = LocalDate.parse(timeFrame);
+                    return new AddCommand(taskType, task, date);
+                } else {
+                    task = taskDescription;
+                    return new AddCommand(taskType, task);
+                }
+            } else {
+                throw new UnrecognisedCommandException();
+            }
```
lwj1711: The logic in this block looks rather convoluted. Perhaps the use of switch statements can help?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * This Deadline class implements the characteristics of a task
+ * that needs to be done before a specific date/time.
+ */
+public class Deadline extends Task {
+
+    protected LocalDate by;
```
lwj1711: I thought this could be a little ambiguous. Perhaps a name such as 'timeBy' might be better.

lwj1711 (commented):
Looks great generally, only extremely minor nits on some parts. 🦄

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,78 @@
...
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.TaskList;
+import duke.task.Todo;
+
+/**
+ * This AddCommand class represents a command to add a task to the task list.
+ */
+public class AddCommand extends Command {
```
Zhou-Jiahao-1998, line 15: AddCommand sounds like an action. Will CommandAdder sound more like a class? Same goes for other subclasses of Command.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,50 @@
...
+        // Removes a Task from the TaskList instance
+        Task removedTask = tasks.deleteTask(taskId);
+
+        // Displays a message indicating the task has been successfully deleted from the list
+        String message = "Got it! I've removed this task:\n" + "  " + removedTask + "\n";
+        if (tasks.getNumberOfTasks() <= 1) {
+            message += "Now you have " + tasks.getNumberOfTasks() + " task in the list.";
+        } else {
+            message += "Now you have " + tasks.getNumberOfTasks() + " tasks in the list.";
+        }
```
zhoukerrr: Just a nitpick, "Now you have " can probably be abstracted out and placed together in line 38.

zhoukerrr (commented):
Overall very well done. Classes are abstracted and categorised clearly. Exceptions are well handled as well.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,43 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * This Event class implements the characteristics of a task
+ * that is held on a specific date.
+ */
+public class Event extends Task {
+
+    protected LocalDate at;
```
zhoukerrr: Maybe can be more specific? Similar point as Deadline.

src/main/java/duke/parser/Parser.java
JunWei3112: Good point, I will make changes to this block of code! Thanks for the catch, and also for the suggestion! 😄

src/main/java/duke/task/Deadline.java
JunWei3112: Good catch, I will change this variable name as per your suggestion! Thanks! 👍🏽

src/main/java/duke/command/DeleteCommand.java
JunWei3112: That's a good catch, I will make the following changes! Thanks! 😁

src/main/java/duke/task/Event.java
JunWei3112: Good point, I will change this variable name too! Thanks! 😄

PR #38, opened 2021-08-20
[Arya Nagatama Giat] iP

aryagiat:
# Duke x BORAT

> Very nice! - Borat Sagdiyev

### Forget your tasks no more with _**Duke x BORAT**_. Now you can:
- [ ] Add *todo* tasks
- [ ] Add *deadlines*
- [ ] Add *events*
- [ ] Mark them as done
- [ ] Delete tasks
- [ ] Find tasks

### Why _**Duke x BORAT**_?
- **SIMPLE!** 💎
- **Incredibly** Fast! ⚡
- Data ~can be leaked~ is **SAFE** with you and only you 😏
- **EASY** ✏

### How to use?
1. Download it from [here](https://github.com/aryagiat/ip/releases/tag/v0.1) (**Definitely** safe)
2. Open it
3. do `help` to see all commands
4. add your first todo task by doing:
```
todo watch Borat! Cultural Learnings of America for Make Benefit Glorious Nation of Kazakhstan
```
<hr>

### Help develop:
Here is the entry point: go to `Duke.java`:

``` java
/**
 * Entry point of the Duke program.
 * @param args Command line arguments.
 */
    new Duke("./data", "duke.txt").run();
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,153 @@
...
+    private final Ui ui;
+
+    /** Hard disk storage */
+    private final Storage storage;
+
+    /** User input parser */
+    private final Parser parser;
+
+    /** The list of tasks */
+    private TaskList list;
```
beefham: Perhaps "taskList" or "tasks" may be a better name for this variable.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,40 @@
+package duke;
+
+/**
+ * Exceptions for the Duke app.
+ */
+public class DukeException extends Exception{
```
beefham: Very small nitpick, there is a missing space between "Exception" and "{"

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,316 @@
...
+            if (inputs.length < 2) {
+                throw new DukeException(DukeException.Errors.MISSING_DESCRIPTION.toString()
+                        + " (example: 'todo watch Borat')");
+            }
+            String description = combineStringArray(inputs, 1, inputs.length);
+            return new String[] {commandStr, description};
+
+        case DEADLINE:
+            if (inputs.length < 2) {
+                throw new DukeException(DukeException.Errors.MISSING_DESCRIPTION.toString() + " (example: 'deadline watch Borat /by 2021-08-21 18:00')");
```
beefham: I like the use of an example string to guide the user to a correct format for each command, but perhaps this line should be broken up because it is too long (over 120 characters)

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,33 @@
+package duke;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A Task of type Deadline.
+ */
+public class Deadline extends Task{
```
beefham: Minor nitpick, there is a missing space between "Task" and "{"

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,316 @@
...
+                System.out.println("Time is not a number");
+                throw new DukeException(DukeException.Errors.INVALID_TIME.toString());
+            }
+        }
+        if (splitTime.length == 2) {
+            // in the form of [hh, mm]
+            if (
+                    (splitTime[0].length() == 2 || splitTime[0].length() == 1) &&
+                            (splitTime[1].length() == 2)
+            ) {
```
stephanie-csy: Not sure if this is the correct way to split an `if` condition into different lines. Maybe 

```javascript
 if ((splitTime[0].length() == 2 || splitTime[0].length() == 1) 
    && (splitTime[1].length() == 2)) {
```
would be better? Applies to several instances of `if` below as well.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,46 @@
...
+        return isDone
+                ? "X"
+                : " ";
+    }
+
+    /**
+     * Returns the string representation of a task.
+     * @return The description of the task.
+     */
+    public String toString() {
```
stephanie-csy, line 53: Nitpicking but can consider adding an `@Override` above the method as it overrides the toString() method in Object.

stephanie-csy (commented):
Very neat and easily readable code, follows coding standards closely generally except for some tiny spacing/indentation errors here and there. Methods and variables are named appropriately. Really like that you handled all kinds of exceptions and gave the user information on what exactly the error is and what they should do instead. Also really like the Borat references hahaha good job!

PR #39, opened 2021-08-20
[Siddanth Umralkar] iP

wowsiddanth:
# Petal 🌸
> *In joy or sadness, flowers are our constant friends.*
> **Okakura Kakuzo**

It's okay. Petal is here. Drop your worries and sink into the petals of relief; Petal is here for you.

Petal has the following abetments: 
* Text-based 
* Simple commands
* Task tracking

## Installation

A logical thing you might do is wonder—*how do I install Petal*?

There are two methods:
      
### 1. Run it in an IDE
1. Go to this [link](https://github.com/wowsiddanth/ip).
2. Download the folder.
3. Open it in IntelliJ or your favorite IDE.
4. Then, under the Petal class. add the following main method:
```java
        Petal p = new Petal();
        p.run(); 
 }
```
### 2. Run the .jar file
1. Go to this [link](https://github.com/wowsiddanth/ip).
2. Under the __Releases__ tab, open __A-Jar__ and download the .jar file.
3. Download the folder.
4. Open command prompt from that folder, or your favorite IDE.
5. Then, type this into the command prompt: `java -jar iP.jar`

## Features
Petal can:
- [X] Keep track of tasks
- [X] Keep track of the dates of tasks, including deadlines and events.
- [ ] Set reminders (coming soon)


src/main/java/petal/Petal.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public Petal() {
+        ui = new Ui(this);
+        taskList = new TaskList(ui);
+        storage = new Storage(taskList, ui);
+        parser = new Parser();
+    }
+
+    /**
+     * Flips the boolean isBye and terminates the process
```
vimuthm: It might be better to comment, "Sets the boolean isBye to false..." to remove possible confusion arising from using the word "flip"

src/main/java/petal/components/Calendar.java
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @param timeable The Timeable to be added
+     */
+    public void addToCalendar(Timeable timeable) {
+        LocalDate date = timeable.getDate();
+        Optional<ArrayList<Timeable>> current = Optional.ofNullable(dateAndTimeable.get(date));
+        if (current.isPresent()) {
+            current.get().add(timeable);
+        } else {
+            ArrayList<Timeable> firstList = new ArrayList<>();
```
vimuthm: Coding Standard mentions:

> Plural form should be used on names representing a collection of objects.

`firstList` might not strictly adhere to this but I'm not sure if it's an outright violation. What do you think?

src/main/java/petal/components/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Parses the text from Tasks.txt in tasks and retrieves it as tasks
+     *
+     * @return True if tasks were retrieved, false if no tasks (new user) or exception occurred
+     */
+    public boolean retrieveTasks() {
+        try {
+            File tasks = new File(filePath);
+            Scanner scanner = new Scanner(tasks);
+            ArrayList<Task> toBeAdded = new ArrayList<>();
```
vimuthm: Same issue as previous collections variable naming

src/main/java/petal/task/ToDo.java
```diff
@@ -0,0 +1,38 @@
+package petal.task;
+
+/**
+ * The To.Do class that is a subclass of Task
+ * (Used a . in between to prevent IntelliJ from highlighting it)
```
vimuthm, line 5: Nice :) I faced the same issue!

vimuthm (commented):
Overall really good use of OOP and nice job on the Calendar feature!

One thing I noticed with the design implementation was a persisting check on todo/deadline/event. Within `Parser.java`, all 3 cases are grouped to return a `TaskCommand` which then calls upon `handleTask` in `TaskList.java` where you parse it to different cases again.

Personally, I feel maybe having a subclass of `TaskCommand` to handle these separately might be better. After which, handling the command input parsing seems more aligned to the duties of the corresponding command class and not within `TaskList` (if command format changes for example, current implementation means you need to modify the `TaskList` methods and not the `TaskCommand` class)

Any thoughts on this?

PR #48, opened 2021-08-20
[Linglong] iP

huisthat:
# DukePro 

* text-based
* easy to learn
* ~~FAST~~ *SUPER FAST* to use

All you need to do is,
1. download it from [here](https://github.com/huisthat/ip).
1. double-click it. 
1. add your tasks.
1. let it manage your tasks for you. :smile: 

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---

```
public static void main(String[] args) throws IOException {
    new Du("data/du.txt").run();
}
```


src/main/java/Du/Parser.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param tasks
+     */
+    public Parser(TaskList tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * takes in commands from user and parse them
+     */
+    public void parse() {
```
pbthang, line 23: I think the responsibility of Parser should only be parsing 1 command at a time, and the loop should be in Duke class. What do you think?

src/main/java/Du/Event.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * toString method
+     * @return String for printing
+     */
+    @Override
+    public String toString() {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd yyyy, HHmm");
+        return "[E][" + this.getStatus() + "] " + this.name
+                + "(at: " + time.getMonth() + " " + time.format(formatter) + ")";
```
pbthang, line 50: In this case, should we use `String.format()` because I find it less complicated than string concatenation?

data/du.txt
```diff
@@ -0,0 +1,5 @@
+T , 1 , sth
+D , 0 , sth else  , 2021-08-22 18:30
+T , 0 , yay
+T , 0 , hi
+D , 0 ,  , 2021-08-16 16:30
```
pbthang, line 5: Should we put the data file in `.gitignore`, because I think each person needs their own data?

src/main/java/Du/Task.java
```diff
@@ -0,0 +1,70 @@
+package Du;
+
+/**
+ * Class for each individual task and a list of the tasks
+ */
+public class Task {
```
pbthang, line 6:
```suggestion
public abstract class Task {
```

pbthang (commented):
Sorry wrong PR :))

src/main/java/Du/TaskList.java
```diff
@@ -80,4 +80,29 @@ public void find_finished_task(int order) {
...
+        for (Task t : this.getList_of_tasks()) {
+            if (t.getName().contains(substring)) {
+                have_substring.add(t);
+            }
+        }
+        return have_substring;
+    }
+
+    public void print(ArrayList<Task> arraylist) {
+        System.out.println("____________________________________________________________\n" +
```
sj1999-BIT: This is what I did at first as well. It turns out that a more recommended approach is to set a global final String variable that contains this line. It could potential make the code cleaner.

src/main/java/Du/Deadline.java
```diff
@@ -19,14 +19,14 @@ public Deadline(String name, boolean done, TaskList tasklist, LocalDateTime time
     }
 
     /**
-     * record the Deadline in a certain format to save to the file
+     * records the Deadline in a certain format to save to the file
```
sj1999-BIT: One of the JavaDocs requirements is that every function descriptor should start with a capital letter .It is recommended to follow the general guidelines

src/main/java/Du/Du.java
```diff
@@ -25,7 +28,7 @@ public Du(String filepath) {
 
 
     /**
-     * run Du itself
```
sj1999-BIT: Generally speaking, function descriptors should be more specific and more detailed, unless the function is a very generic one such as getters and setters. I would recommend adding in more details such as what purpose exactly do run do, perhaps 2 to 3 lines will do?

sj1999-BIT:
Overall the logic is pretty well done. The presentation could be improved, but then again my own work is also not that presentable XP. Jiayou!

src/main/java/Du/TaskList.java
```diff
@@ -0,0 +1,108 @@
+package Du;
+
+import java.util.ArrayList;
+
+public class TaskList {
+
+    protected static ArrayList<Task> list_of_tasks = new ArrayList<>();
```
luluyousef, line 7: can consider changing the name to listOfTasks or taskList instead to follow the camelCase coding standard! :)

luluyousef:
Overall, your code looks pretty good to me. If you want you can consider changing your variables to camelCase format, but good job so far!! 💯

PR #50, opened 2021-08-20
[Lee Wei, David] iP

itsyme:
Duke manages your tasks, it
* Keeps Track of Tasks (and their deadlines! :grin:)
*  Is text based
*  **SUPER** fast to use

All you have to do is,
1. Download it from [here](https://github.com/itsyme/ip/releases/tag/A-Jar)
2. double-click it
3. add your tasks
4. let it manage your tasks for you!

And it's *FREE*!

Features:
- [x] Managing Tasks
- [x] Managing Deadlines

If you're a Java programmer, you can use it to practice Java too! Here is the `main` method:
```java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Constructs an instance of <code>Duke</code>.
+     * @param filePath File path to file storing tasks
+     */
+    public Duke(String filePath) {
+        this.storage = new Storage(filePath);
+        this.ui = new Ui();
+        try {
+            this.tasks = new TaskList(this.storage.loadSave());
+        } catch (Exception e) {
```
ryanongra, line 20: Might be better to catch specific exceptions like IOException or FileNotFoundException?

src/test/java/duke/DeadlineTest.java
```diff
@@ -0,0 +1,33 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+import java.time.LocalDate;
+import static org.junit.jupiter.api.Assertions.*;
```
ryanongra: Maybe should just import the assertEquals method?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param date Date of task
+     */
+    Event(String body, boolean done, LocalDate date) {
+        super(body, done);
+        this.date = date;
+    }
+
+    /**
+     * Returns date of <code>Event</code>
+     * @return Date
```
ryanongra, line 35: Very small thing but maybe can be a little more specific in the javadocs (e.g. returns the data that event is happening)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,48 @@
...
+        this.ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = this.ui.getNextLine();
+                Command command = Parser.parse(fullCommand);
+                command.execute(this.tasks, this.ui, this.storage);
+                this.storage.updateFile(this.tasks.getTasks());
+                isExit = command.isBye();
+            } catch (Exception e) {
```
ryanongra: Same as before, might be better to list the specific exceptions

ryanongra (commented):
LGTM! Just a few minor things :)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Constructs an instance of <code>Duke</code>.
+     * @param filePath File path to file storing tasks
+     */
+    public Duke(String filePath) {
+        this.storage = new Storage(filePath);
+        this.ui = new Ui();
+        try {
+            this.tasks = new TaskList(this.storage.loadSave());
+        } catch (Exception e) {
```
yewmeng28: Perhaps it would be better to specify the exceptions you would like to catch such as `IOException`.

src/main/java/duke/Event.java
yewmeng28, line 35: Perhaps it would be better to be more descriptive like the `date` param in the `Event` constructor JavaDocs.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,62 @@
...
+    boolean isDone() {
+        return this.isDone;
+    }
+
+    /**
+     * Returns <code>Task</code> with status set to done
+     * @return Done <code>Task</code>
+     */
+    Task setDone() {
+        return new Task(this.body, true);
```
yewmeng28, line 50: Perhaps it would be better to simply set the `isDone` of the current `Task` to `true`.

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,42 @@
...
+        super(body, done);
+    }
+
+    /**
+     * Returns <code>Todo</code> with done set to true
+     * @return Done <code>Todo</code>
+     */
+    @Override
+    Task setDone() {
+        return new Todo(this.getBody(), true);
```
yewmeng28, line 31: It would be easier to use the parent class's `setDone()` method (after changing it as per the previous comment) so you would not need to override this method in all the different task types.

yewmeng28 (commented):
LGTM! Only some minor tweaks to make your code look neater and make life easier for you :)

PR #55, opened 2021-08-20
[Wong Pei Xian] iP

trxe:
# DukeAgain

```
Hello from
 ____        _        
|  _ \ _   _| | _____ 
| | | | | | | |/ / _ \
| |_| | |_| |   <  __/
|____/ \__,_|_|\_\___|
```

If you are cool with a ~buggy~ TUI project with weird input formatting, use DukeAgain as a:

1. terminal-based task manager
2. easy to use
3. simple task management options

Download **Duke 0.1** from _**Releases**_ and run `java -jar duke.jar` on your terminal in the same directory as the download. Make sure you've installed Java Runtime Environment on your device (available [here](https://www.oracle.com/java/technologies/javase-jre8-downloads.html)).

-----

## Quick start documentation
The following are the recognized formats for 3 different types of tasks (items in round brackets are optional:

- `todo [description]`
- `deadline [description] /by [(yyyy-MM-dd) (HH:mm)] (/ details)`
- `event [description] /by [(yyyy-MM-dd) (HH:mm)] ~ [(yyyy-MM-dd) (HH:mm)] (/ details)`
- `find [keyword(s)]`
- ℹ️ `help` to learn other commands! 

-----

## Features to be added

- [ ] Sort by time
- [ ] More flexible formatting
- [ ] Reminders

> _"yea boi"_
> ~ internet guy


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+            return new Record("Noted. I've removed this task:\n\t   " + t + sizeMsg());
+        } catch (NumberFormatException | IndexOutOfBoundsException e) {
+            if (db.size() == 0) {
+                throw new DukeException("You have no tasks.");
+            }
+            throw new DukeException(String.format("Enter a valid index (from 1 to %d).", db.size()));
+        }
+    }
+
+    private Record todo(String args) throws DukeException {
```
wonyeji, line 169: Perhaps you could use a verb like `addTodo` instead of `todo` for the method name? Same for the other addTask methods below!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private Record find(String raw) throws DukeException {
+        TaskList filtered = db.find(raw);
+        if (filtered.size() == 0) {
+            return new Record("No matching tasks found.");
+        }
+        return new Record("Here are the matching tasks in your list:\n " + filtered);
+    }
+}
```
wonyeji, line 262: There seems to be a missing newline at the end of the file! I noticed it for a few other files as well.

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,74 @@
...
+            details = rawArgs;
+        }
+    }
+
+    /**
+     * Returns this deadline formatted as a database entry.
+     *
+     * @return String representing this deadline in database format.
+     */
+    public String toDB() {
```
wonyeji:
```suggestion
    public String toDatabase() {
```

src/main/java/duke/utils/Record.java
```diff
@@ -0,0 +1,62 @@
+package duke.utils;
+
+public class Record {
+    private final String message;
+    private final boolean terminate;
+
+    /**
+     * Creates a new record.
+     *
+     * @param message
```
wonyeji: I think these may be missing their descriptions?

src/main/java/duke/utils/Record.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return message.
+     */
+    public String msg() {
+        return message;
+    }
+
+    /**
+     * Getter for terminate.
+     *
+     * @return
```
wonyeji: This is missing the description as well!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+        cmds.put("help", this::help);
+        cmds.put("clear", this::clear);
+        cmds.put("find", this::find);
+    }
+
+    /**
+     * Takes in a command, and if command is found, passes to it the remaining arguments.
+     *
+     * @param input
+     * @return
```
wonyeji, line 62: Missing return description!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+                return cmds.get(cmd).apply(raw);
+            } else {
+                throw new DukeException("I'm sorry, but I don't know what that means :-(");
+            }
+        } catch (NoSuchElementException e) {
+            throw new DukeException("Please enter a command.");
+        }
+    }
+
+    private String sizeMsg() {
```
wonyeji: Perhaps you could use a verb for this method name? Something like `getTaskSize()`

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+import duke.utils.DukeException;
+import duke.utils.Record;
+
+import java.util.HashMap;
+import java.util.NoSuchElementException;
+import java.util.Scanner;
+
+public class Parser {
+
+    private final TaskList db;
```
wonyeji: Perhaps a more meaningful name like `dataBase` could be used? 👍

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public Task parse(String line) throws DukeException {
+        try {
+            String[] args = line.split("( \\| )", 3);
+            if (args.length < 3) {
+                return null;
+            }
+            boolean done = Integer.parseInt(args[1]) == 1;
+            switch (args[0]) {
+            case "T":
```
wonyeji: Rather than just the char here, perhaps you could add a char constant to the different Task classes to enhance readability? Also to make it easier to edit in the future when task types with overlapping char representations are created! :~)

wonyeji (commented):
Overall, brilliant job! Just a few nits to fix, otherwise your code is amazing! 💯

src/main/java/duke/Parser.java
trxe: Newlines are added by the static method `Ui.reply`, thanks though!

src/main/java/duke/Parser.java
trxe: Personally I think its clearer in my code setup that the command the user issues matches the name of private function in the `Parser` class that corresponds to it, so I don't thnk I will be changing these names

src/main/java/duke/tasks/Deadline.java
trxe: Changed to `toDatabaseFormat`, thanks!

src/main/java/duke/Storage.java
trxe: Thank you for the suggestion, will implement

PR #59, opened 2021-08-20
[Bryann Yeap] iP

BryannYeap:
# DUKE (INDIVIDUAL PROJECT)

## Gary Bot 🤖 

> “There is satisfaction," he said to Dalinar, "in creating a list of things you can actually accomplish, 
> then removing them one at a time.” - *Brandon Sanderson, author of Oathbringer* ([source](https://www.goodreads.com/quotes/10211374-there-is-satisfaction-he-said-to-dalinar-in-creating-a))

Gary makes sure you keep tabs on all the things that you have to do! Gary would be **very** disappointed if you forget to finish your tasks... ~~You don't want to know the things he'd do...~~

Here are some attributes of Gary:
* Text-based CLI
* Intuitive and simple
* Responsive

To get Gary, just follow these simple steps:
1. Download from [here](https://github.com/BryannYeap/ip/releases).
2. Open your command line and navigate to the directory containing the `iP.jar` file.
3. Type this command: `java -jar {filename}.jar` where `{filename}` is the name of the file (duh!)
4. Add your tasks and let Gary take care of the rest!  

Features:
- [x] Basic functionality of adding, removing and marking tasks as done.
- [x] Querying / Searching for tasks.
- [x] Built as a downloadable jar
- [x] GUI (coming soon)
- [ ] Adding times to Deadline tasks

It is a (hopefully) elegant program as well! Here is the `main` method:
```java
        new Duke().run();
    }
```


src/test/java/duke/task/TodoTest.java
```diff
@@ -0,0 +1,19 @@
+package duke.task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TodoTest {
+     @Test
+     public void toString_descriptionOfTodo1_todoStringReturned() {
```
HolmesJJ: Perhaps it would be better to merge these two test cases in the same method and add message parameter in `assertEquals` method to distinguish them rather than name it as `toString_descriptionOfTodo1` and `toString_descriptionOfTodo2` since I think they are testing the same method in the same scenario and returning the same type of result, and the naming of 1 and 2 looks confusing? 🤔

src/test/java/duke/task/TaskTest.java
```diff
@@ -0,0 +1,38 @@
+package duke.task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+import org.junit.jupiter.api.Test;
+
+public class TaskTest {
```
HolmesJJ, line 8: I like your very complete test cases! 😄

src/test/java/duke/task/EventTest.java
```diff
@@ -0,0 +1,41 @@
+package duke.task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+import duke.exception.InvalidParamException;
+
+public class EventTest {
```
HolmesJJ, line 10: I like your very complete test cases +1! 😄

src/test/java/duke/task/DeadlineTest.java
```diff
@@ -0,0 +1,50 @@
+package duke.task;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+import duke.exception.InvalidParamException;
+
+public class DeadlineTest {
```
HolmesJJ, line 10: I like your very complete test cases +1! 😄

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        Ui.printDoubleDivider();
+        return true;
+    }
+
+    private void findTask(String input) {
+        try {
+            System.out.println("Output: These tasks have descriptions that contain the phrase '"
+                    + input.substring(5) + "'!\n");
```
HolmesJJ: Should this magic number be moved to top by using named constant? 🤔

src/main/java/duke/task/Storage.java
```diff
@@ -0,0 +1,122 @@
...
+                String description = task.getDescription();
+
+                // Add a marker to tell if its done or not
+                toBeWritten.append(task.getIsDone() ? "1" : "0").append(SEPARATOR);
+
+                //Add the task type and description
+                toBeWritten.append(initialOfTask).append(SEPARATOR).append(description);
+
+                switch (initialOfTask) {
+                case 'T':
```
HolmesJJ:
```
private static final char TASK_TYPE = 'T';
```
Should this magic character be moved to top by using named constant or use enum? 🤔
And I noticed the same issue in several other places below.

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * @param input Input string from the user.
+     */
+    private void vetoTask(String input) {
+        Task newTask = null;
+
+        try {
+            if (!input.startsWith("todo") && !input.startsWith("deadline") && !input.startsWith("event")) {
+                throw new duke.exception.InvalidInputException();
+            } else {
```
HolmesJJ: Perhaps the `else` can be removed ? 🤔

HolmesJJ (commented):
LGTM! I like your super clear code and all the code structure is very complete. Only a few code suggestions and magic numbers and characters should be moved to top by using named constant or use enum. Well done! 😃

src/main/java/duke/main/Parser.java
Agentum07: I could not understand the purpose of the '5'. I suggest you declare it as a constant so that it's easier to understand the context?

src/main/java/duke/main/Parser.java
Agentum07, line 161: I think the `else` is not needed.

Agentum07: I think you can remove this `else`. If the `if` condition is satisfied, the program will directly go to the catch and skip all the code

Agentum07 (commented):
Overall, pretty well written code. Quite a good job with the javadocs, very easy to understand! Good job!

src/test/java/duke/task/TodoTest.java
BryannYeap: Done!

src/test/java/duke/task/TaskTest.java
BryannYeap: Thanks!

src/main/java/duke/main/Parser.java
BryannYeap: Done!

src/main/java/duke/task/Storage.java
BryannYeap: Hey I tried taking this into consideration and changing it to an enum but apparently it doesn't work. Switch block cases require 'constant expressions', and even when I set the enum field of `initial` to be `final`, it did not work. So I guess I'll just leave it as it is. Thanks for the suggestion!

src/main/java/duke/main/Parser.java
BryannYeap: Done!

src/main/java/duke/main/Parser.java
BryannYeap: Done!

src/main/java/duke/main/Parser.java
BryannYeap: Done!

MGRL2201 (commented):
Very clean, well written, and well-documented code!

Chesterwongz (commented):
LGTM! Very nice and clean code.

PR #64, opened 2021-08-21
[Amadeus] iP

Amadeus-Winarto:
# BobCat
> “This is very fun!” – Me 

## What is BobCat
BobCat is: 
- text-based
- easy to learn
-  ~~FAST~~ SUPER FAST (hopefully)

## How do I use this thing? 
To use it, follow these steps: 
1. Download it from [here](https://www.google.com)
2. Double-click it the jar file
3. Add your tasks.
4. And be **happy!** 😉

## Features
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)
- [ ] If you Java programmer, you can use it to practice Java too. Here's the `main` method:

```java
public class Main {
    }
}
```


src/main/java/BobCat/executor/parser/CommandParser.java
```diff
@@ -0,0 +1,9 @@
+package BobCat.executor.parser;
+
+/**
+ * Interface that all command parsers in BobCat.BobCat must obey. A <code>CommandParser</code> checks if the given command
+ * and commandArgs are appropriate for each other. Specifically, it checks for the number of given arguments.
+ */
+public interface CommandParser {
+    String[] parse(String command, String[] commandArgs);
+}
```
clyveycui: Interesting design decision here. May I know why did you decide to create so many different CommandParsers?

src/main/java/BobCat/model/TaskList.java
```diff
@@ -0,0 +1,53 @@
+package BobCat.model;
+
+import BobCat.exception.InvalidOpsException;
+import BobCat.model.task.Task;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private final ArrayList<Task> storage;
```
clyveycui: The variable name here should be in uppercase I believe.

clyveycui (commented):
Awesome job! Just one quick fix and should be good to go.

src/main/java/BobCat/executor/parser/CommandParser.java
Amadeus-Winarto: Hi. The decision was made due to the intention to categorise a *primitive* Command into 3 kinds: a BasicCommand which does read-only operations, a TaskCreation command which creates a new Task, as well as a TaskMutationCommand, which mutates a given Task.

The reason behind this hierarchy is to force an implementer to create commands with a clear, small job i.e. either read-only, create or mutate.  Other kinds of commands can then be created as a composition of these primitive commands, thus forcing reuse of the already written primitive commands. 

The myriad of parsers are written to implement this idea. This leads to BasicCommandParser, CreationCommandParser, and MutationCommandParser. An alternative approach (that I can think of) is to have a `Command` class, with methods `execute` and `parse`. Not too sure which one is better, if there's even a difference, so currently sticking to this approach. 

QueryParser is a general parser that calls the appropriate CommandParser. It does not implement CommandParser. The same holds for StorageParser, which parses the storage file. This is done to allow flexibility in how storage files are written.

src/main/java/bobcat/executor/parser/QueryParser.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A parser object which process queries. Queries are grouped based on its behavior on <code>TaskList</code> i.e.
+ * creation, mutation, or read-only. QueryParser will call the appropriate <code>CommandParser</code> based on these
+ * sets of commands.
+ */
+public class QueryParser {
+    private static final Set<String> taskCreation = Set.of("todo", "event", "deadline");
+    private static final Set<String> taskMarking = Set.of("done", "delete");
+    private static final Set<String> basicCommand = Set.of("list", "find", "bye");
```
SpdPnd98: Should these be all caps and snake case?

SpdPnd98 (commented):
LGTM! Code looks well organised!

src/main/java/bobcat/exception/BobCatException.java
```diff
@@ -0,0 +1,7 @@
+package bobcat.exception;
+
+public class BobCatException extends RuntimeException {
```
CrownKira, line 7: Perhaps you would like to include a header comment for this class since it is public?

CrownKira: Noticed this in several other places too.

src/main/java/bobcat/executor/ExecutionUnit.java
```diff
@@ -0,0 +1,92 @@
...
+        storeExecutor.initStorage(storagePath, taskList);
+    }
+
+    public void clearStorage() {
+        taskList = new TaskList();
+    }
+
+    /**
+     * Executes the given query. ExecutionUnit will pass the query to the appropriate parser and execute the
+     * arguments according to the command recognised by the parser.
```
CrownKira, line 44: Should you have a separation line between subject and param?

src/main/java/bobcat/executor/ExecutionUnit.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param query The given query to the chat app
+     * @return Array of strings representing information received after query is executed
+     * @throws IOException may be thrown if given storagePath to ExecutionUnit does not exist
+     */
+    public String[] executeCommand(String query) throws IOException {
```
CrownKira: Perhaps you can refactor this method into multiple parts? eg. extract the logic for each action (list, bye, etc)? I would recommend to keep a method to <30 LoC

src/main/java/bobcat/executor/parser/QueryParser.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final BasicCommandParser basicCommandParser = new BasicCommandParser();
+    private final CreationCommandParser taskCreationParser = new CreationCommandParser();
+    private final MutationCommandParser taskMarkingParser = new MutationCommandParser();
+
+    /**
+     * Process user query and call the appropriate <code>CommandParser</code>.
```
CrownKira, line 24: Processes*

CrownKira: calls

src/main/java/bobcat/model/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+
+import bobcat.exception.InvalidOpsException;
+import bobcat.model.task.Task;
+
+public class TaskList {
+    private final ArrayList<Task> storage;
+    private int index;
+
+    /**
+     * Constructor for a task list
```
CrownKira, line 35: Instead of starting with a noun, perhaps you could phrase it this way: Constructs an empty task list

src/main/java/bobcat/view/DialogBox.java
```diff
@@ -0,0 +1,75 @@
...
+    public DialogBox(Label l, ImageView iv) {
+        l.setWrapText(true);
+        iv.setFitWidth(100.0);
+        iv.setFitHeight(100.0);
+
+        this.setAlignment(Pos.TOP_RIGHT);
+        this.getChildren().addAll(l, iv);
+    }
+
+    //    private void flip() {
```
CrownKira: Perhaps you would like to 'garbage collect' the code if deprecated ?

CrownKira (commented):
LGTM!

src/main/java/BobCat/model/TaskList.java
Amadeus-Winarto: Can I clarify that the use of uppercase here is due to the "final" keyword? Was under the impression that constants in Java refer to static final fields, not just final

PR #66, opened 2021-08-21
[Geng Ning] iP

zognin:
Duke is a task manager. It allows you to

- Add, delete and mark tasks as done
- View your tasks and find them easily

**All you need to do is**

1. Download the JAR release (first release [here](https://github.com/zognin/ip/releases/tag/A-Jar))
2. Double click it
3. Enter commands to add your tasks

If you are a programmer, you can clone the repo to practice Java
Run the main method in `duke.Launcher` to start the program
```java
public class Launcher {
    public static void main(String ...args) {
    }
}
```

**Features** :monocle_face:
- [ ] Manage tasks
- [ ] GUI


chunweii:
LGTM! I can see that you really put in a lot of effort into the iP! Code style is consistent and follows the CS2103/T coding standards. 💯

src/main/java/controller/DialogBox.java
```diff
@@ -0,0 +1,68 @@
...
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+import javafx.scene.shape.Ellipse;
+
+public class DialogBox extends HBox {
```
chunweii, line 22: Consider adding some JavaDoc here to describe what DialogBox does

src/main/java/tasklist/Task.java
```diff
@@ -0,0 +1,127 @@
...
+        this.description = description;
+        this.isDone = isDone;
+    }
+
+    /**
+     * Creates a specific `entity.list.DukeTask` based on the prefix in the user input.
+     * Supported types of tasks include todo, deadline and event.
+     *
+     * @param description the full user input that may or may not be a valid type of task
+     * @return A new `entity.list.DukeTask`
```
chunweii: Should `A` be `a` instead?

src/main/java/tasklist/TaskList.java
```diff
@@ -0,0 +1,139 @@
...
+        } catch (IOException e) {
+            throw new ErrorAccessingFile("delete");
+        }
+    }
+
+    /**
+     * Marks task in the list as done, finding the task by its task number.
+     *
+     * @param taskNumber Task number of the task in the list, starting from 1.
+     * @return Task.
```
chunweii: To be consistent with your other documentations, consider making this into code using markdown ``

src/main/java/type/DukeCommandTypeEnum.java
```diff
@@ -0,0 +1,22 @@
+package type;
+
+public enum DukeCommandTypeEnum {
```
chunweii: Do add a JavaDoc here as well

JasonC01 (commented):
Looks really good to me. OOP was well used and I couldn't find any possible flaws. Learnt a lot looking at your code.

src/main/java/type/DukeCommandTypeEnum.java
zognin: Alright I've addressed all your comments, thank you!

src/main/java/duke/Main.java
```diff
@@ -0,0 +1,30 @@
...
+import controller.MainWindow;
+import javafx.application.Application;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+
+/**
+ * Encapsulates main class that starts the application with duke and the user interface.
+ */
+public class Main extends Application {
+    private final Duke duke = new Duke();
```
bingcheng45: Maybe for duke can be in all caps to represent a constant.
" Constant names must be all uppercase using underscore to separate words."

src/main/java/tasklist/DeadlineTask.java
```diff
@@ -0,0 +1,114 @@
...
+        this.deadline = deadline;
+    }
+
+    /**
+     * Processes the input string to create a deadline task with an action and deadline.
+     *
+     * @param description Input task string.
+     * @return App representation of a task containing an action description and deadline information.
+     */
+    public static DeadlineTask createTask (String description) throws
```
bingcheng45: I think the space between createTask (String ...) can be removed.

src/main/java/message/Message.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Instantiates a Message holding texts to be output to the user.
+     *
+     * @param texts Texts are vararg of strings.
+     */
+    public Message(String ...texts) {
+        assert texts.length > 0 : "A message should have least one string";
+
+        this.texts = texts;
+    }
```
bingcheng45, line 21: I think it would provide clarity to add in throw AssertionError to indicate there could be potential errors to catch.

bingcheng45 (commented):
LGTM, I love how you utilise what was taught in 2030 concepts such as monads, good use of inheritance, streams etc. Your java conventions are on point as well.

src/main/java/message/Message.java
zognin: Would it be necessary to throw exceptions for assertions? Because assertions are used to catch programmer mistakes and are separate from exceptions. I would think that if an AssertionError happens, it should be picked up during development, be fixed and not allowed to pass to production? Additionally, they can be disabled?

src/main/java/duke/Main.java
zognin: Thanks for point this out!

PR #68, opened 2021-08-21
[Lee Zheng Han] iP

solarrabbit99:
# Zheng Han's Markdown
### Unordered Lists
- This is an un-ordered list
- Second element of un-ordered list
### Ordered Lists
1. First element
2. Second
## Code block
```java
  System.out.println("Follow me on instagram at zhenghanlee");
}
```
## Task list
- [x] CS2103T
- [ ] CS3230
- [ ] CS2101
- [ ] GEQ1000
- [ ] ST2334


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,157 @@
...
+        sendButton.setOnMouseClicked((event) -> {
+            handleUserInput();
+        });
+
+        userInput.setOnAction((event) -> {
+            handleUserInput();
+        });
+
+        ui.greet();
+    }
```
Tianqi-Zhu: This method looks super long. Maybe it is better to split it into several shorter methods.

src/main/java/duke/DukeList.java
```diff
@@ -0,0 +1,95 @@
...
+        return this.database.findTasksByName(pattern);
+    }
+
+    // TODO Change to get size instead
+    /**
+     * Returns a message regarding the current size of the list.
+     *
+     * @return message on the current list size
+     */
+    public String currentSizeMessage() {
```
chingh20, line 81: You might want to change the method name so that it starts with a verb. Perhaps getCurrentSizeMessage?

src/main/java/duke/command/DoneCommand.java
```diff
@@ -0,0 +1,45 @@
...
+public class DoneCommand implements Command {
+    private final DukeList dukeList;
+    private final UserInterface ui;
+
+    /**
+     * Creates a command that marks a task from the given dukelist as completed, and
+     * provide its response to the given ui.
+     *
+     * @param dukeList list of tasks from which the task will marked as completed
+     * @param ui       user interface for displaying responses
```
Tianqi-Zhu, line 21: It seems that there is an unnecessary space here.

src/main/java/duke/database/Database.java
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Recreates a task based on provided information. Used when retrieving task
+     * from SQL database.
+     *
+     * @param type        type of task
+     * @param name        name of task
+     * @param isCompleted {@code true} if the task has been completed
+     * @param date        date of the task, nullable
+     * @return
```
chingh20: I think you might have forgotten to finish writing the comment for @return

src/main/java/duke/command/CommandManager.java
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+            this.defaultExec();
+        }
+    }
+
+    /**
+     * Executes the default behaviour with the input if there are no commands
+     * called.
+     */
+    private void defaultExec() {
```
chingh20, line 56: maybe rename the method as executeDefult?

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class ListCommand implements Command {
+    private final DukeList dukeList;
+    private final UserInterface ui;
+
+    /**
+     * Creates a command that lists tasks in the given dukelist onto the specified
+     * ui.
+     *
+     * @param dukeList of which tasks
```
chingh20, line 17: might be better if you write @param dukeList User's todo list (Start the description with a capitalized letter). Remember to put a punctuation after each parameter description :)

chingh20 (commented):
Good job on following the coding standard! I couldn't find many violations to comment on. One small thing I guess you can pay more attention to is the parameter description for comments. Make sure to end each description with a punctuation. It's not necessary to align one parameter description with others, and starting the description with a capitalize letter might also make it neater.  Overall, your code was very nice and easy to read~ well done :))

src/main/java/duke/Duke.java
solarrabbit99: Indeed XD

src/main/java/duke/command/DoneCommand.java
solarrabbit99: Thanks for the review! It is a linter that auto-formats the param description to auto-align itself. I thought it was ok to leave it as such, but thanks anyways!

src/main/java/duke/command/CommandManager.java
solarrabbit99: Yep, that's better!

src/main/java/duke/command/ListCommand.java
solarrabbit99: It seems like capitalising the first letter of the description isn't a common practice in general, if I were to direct you the java's String class's [documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html). Again, I may be wrong and that it is part of the module's guideline. Thanks anyways!

src/main/java/duke/database/Database.java
solarrabbit99: Haha, indeed...

src/main/java/duke/DukeList.java
solarrabbit99: Yes, that is better! Thanks for the suggestion!

src/main/java/duke/command/ListCommand.java
solarrabbit99: On the point of having punctuation, I'm also not sure if it is a [convention](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html) or a standard in the module. It does make sense that it looks neater that way, however.

PR #72, opened 2021-08-21
[Myat Htet Kyaw] iP

myathk:
# T-800
> "Please give me sensible commands..." -  You dear Terminator

T-800 helps you manage your tasks. It, 
  * supports tasks with repetition
  * is easy to pick up with thorough error handling
  * is fast to use without the need to click
 
You really should,
  1. let T-800 into your life from [this link](https://github.com/NUSmhk/ip/releases/tag/A-Jar)! 
  2. start using it to manage your tasks fast and efficiently
  3. learn it well enough to not piss T-800 off too much

And it is **FREE**! 😃 

Features: 
  - [x] Managing Tasks
  - [x] Managing Deadlines
  - [x] Adding Tasks with repetition

If you LOVE Java OOP and want to improve T-800, here's the `main`  method:

```javascript
        new Duke("./Duke.txt").run();
    }
```


src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,74 @@
...
+import java.time.LocalDate;
+import java.time.LocalTime;
+
+import duke.Deadline;
+import duke.Event;
+import duke.Task;
+import duke.TaskList;
+import duke.ToDo;
+import duke.Ui;
+
```
kflim: Maybe you could add some header comments to the class and the public methods? I noticed most of the Command files are missing these comments as well, perhaps because they are newly committed files

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+
+
+/**
+ * Represents a translator that makes sense of user input commands.
+ */
+public class Parser {
+    private static String detectedFormat;
+    private static LocalTime startTime;
+    private static LocalTime endTime;
+    private static final String[] dateTimeFormats = {
```
kflim, line 24: I like the attention to detail, nice job!

kflim (commented):
In general, the code is compliant with the coding standard. Just some tweaks for some files and it will be good!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+        "dd/MM/yyyy HHmm",
+        "dd-MM-yyyy HHmm",
+        "yyyy/MM/dd HHmm",
+        "yyyy-MM-dd HHmm",
+        "yyyy/MM/dd hh:mm a",
+        "yyyy-MM-dd hh:mm a",
+        "yyyy/MM/dd HHmm",
+        "yyyy-MM-dd HHmm",
+        "dd MMM yyyy hh:mm a"
+    };
```
kflim, line 41: I like the attention to detail, nice job!

myathk: Thank you!

src/main/java/duke/commands/AddCommand.java
myathk: You are right, thank you for pointing it out! I have only added header comments for half of my class and public methods and I am still working towards completing for all of them.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,46 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a type of task that contains a due by deadline.
+ */
+public class Deadline extends Task {
+
+    protected LocalDate date;
+    protected LocalDateTime dateTime;
```
jayasting98: Since if I am not wrong, there are no external package subclasses that require these two variables, perhaps you can consider making these private? On the other hand, even if there were any such subclasses, maybe you could consider making these private regardless and use accessor methods instead? I noticed this in the `Event` class as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+     * This method takes in the user input as a String and runs checks on it to determine
+     * the type of command it is, also throwing exception if the input is invalid.
+     * It also takes in a TaskList that Duke bot uses to load and
+     * store tasks the user adds or deletes.
+     * @param fullCommand String of user input.
+     * @param taskList a list to store tasks the user adds or deletes.
+     * @return Command which Duke bot can execute afterwards.
+     * @throws DukeException if user input cannot be understood.
+     */
+    public static Command parse(String fullCommand, TaskList taskList) throws DukeException {
```
jayasting98: While I do like that your Javadoc header comments are very descriptive, I think it may be desirable to have a short first sentence as a summary (usually one line). Maybe you could consider opting to use more subsequent sentences for describing in more detail. I noticed this in other places as well, though it is a minor issue and it is definitely better that you are descriptive (which I really like) than having no description at all.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,43 @@
...
+    protected LocalDate date;
+
+    /**
+     * This constructor creates an Event task type object.
+     * @param input text description of Event task.
+     * @param date date of occurrence of Event.
+     * @param startTime start time of Event.
+     * @param endTime end time of Event.
+     */
+    public Event(String input, LocalDate date, LocalTime startTime, LocalTime endTime) {
```
jayasting98: Is there not supposed to be an empty line between the description and the parameter section in the Javadoc header comment? I noticed this in your other Javadoc header comments as well, though it is a minor issue.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,43 @@
...
+     * This method formats task description and duration in a user-friendly way,
+     * additionally marking the Event task type with [E]
+     * @return String of formatted task description and deadline.
+     */
+    @Override
+    public String toString() {
+        return String.format("[E]%s (at: %s from: %s)", super.toString(),
+                this.date.format(DateTimeFormatter.ofPattern("dd MMM yyyy")),
+                    this.startTime.format(DateTimeFormatter.ofPattern("hh:mm a")) + "-"
+                        + this.endTime.format(DateTimeFormatter.ofPattern("hh:mm a")));
```
jayasting98: Maybe you can consider [using intermediate variables](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions) to make this clearer.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+                String content = fullCommand.split(" ", 2)[1].trim();
+                return parseDeadline(content);
+            } else if (fullCommand.startsWith("event")) {
+                String content = fullCommand.split(" ", 2)[1].trim();
+                return parseEvent(content);
+            } else {
+                throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+        }
+    }
```
jayasting98: Perhaps you could consider extracting several sub-methods from this to [avoid this pretty long method](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods).

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,286 @@
...
+                } else {
+                    throw new DukeException("OOPS!!! Please enter a valid date in duration!");
+                }
+            } else {
+                throw new DukeException("OOPS!!! The duration of an event cannot be empty.");
+            }
+        } else {
+            throw new DukeException("OOPS!!! The duration of an event cannot be empty.");
+        }
+    }
```
jayasting98: Is this [amount of nesting](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting) necessary? Maybe you can consider moving each nested `if` block out of its parent `if` and then put it after the parent `if` block since you can take advantage of the thrown exceptions to make subsequent statements unreachable as intended. Not sure if that makes sense so here's an example.
From this:
```java
if (isCondition1) {
    // everything within these braces is unreachable if isCondition1 is false
    if (isCondition2) {
        // ...
    } else {
        // ...
    }
} else {
    // ...
    throw new DukeException("...");
}
```
To this:
```java
if (isCondition1) {
    // ...
} else {
    // ...
    throw new DukeException("...");
}
// everything after this is still unreachable if isCondition1 is false because
// the exception would then be thrown before this
if (isCondition2) {
    // ...
} else {
    // ...
}
```

jayasting98 (commented):
Overall, pretty good code, just a few minor issues. I really love the amount of detail you go to such as checking for many possible cases for exceptions. I also like how descriptive your Javadoc header comments are.

src/main/java/duke/Deadline.java
myathk: Thank you! I just found out the difference between private and protected so I will change them accordingly.

src/main/java/duke/Parser.java
myathk: I also just found out that a shorter description seems to be the norm so I will stick to that in the future, thank you!

src/main/java/duke/Event.java
myathk: You are correct, I seem to have missed out on that, it is indeed a violation of the coding standard. Thanks for pointing it out, I will change them.

src/main/java/duke/Event.java
myathk: Thank you, I've changed it.

src/main/java/duke/Parser.java
myathk: Thank you! You are right, it is my mistake. I have improved it.

src/main/java/duke/Parser.java
myathk: Thank you for your clear explanation, it helped me understand the necessary changes which I've implemented!

PR #73, opened 2021-08-21
[Lau Teng Hon] iP

Th-429B:
# DukeNotPro
> "The more I C, the less I see."
> -Annoymous

DukeNotPro came at the cost of many hours and sleep. It's
- definitely not text-based
- easy to use(?)
- ~~FAST!~~ Super Fast... maybe 

Follow these instructions to download:
1. download it from [here](https://www.youtube.com/watch?v=6n3pFFPSlW4)
2. look at 4.
3. there's no 4 :rofl:
**teehee**

Features:
- [x] Managing tasks.
- [ ] GUI (coming soon)

Now for some random filler code here's the ```main``` method.
```java
public class Main {
        System.out.print("Have a nice day! Take a break sometimes.");
    }
}
```


src/main/java/duke/ChatBot.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * ChatBot is a class that encapsulates what a chat bot has.
+ * It functions as a chat bot that responds to user inputs and
+ * does tasks specified by the user.
+ */
+public class ChatBot {
+
+    private ArrayList<Task> temp;
+    private TaskList tasklist;
```
loose-bus-change, line 20: Are these two collections ?

src/main/java/duke/ChatBot.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public String handleDone(int index) {
+        Task t = tasklist.completeTask(index);
+        s.saveTasks(tasklist);
+        return ui.showCompleteTaskMessage(t);
+    }
+
+    /**
+     * This method is to add a new deadline to the storage and
+     * print out the add message when the task is added successfully.
```
loose-bus-change: Personally I like this header, its very clear and neat! You could also consider just beginning the description with 'returns' or 'adds' or 'sends' etc. So for instance, 'adds a new deadline ....'.

src/main/java/duke/ChatBot.java
```diff
@@ -0,0 +1,170 @@
...
+     *
+     * @param name Name of the deadline.
+     * @param deadline The date and time of the deadline.
+     */
+    public String handleDeadline(String name, LocalDateTime deadline) {
+        Task t = tasklist.addDeadline(name, deadline, false);
+        s.saveTasks(tasklist);
+        return ui.showAddTaskMessage(t, tasklist.getTotalTasksNumber());
```
loose-bus-change, line 82: I like the names of these methods, very descriptive and clear!

src/main/java/duke/ChatBot.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Calls the UI to return the start message.
+     */
+    public String handleStart() {
+        return ui.showStartMessage();
+    }
+
+    /**
+     * This method gets the total number of tasks.
+     * @return total number of tasks.
```
loose-bus-change, line 170: I think there should be a blank line between the description and the param/return

loose-bus-change (commented):
LGTM!

src/main/java/duke/ChatBot.java
Th-429B: Yes they are, just that TaskList is an abstraction of the collection of tasks.

src/main/java/duke/ChatBot.java
Th-429B: Will update the javadocs to fit documentation standards

src/main/java/duke/ChatBot.java
Th-429B: Noted, will make the changes

PR #76, opened 2021-08-21
[Ivan Wong] iP

wpinrui:
# Good Janet :hugs:
> Elenor told me that instead of being sad, I should, "Go get it, girl." So I'm going to go get it, girl. I'll get everything, just to be safe. - Good Janet [(source)](https://img.buzzfeed.com/buzzfeed-static/static/2019-09/5/18/asset/38945558c66c/sub-buzz-283-1567708506-1.jpg?downsize=600:*&output-format=auto&output-quality=auto)

Good Janet helps you track your tasks. Here are some features:
- GUI
- Easy to learn
- **Super fast to use!**

## Getting started
1. Download the **GUI version** of the JAR file in the releases
2. Open a terminal window in the JAR file location
3. Run the following: `java -jar janet.jar`

## Planned features
- [x] Add tasks
- [x] Delete tasks
- [ ] Update tasks

## Advanced Users
The save file is currently hardcoded. If you want to change where the save file is located, you can edit the source code and recompile the JAR file. Simply change "janet.txt" in the constructor to any custom path you like, e.g. "C:/Users/User/Documents/janet.txt" and it should work *as long as the folder exists*. 

```java
public Janet() {
        storage = new Storage("janet.txt"); // this path can be changed
        try {
            tasks = storage.readSave();
        } catch (IOException | ClassNotFoundException e) {
            System.out.println(e);
        }
    }
```


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Returns the specific operation as a String that Duke is able to understand.
+     *
+     * @return Duke operation
+     */
+    public String getOperation() {
+        return operation;
+    }
```
tzejit: I like how you used getters to get information instead of accessing it directly from the instance itself!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Returns a Command that Duke understands from a string passed to Duke by
+     * the user.
+     *
+     * @param input String that user passed to Duke
+     * @return Command that Duke is able to execute
+     * @throws DukeException If invalid or insufficient arguments are provided
+     */
+    public static Command parseUserInput(String input) throws DukeException {
+        String[] strings = input.split(" ");
```
tzejit: Should this variable have a more intuitive name?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,66 @@
...
+    public Task delete(int index) {
+        return tasks.remove(index);
+    }
+
+    /**
+     * Returns the number of tasks in the task list.
+     *
+     * @return Number of tasks in list
+     */
+    public int size() {
```
tzejit: Maybe the method can be named after a verb more such as getSize?

tzejit (commented):
Overall really good code quality! Almost all namings follow the proper convention, good OOP practices are followed and methods have well written javadoc comments!

src/main/java/duke/Parser.java
wpinrui: Thanks for spotting it. I've changed it to `inputStringParts`.

src/main/java/duke/TaskList.java
wpinrui: I'm unsure about this, because the method for ArrayList is named 'size'.

src/main/java/janet/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            byDate = LocalDate.parse(by);
+        } catch (DateTimeParseException e) {
+            byDate = null;
+        }
+    }
+
+    /**
+     * Returns a formatted String representing the deadline of the task in the form
+     * MMM d YYYY.
```
itsraveen, line 35: I like how your JavaDoc here is very clear with what the method does. I think its great that you included 'MMM d YYYY' in your code.

src/main/java/janet/Janet.java
```diff
@@ -0,0 +1,126 @@
...
+        case "find":
+            return handleFind(command);
+        case "schedule":
+            return handleSchedule(command);
+        case "echo":
+            return command.getDescription();
+        default:
+            return "";
+        }
+    }
```
itsraveen, line 75: Code very well written with each of the events being handles very well. It becomes clear when that you know what each of the cases does since you have a method to handle each cases.

src/main/java/janet/Janet.java
```diff
@@ -0,0 +1,126 @@
...
+        storage.writeSave(tasks);
+        return Ui.addedString(task, tasks.size());
+    }
+
+    private String handleNewEvent(Command command) throws IOException {
+        Event task = new Event(command.getDescription(), command.getTime());
+        tasks.add(task);
+        storage.writeSave(tasks);
+        return Ui.addedString(task, tasks.size());
+    }
```
itsraveen: Would this be better if it was all in a single method and handles the different cases? Most of the code are repeated code so it might help to make it shorter. Other than that, looks good to me.

src/main/java/janet/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+        case "event":
+            return parseEvent(input, inputStringParts);
+        case "delete":
+            return parseDelete(inputStringParts);
+        case "find":
+            return parseFind(input, inputStringParts);
+        case "schedule":
+            return parseSchedule(input, inputStringParts);
+        default:
+            throw new JanetException(Ui.UNRECOGNISED_OPERATION);
```
itsraveen, line 44: Once again the switch statements are handled very well!

itsraveen (commented):
Code looks clean with all the JavaDocs being clear and concise. Looks good to me!

src/main/java/janet/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+    static final String UNRECOGNISED_OPERATION = "Sorry, I do not understand this command.";
+    static final String INDEX_OUT_OF_BOUNDS_STRING = "Sorry, there is no task with the index number ";
+
+    /**
+     * Returns a string representation of the tasks in the list.
+     *
+     * @param taskList Janet's list of tasks
+     * @return String representation of list
+     */
+    public static String taskListString(TaskList taskList) {
```
Linda124: Perhaps you could name this method as printTaskListString so that it sounds more like a verb?

src/main/java/janet/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+                        + "\nNow, you have %d tasks in the list.\n",
+                task, size);
+    }
+
+    /**
+     * Returns a string that is shown when a user tries to access a non-existent task index.
+     * @param index The index of the non-existent task that the user tried to access
+     * @return String to inform the user of index out of bounds error
+     */
+    public static String outOfBoundsString(int index) {
```
Linda124: Would printOutOfBoundsString be a bit more intuitive here? I noticed that a similar approach could be used for the methods "deletedString", "doneString" and "addedString".

src/main/java/janet/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+        assert(index <= tasks.size());
+        return tasks.remove(index);
+    }
+
+    /**
+     * Returns the number of tasks in the task list.
+     *
+     * @return Number of tasks in list
+     */
+    public int size() {
```
Linda124, line 46: Would it be better to name this method getSize so that the function of the method is better understood?

src/main/java/janet/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+        return tasks.size();
+    }
+
+    /**
+     * Returns the task stored at a particular index in the list.
+     *
+     * @param index Index of the task in the list of tasks
+     * @return Task found at that index in the list
+     */
+    public Task get(int index) {
```
Linda124, line 56: Since I suggested that size could be changed to getSize, perhaps you could consider changing this method name to getIndex as well.

src/main/java/janet/Janet.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private String handleTaskDone(Command command) throws IOException {
+        Task task = tasks.get(command.getIndex() - 1);
+        task.setDone(true);
+        storage.writeSave(tasks);
+        return Ui.doneString(task);
+    }
+
+    private String handleNewToDo(Command command) throws IOException {
```
Linda124, line 89: I like how most of your method names and variable names are easy to understand.

Linda124 (commented):
Overall, the code quality is really good! Generally, good naming conventions were used.

src/main/java/janet/Janet.java
wpinrui: Thanks for your feedback. I was considering doing this but realised I would have many if-else statements. Now that I think about it further, it may be possible to use polymorphism to achieve this 😄

src/main/java/janet/Ui.java
wpinrui: Ahh, good catch. Thank you!

src/main/java/janet/Ui.java
wpinrui: I'm a little unsure on this, because this method returns a String whereas printXXX might give the impression of a void method that has the side-effect of logging to console. Thanks for pointing it out, I may open an issue to ask for more opinions.

PR #77, opened 2021-08-21
[Gu Geng] iP

jackgugz:
# DUKE the Super-Duper

> Free thing is great thing. 
> -[Jack](https://github.com/jackgugz) 

### Duke frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use
Thumb up image

### All you need to do is,

1. download it from [here](https://github.com/jackgugz/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 👍 
And it is **FREE**! ⚡ 

Features:

 - [X] Managing todos
 - [X] Managing deadlines
 - [X] Managing events
 - [X] Find tasks on specified date
 - [X] Find tasks with specified keywords

#
```
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,116 @@
...
+            FileWriter fw = new FileWriter(temp.toString());
+            File tempFile = new File(temp.toString());
+            for (int i = 0; i < taskList.size(); i++) {
+                duke.task.Task task = taskList.getTask(i);
+                fw.write(task.record() + System.lineSeparator());
+            }
+            fw.close();
+            Files.delete(filePath);
+            Files.copy(temp, filePath);
+            //noinspection ResultOfMethodCallIgnored
```
Jeromesyl, line 110: What does this comment mean? Should it be included?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return result.toString();
+    }
+
+    /**
+     * Marks the specified Task in the ArrayList as done.
+     *
+     * @param index An int indicating the position of Task in the ArrayList that is to be marked as done.
+     */
+    public void doneTask(int index) {
```
Jeromesyl, line 106: Should TaskList and Task classes have the same method name here? It might be confusing.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,67 @@
...
+public class AddCommand extends Command {
+    private duke.task.Task task;
+
+    /**
+     * Returns a AddCommand object with the information provided.
+     *
+     * @param command A String containing information that can possibility be used to create an AddCommand object.
+     * @throws duke.DukeException Will be thrown if information provided are insufficient/incorrect.
+     */
+    public AddCommand(String... command) throws duke.DukeException {
```
Jeromesyl, line 20: Perhaps it will be better formatting to use 
```
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}
``` 
instead of nested
```
if (condition) {
    statements;
} else {
    statements;
}
```

Jeromesyl (commented):
Very neat and nice formatting overall! I liked how evenly spaced everything is. All public methods had javadocs and most method names are logical. Hardly any coding standard violations.

src/main/java/duke/Storage.java
Jeromesyl: What does this comment mean? Should it be included?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,140 @@
...
+        Task holder = taskList.get(index);
+        holder.doneTask();
+    }
```
Jeromesyl: Should TaskList and Task object have the same method name here? It might be confusing.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,67 @@
...
+                            " ☹ HEY DEAR! Please enter a date after / following the task description");
+                }
+                if (command[0].equals("deadline")) {
+                    task = new duke.task.Deadline(command);
+                } else {
+                    task = new duke.task.Event(command);
+                }
+            }
+        }
+    }
```
Jeromesyl: Perhaps it will be better formatting to use 
```
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}
``` 
instead of nested
```
if (condition) {
    statements;
} else {
    statements;
}
```

src/main/java/duke/Storage.java
jackgugz: You made me wonder also, I think it is a garbage comment I forgot to delete! Thanks HAHA!

src/main/java/duke/TaskList.java
jackgugz: Hmm a good point from you Jerome so any name you would like to suggest please? Cause it seems regardless what alternative I can currently think of, it will sound pretty similar and still cause the confusion.

src/main/java/duke/command/AddCommand.java
jackgugz: You have a valid point here! Thanks!

PR #81, opened 2021-08-21
[Gan Hong Yao] iP

ganhongyao:
## Duke

> "If you fail to plan, you are planning to fail." - Benjamin Franklin

Try out Duke **today**!

### Features
- Add tasks
- Mark tasks as done
- Delete tasks
- Save tasks to local file
- Find tasks by name or date

### Getting started
To get started with Duke, it's as easy as 1️⃣ 2️⃣ 3️⃣ !
1. Download the latest release from [here](https://github.com/ganhongyao/ip/releases).
2. Double-click to open it.
3. _Happy Duking!_

### To-do
- [x] Add GUI
- [ ] Add more unit tests (any help is deeply appreciated 😆 )

Unit test methods should adopt the following naming convention `whatIsBeingTested_descriptionOfTestInputs_expectedOutcome`. 

Example of a test method:
```java
@Test
public void execute_nonIntegerArgument_exceptionThrown() {
    try {
        new DeleteTaskCommand("blah")
                .execute(new TaskManager(), new Storage("./data/tasks.txt"));
        fail();
    } catch (DukeException e) {
        assertEquals("Invalid task number.", e.getMessage());
    }
}
```

If you find any bugs, please ~~fix it~~ report it to me!


src/main/java/duke/parser/CommandParser.java
```diff
@@ -0,0 +1,49 @@
...
+     * Parses the command as well as any arguments entered by the user.
+     * @param fullCommand the full command entered by the user
+     * @return the corresponding <code>Command</code>, or the <code>HelpCommand</code> if the command is not recognised
+     */
+    public static Command parse(String fullCommand) {
+        final String trimmedCommand = fullCommand.trim();
+        final String commandWord = trimmedCommand.split("\\s+")[0];
+        final String commandArguments = trimmedCommand.replace(commandWord, "").trim();
+
+        switch (commandWord) {
```
guangxuan, line 30: Consider using an enum instead of String

src/main/java/duke/task/ToDo.java
```diff
@@ -0,0 +1,25 @@
...
+        super(name);
+    }
+
+    public ToDo(String name, boolean isDone) {
+        super(name, isDone);
+    }
+
+    @Override
+    public String toText() {
+        String[] props = new String[]{"T", super.getStatusIcon(), super.getName()};
```
guangxuan:
```suggestion
        String[] props = {"T", super.getStatusIcon(), super.getName()};
```

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,54 @@
...
+     * Returns the name of the task.
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * Returns whether the task is marked as completed.
+     */
+    public boolean checkTaskDone() {
```
guangxuan:
```suggestion
    public boolean isDone() {
```

guangxuan (commented):
LGTM. Very comprehensive analysis of exceptions, commands, and even Parsers!
There is not much to be said about Code Quality. Good job!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    protected Task(String name) {
+        this(name, false);
+    }
+
+    protected Task(String name, boolean isDone) {
+        this.name = name;
+        this.isDone = isDone;
+    }
+
+    protected Task markAsDone() {
```
YoYoCiti, line 22: Maybe method name should be setDone? As coding standard said setter methods for boolean variables should be of that form.

src/main/java/duke/task/Task.java
YoYoCiti: I agree with this change as well

src/main/java/duke/parser/CommandParser.java
```diff
@@ -0,0 +1,49 @@
...
+import duke.command.HelpCommand;
+import duke.command.ListCommand;
+
+/**
+ * Parses user input.
+ */
+public class CommandParser {
+    /**
+     * Parses the command as well as any arguments entered by the user.
+     * @param fullCommand the full command entered by the user
```
YoYoCiti: I think there should be an empty line between the description and parameter section? Noticed this is a recurring issue

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,208 @@
...
+            String pluralised = undoneTaskCount > 1 || undoneTaskCount == 0 ? "tasks" : "task";
+            return String.format(MARKED_TASK_AS_DONE_MESSAGE, task, undoneTaskCount, pluralised);
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeException(TASK_NOT_FOUND_MESSAGE);
+        }
+    }
+
+    /**
+     * Deletes a <code>Task</code> from the list.
+     * @param taskNumber the number of the task to be deleted
```
YoYoCiti: I think a punctuation is required behind the parameter description?

YoYoCiti (commented):
Overall, the code was very easy and pleasant to read. Only picked out extremely nitpicky and minor coding standard violations. Thanks for the read :smile:

src/main/java/duke/task/ToDo.java
ganhongyao: Done.

src/main/java/duke/task/Task.java
ganhongyao: Thanks for the suggestion. Have changed it! 😄

src/main/java/duke/task/Task.java
ganhongyao: I personally think `setDone` would be a more appropriate name for an actual setter method i.e. 
```java
protected Task setDone(boolean isDone) {
    this.isDone = isDone;
}
```
, which may be implemented in future. Hence I chose the name `markAsDone` so that we don't confuse the methods.

Thanks for the suggestion though!

src/main/java/duke/task/TaskManager.java
ganhongyao: Yup you're right. Have fixed this!

src/main/java/duke/parser/CommandParser.java
ganhongyao: You are right. Thanks for pointing this out to me! Have fixed this. 😄

PR #85, opened 2021-08-21
[Lye Wen Jun] iP

xlzior:
# Iris 🤖

> Who needs _Siri_ when you have **Iris**?!

## Features

- simple, text-based interface
- written in Java: _write once, run anywhere_

## How to use

1. Create a todo with `todo [task name]`
2. View all tasks with `list`
3. Mark the todo as complete with `done 1`
4. Delete the completed todo with `delete 1`

## Developer Experience

Iris is easily extendable -- simply create a new class in the `command` package which extends the `Command` base class, and implement `runSilently(TaskList tasks)` and `say(TaskList tasks, Ui ui)`.

```java
public class FindCommand extends Command {
    private String searchTerm;
    private List<Task> searchResults;

    public FindCommand(String searchTerm) {
        this.searchTerm = searchTerm;
    }

    @Override
    public void runSilently(TaskList tasks) throws IrisException {
        this.searchResults = tasks.find(this.searchTerm);
    }

    @Override
    public void say(TaskList tasks, Ui ui) {
        for (int i = 0; i < searchResults.size(); i++) {
            ui.say(String.format("%d. %s", i + 1, searchResults.get(i)), i == 0);
        }
    }
}
```

### Tasks

- [x] Add different types of tasks (todos, deadlines, events) to your todo list
- [X] Manage tasks in your todo list by marking them as complete or deleting them
- [X] Find tasks in your todo list by searching for them
- [ ] Access Iris through a GUI

Original repository: [link](https://github.com/xlzior/ip)


src/main/java/iris/Parser.java
```diff
@@ -0,0 +1,74 @@
...
+            int index = parseInt(getMetadata(command));
+            return new DeleteCommand(index);
+        } else if (command.startsWith("find")) {
+            String searchTerm = getMetadata(command);
+            return new FindCommand(searchTerm);
+        } else if (command.startsWith("bye")) {
+            return new ByeCommand();
+        } else {
+            throw new IrisException("I'm sorry, but I don't know what that means.");
+        }
```
clyveycui: Using switch case might make it look neater, but that's just my personal opinion.

src/main/java/iris/TaskList.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Gets the Task from the TaskList at a given index
+     *
+     * @param index index of task to get
+     * @return Task at the given index
+     */
+    public Task get(int index) {
+        return tasks.get(index);
+    }
```
clyveycui: Naming it getTask might make it's usage clearer, but this is fine too.

src/main/java/iris/task/Deadline.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Override
+    public String toCommand(int index) {
+        return String.format(
+                "deadline %s /by %s\n%s",
+                this.name,
+                this.by,
+                super.toCommand(index)
+        );
+    }
```
clyveycui: Pretty cool design here, first one that I have seen using a toCommand method for each individual task object.

clyveycui (commented):
Awesome, just a few minor nitpicks, but lgtm.

src/main/java/iris/Parser.java
xlzior: tbh I did if/else on purpose because I was worried they would add a new requirement for commands which aren't a single word, but yeah I think I can change it to switch/case now. thank you!

src/main/java/iris/TaskList.java
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private void validateTaskIndex(int index) throws IrisException {
+        if (index <= 0) {
+            throw new IrisException("Please enter a valid task index.");
+        }
+        int count = tasks.size();
+        if (index > count) {
+            throw new IrisException(String.format("Your task list only has %d items", count));
+        }
+    }
```
Amadeus-Winarto: I like this explicit validation method

src/test/java/iris/DeadlineTest.java
```diff
@@ -0,0 +1,46 @@
+package iris;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+import iris.task.Deadline;
+
+public class DeadlineTest {
```
Amadeus-Winarto, line 10: General question (and may have missed it also), but will you be doing a test on storage? Like if it successfully read/write to storage file?

Amadeus-Winarto (commented):
Great job! Nice design too!

PR #96, opened 2021-08-21
[Liew Xin Yi] iP

xyliew25:

* text-based
* easy to learn
* ~~FAST~~ *SUPER* FAST to use

All you need to do is,

1. download it from [here](https://github.com/xyliew25/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :wink:

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

````java
public class Main {
    }
}
````

### CODE REVIEW

File: src/main/java/duke/command/Command.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Returns the response of this command.
+     *
+     * @param tasks   Tasks of the Duke program.
+     * @param ui      Ui of the Duke program.
+     * @param storage Storage of the Duke program.
+     * @throws DukeException If there is/are invalid argument(s) passed into this command,
+     *                       or changes cannot be saved to storage.
+     */
+    public abstract String executeAndGetResponse(TaskList tasks, Ui ui, Storage storage) throws DukeException;
```
wei2912, line 21: This function seems to be essentially equivalent to `execute` other than the fact that it has a `String` return type. Was there any specific reason why you chose to have a separate abstract function, rather than to change the return type of `execute` and return empty/null strings if there is no response?

Looking at the classes that extend from `Command`, it seems that this has led to multiple [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) violations. I think reworking your codebase to have just the `execute` method would be beneficial for code quality.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param tasks Tasks of the Duke program.
+     * @param ui Ui of the Duke program.
+     * @param storage Storage of the Duke program.
+     */
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        TaskList matchingTasks = new TaskList();
+        for (int i = 0; i < tasks.getSize(); i++) {
+            Task currTask = tasks.getTask(i);
```
wei2912: You might prefer to use the enhanced `for` loop here, i.e. `for (Task currTask : tasks)`. (This also applies to the loops in other command classes.)

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,82 @@
...
+            if (currTask.toString().contains(keyword)) {
+                matchingTasks.addTask(currTask);
+            }
+        }
+        String response;
+        if (matchingTasks.getSize() == 0) {
+            response = "There are no matching tasks.";
+        } else {
+            response = "Here are the matching tasks in your list:\n" + matchingTasks;
+        }
```
wei2912: I think the ternary operator (`x ? y : z`) can be used here to make the code slightly neater.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Returns the string representation of this Deadline task to be saved in storage.
+     *
+     * @return the string representation of this Deadline task to be saved in storage.
+     */
+    @Override
+    public String toStringData() {
+        return "D | " + super.toStringData() + " | " + this.by;
```
wei2912: Consider using `String.format` which might be easier to read compared to string concatenation.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,44 @@
+package duke.task;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents the Deadline task in the Duke program.
+ */
+public class Deadline extends Task {
+    protected LocalDate by;
```
wei2912, line 10: While this isn't technically a coding standard violation (to my knowledge), it might be clearer to name this variable with a noun, e.g. `date`.

wei2912 (commented):
Great work so far! I really like how you've structured your codebase, as creating specific command classes and exceptions would be beneficial to future expansion.

I have pointed out a few areas where there is room for improvement, let me know if you have any queries!

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,82 @@
...
+            Task currTask = tasks.getTask(i);
+            if (currTask.toString().contains(keyword)) {
+                matchingTasks.addTask(currTask);
+            }
+        }
+        String response;
+        if (matchingTasks.getSize() == 0) {
+            response = "There are no matching tasks.";
+        } else {
+            response = "Here are the matching tasks in your list:\n" + matchingTasks;
```
mweeruien: Would it be better to leave a line break in the middle for greater readability?

src/main/java/duke/task/Deadline.java
mweeruien: I agree with @wei2912. The usage of `String.format()` can also be considered in other areas of your code.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,153 @@
...
+                throw new MissingArgumentException("Deadline", "/by");
+            }
+            // Extract and parse by argument
+            try {
+                by = LocalDate.parse(arguments.split(" /by ")[1]);
+            } catch (DateTimeParseException err) {
+                throw new InvalidDateFormatException();
+            }
+            // If pass all checking, create and return deadline command obj
+            return new DeadlineCommand(description, by);
```
mweeruien: I really like the comments, they help in understanding the code! However, would it be better to leave several line breaks here as well to improve readability? The other `case` blocks can also be reviewed to see if the addition of line breaks would improve readability.

mweeruien (commented):
Your code adheres to the coding standards well, and I think the comments and Javadocs that complement your code were also well-written! Though there are several small changes that can be considered to increase the readability, it was overall a good job!

src/main/java/duke/command/Command.java
xyliew25: This was a quick fix after adding the GUI as the response of each command is to be returned, unlike in the CLI, the response is just to be printed on the console. However, your comment gave me an idea to refactor my code. Something as follows:
````java
public void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException {
    String response = executeAndGetResponse(tasks, ui, storage);
    ui.showResponse(response);
}
````

Also, your comment made me realized that I should have mentioned its purpose in the header comment.

Thanks!

src/main/java/duke/command/FindCommand.java
xyliew25: Yes, thank you for the suggestion!

src/main/java/duke/command/FindCommand.java
xyliew25: Yes, indeed it will be much neater and I won't need to declare `response` without assigning it with a value. Thanks!

src/main/java/duke/task/Deadline.java
xyliew25: Hm I see. Thank you for the suggestion!

src/main/java/duke/task/Deadline.java
xyliew25: I reserved this suggestion as I see this variable as the argument `by` of a `Deadline` task instead of the preposition 'by'. However, you do have a point as it does not sound like a noun.

src/main/java/duke/command/FindCommand.java
xyliew25: Thank you for the suggestion!

src/main/java/duke/util/Parser.java
xyliew25: Glad the comments help! Yes, I shall add in the line breaks.

PR #105, opened 2021-08-21
[Kleon Ang] iP

kleonang:
# Duke™
> "Object-oriented programming is an exceptionally bad idea which could only have originated in California." 
> – Edsger W. Dijkstra ([Source](https://www.cs.scranton.edu/~mccloske/dijkstra_quotes.html))

Duke _wipes_ your mind of having to remember things you need to do. It is 
- Text-based
- Easy to learn
- ~~FAST~~ ~~SUPER FAST~~ BLAZING FAST to use

All you need to do is:
1. Download it from [here](https://github.com/kleonang/ip/releases/download/A-Release/duke.jar)
2. Double-click it
3. Add your tasks
4. ~~Hope it doesn't crash and~~ Let it manage your tasks for you 😊

What's more, it's **free*** and _currently_ **open-source****!

Features:
- [x] Add tasks
- [x] List tasks
- [x] Delete tasks
- [x] Find tasks
- [x] Mark tasks as done
- [x] Assign priority to tasks
- [x] Interactive GUI 
- [ ] Enlightens you with the meaning of life
#
```java
/**
 * Compulsory JavaDocs to avoid getting arrested by the Coding Standard™ Police.
 *
 * @param args Command-line arguments.
 */
     new Duke("duke.txt").run();
 }
```
#
<sub>
* For the one (1) month trial period. At the end of the one (1) month trial, you will be charged the standard monthly Duke subscription price of $19.99 USD per month.
</sub>  
<br>
<sub>
** Patent pending. All rights reserved. Duke LLC has extensive legal resources at its disposal.
</sub>


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+        this.parser = new Parser(this.tasks);
+    }
+
+    /**
+     * Run Duke, accepting commands until a "bye" command is issued.
+     */
+    public void run() {
+        while (true) {
+            try {
+                String readIn = this.ui.nextLine();
```
Joel-Sung: I think it would be good to use nouns for variables. For example, instead of readIn, you could use something like userInput.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+                            matchingTasks.add(task);
+                        }
+                    }
+                }
+                list(matchingTasks);
+                break;
+            }
+            default:
+                throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
```
Joel-Sung: For the switch statement, I do not think it is necessary to put curly braces around each case. As it is not a standard coding convention, it may confuse readers on what the curly braces is for. However, if you think they help readability, then I think it is okay to keep it.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * This class encapsulates the mechanism to parse user commands.
+ *
+ * @author Kleon Ang
+ */
+public class Parser {
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-d H:mm");
+    private final TaskList tasks;
+    private boolean toRewriteData;
+    private boolean isExit;
```
Joel-Sung: I think changing the name to 'toExit' is more intuitive as the variable is to keep track when Duke needs to exit. What do you think?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private void done(int counter) throws DukeException {
+        if (counter <= 0 || counter > tasks.size()) {
+            throw new DukeException("Sorry, no such task of index " + counter + ".");
+        }
+        Task doneTask = tasks.get(counter - 1);
+        doneTask.markAsDone();
+        Ui.printReply("Nice! I've marked this task as done:\n  " + doneTask);
+    }
+
```
Joel-Sung, line 29: I think it would be good to add JavaDocs to list and done. They do not seem like trivial methods.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Getter for the exit status of the last parsed command.
+     *
+     * @return True if ready to exit Duke after "bye" command.
+     */
+    public boolean toExit() {
+        return this.isExit;
+    }
+
```
Joel-Sung: I think for methods, maybe try to keep their names as verbs, such as needsRewrite and needsToExit.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,53 @@
...
+    public void showLoadingError(String fileName) {
+        printReply(fileName + " not found. File has been created.");
+    }
+
+    /**
+     * Returns the next line of the user input.
+     *
+     * @return A string containing the next line of user input.
+     */
+    public String nextLine() {
```
Joel-Sung: I think it is better to name your methods as a verb to improve readability. In this case, the name could be getNextLine.

Joel-Sung (commented):
Overall, LGTM. Maybe just some method and variable names could be changed. The code is generally really well-written and follows the coding standard. Good job! 😺

src/main/java/duke/Parser.java
samyipsh: agreed, but sticking to convention of boolean naming convention perhaps isClosed / isClosing / isExiting

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+            tasksBuilder.append(counter).append(".").append(currentTask);
+            if (i < tasks.size() - 1) {
+                // Append newline for all tasks before last task
+                tasksBuilder.append("\n");
+            }
+        }
+        Ui.printReply(tasksBuilder.toString());
+    }
+
+    private void done(int counter) throws DukeException {
```
samyipsh: "counter" varName might be a bit ambiguous, perhaps "TaskIndex" would be more intuitive

src/main/java/duke/Parser.java
samyipsh: Agreed, since its a getter could also consider more default naming convention like getExitStatus(), getRewriteStatus()

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+     * @return A TaskList containing the updated Tasks after command is parsed.
+     * @throws DukeException A Duke-specific exception that occurs when user input is parsed.
+     */
+    public TaskList parse(String readIn) throws DukeException {
+        if (readIn.equals("bye")) {
+            Ui.printReply("Bye. Hope to see you again soon!");
+            this.isExit = true;
+        } else if (readIn.equals("list")) {
+            list(this.tasks);
+        } else {
```
samyipsh: perhaps last else block is unnecessary and can be removed to neaten code logic.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+            switch (command) {
+            case "done": {
+                if (commandArguments.length < 2) {
+                    throw new DukeException("☹ OOPS!!! The index of '" + command + "' cannot be empty.");
+                }
+                this.toRewriteData = true;
+                int counter = Integer.parseInt(arguments);
+                done(counter);
+                break;
+            }
```
samyipsh: As joel mentioned, might be good to stick to convention and not put {} around switch, case blocks

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+                }
+                this.toRewriteData = true;
+                String[] splitTask = arguments.split(" /at ");
+                if (splitTask.length < 2) {
+                    throw new DukeException("Please indicate the event time frame using '/at'.");
+                }
+                String description = splitTask[0];
+                String atString = splitTask[1];
+                try {
+                    LocalDateTime at = LocalDateTime.parse(atString, FORMATTER);
```
samyipsh: can reconsider naming of var "at", perhaps better would be "eventTime" / "dateTime" / "time"

src/main/java/duke/Reply.java
```diff
@@ -0,0 +1,30 @@
+package duke;
+
+/**
+ * This class encapsulates a reply from Duke.
+ *
+ * @author Kleon Ang
+ */
+public class Reply {
+    private static final String LINE = "\t____________________________________________________________";
+    private final String content;
```
samyipsh: final var name convention -> may be good to rename as all caps: CONTENT

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
+package duke;
+
+import java.io.BufferedWriter;
```
samyipsh, line 3: nice very organised imports

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class encapsulates a list of Tasks.
+ *
+ * @author Kleon Ang
+ */
+public class TaskList extends ArrayList<Task> {
```
samyipsh, line 11: interesting way to go bout w TaskList, perhaps a composition might be more suitable in case you want to add other details inside TaskList like timeAccessed juz to name a random example. With inheritance, its somewhat more limiting if you follow the logic that TaskList is an ArrayList (imo)

samyipsh (commented):
Hi! nice code very neat and organized. Esp like the import, package and clarity of logic statements. Just a few tweaks and LGTM 😄

kleonang:
> Overall, LGTM. Maybe just some method and variable names could be changed. The code is generally really well-written and follows the coding standard. Good job! 😺

Thanks for your PR review! I'll make amendments based on your insightful comments.

kleonang:
> Hi! nice code very neat and organized. Esp like the import, package and clarity of logic statements. Just a few tweaks and LGTM 😄

Thanks for your PR review! I'll make amendments based on your insightful comments.

src/main/java/duke/Parser.java
kleonang: Yes agreed, I'll be changing this field name to `toExit` to make more linguistic sense. I think it should also conform to the naming convention given that other similar prefixes like `is`, `has`, and `was` are also accepted. Thanks for your input everyone!

src/main/java/duke/TaskList.java
kleonang: I think it works to add local fields to TaskList as an inherited ArrayList if we require additional variables, plus inheritance allows me to use the predefined iterator as well as `ArrayList::add` and `ArrayList::remove` functions for testing too 😅. Nevertheless, I'll keep your suggestion in mind in case this change is necessary in the future. Thanks!

PR #112, opened 2021-08-21
[Daniel Sim] iP

danielsimre:
# Duke
> “Life is too complicated not to be orderly.” – Martha Stewart [(source)](https://dansilvestre.com/productivity-quotes)

Whenever your workload starts to pile up, have you ever felt:
- confused?
- disorganized?
- hopeless when you finally realize how much work you have?

Worry no further! Duke is a tool that aims to help you organize your work better, thereby increasing your overall efficiency when tackling your tasks.

### Why should you use Duke?
1. It's **easy to use**
2. It's **lightweight** (text-based)
3. ~~And most importantly,~~ it's **free**

### Features include (but are not limited to):
- [x] Manage tasks, deadlines and events
- [x] Save/load functionality
- [x] Get all deadlines/events happening on a certain date
- [x] Get all deadlines/events occurring within x hours/days/months from now
- [ ] Get all overdue tasks (Planned feature)

### Below is some sample code from Duke:
```java
    Path filePath = Paths.get(System.getProperty("user.dir"), "data", "tasks.txt");
    new Duke(filePath).run();
}
```

So _what_ are you waiting for? Download Duke right now! :grin:


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    private final Ui ui;
+    /**
+     * Object to handle loading from/save to a save file
+     */
+    private final Storage storage;
+    /**
+     * Object to represent the user's task list (e.g. add/delete/mark as done)
+     */
+    private TaskList tasks;
```
Harry-Gao-H: `tasks` here is an object of TaskList class instead of a collection of tasks. So perhaps it is better to name it `taskList`?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,272 @@
...
+            if (input.length() <= 5) {
+                throw new DukeException("The description of a todo cannot be empty.");
+            }
+            task = new Todo(input.substring(5));
+            break;
+        case "DEADLINE":
+            // Add Deadline task
+            String errorMessage = "Command must be in the format: [taskName] /by "
+                    + "[date(YYYY-MM-DD)] [time(HH:MM)].";
+            String[] splitInput = Parser.splitWith(input, 9, " /by ", errorMessage);
```
Harry-Gao-H: Should `splitInput` be in plural form?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,272 @@
...
+            String taskName = splitInput[0];
+            errorMessage = "Date and time must be in the format: YYYY-MM-DD HH:MM.";
+            String[] dateTime = Parser.splitWith(splitInput[1], 0, " ", errorMessage);
```
Harry-Gao-H: Should `dateTime` be in plural form?

src/main/java/duke/Duke.java
darren2pro: This is a valid point. But I also think that "tasks" is appropriate too because it shows a plural form already.

src/main/java/duke/task/TaskList.java
darren2pro: I think this is a valid point.

src/main/java/duke/task/TaskList.java
darren2pro: I think this is a valid point too.

darren2pro (commented):
Your coding standards look good! 👍

src/main/java/duke/Duke.java
danielsimre: Thank you both for the comments! I agree with Harry that taskList is an more appropriate name, as the variable refers to a singular TaskList object, so I will make the changes accordingly.

src/main/java/duke/task/TaskList.java
danielsimre: Yup, thank you for catching this. The variable has been renamed to splitInputs.

src/main/java/duke/task/TaskList.java
danielsimre: Thank you for catching this. The variable has been renamed to dateTimeInputs.

PR #113, opened 2021-08-21
[Xiao Yunpeng] iP

yunpeng1234:
# Duke

> This works
> -Source, "Dude just trust me" :man:

Here's where to [download](https://github.com/yunpeng1234/ip/releases)

This is a bullet list
* This is **line 1**
* This is *line 2*

While this is a numbered list
1. Now this is ~~line 1~~
2. Now this is line 2

- [x] Tasks
- [ ] Not Tasks

```java
void test() {
 System.out.println("Here's some syntax highlighting");
}
```
` System.out.println("Bye")`

Thanks and have a great time.


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
...
+     * Class Constructor based on where the storage file path.
+     *
+     * @param filePath where the storage file is at.
+     */
+    public Duke (String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.parseFile());
+        } catch (Exception e) {
```
koh-jx: Pokemon exception handling (Catching them all) is generally not recommended. May I suggest listing and catching specific intended exceptions? 

eg) catch (ParseException | IllegalArgumentException | FileNotFoundException e) {...}

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
+package duke;
+
+/**
+ * Class that is a chat bot that can store tasks.
+ */
+public class Duke {
+    private boolean isRunning = true;
+    private enum Commands { list, done, todo, event, deadline, delete, find }
```
koh-jx: Commands seem to be used in multiple files (Duke.java and Parser.java). May I suggest abstracting it to a Commands.java enum file? If not you may end up updating one without doing so for the other.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
...
+            for (int i = 0; i < tasks.getSize(); i++) {
+                storage.writeToFile(tasks.get(i).toHistory());
+            }
+            return true;
+        }
+        return false;
+    }
+
+    public static void main(String[] args) {
+        new Duke("./data/duke.txt").run();
```
koh-jx: It may be better to put the file location in the Storage class as it is directly related to storage, which in that case, you wouldn't have to pass it in into the constructor of Duke.

It can be initialised as a private static final string FILE_LOCATION, which makes it easier to find, debug and edit if necessary.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
...
+            if (checkBye(input)) {
+                return ui.sayFarewell();
+            }
+            try {
+                return Parser.parse(input, this.ui, this.tasks);
+            } catch (DukeException e) {
+                return ui.printErrorMessage(e);
+            }
+        } catch (Exception e) {
+            System.out.println("something wrong here");
```
koh-jx: Another case of Pokemon Exception Handling as mentioned above, which is dangerous if you try to continue running the application, as it may lead to runtime errors.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param instruction user's instruction
+     * @throws NoDescriptionError If user inputs nothing after a task command.
+     * @throws UnknownCommandError If user inputs a command that is outside the scope of the chatbot
+     */
+    public static String parse(String instruction, Ui ui, TaskList tasks) throws NoDescriptionError, UnknownCommandError {
+        String[] splitInstructions = instruction.split(" ", 2);
+        String operative = splitInstructions[0];
+        Commands command;
+        String[] temp;
```
koh-jx, line 24: String[] temp is only used in 2 cases, there is no need to initialise it here.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+        String item;
+        String date;
+        String description;
+        String keyword;
+        Task toAdd;
+        int taskPointer;
```
koh-jx, line 30: Take note of the below coding standard.

image

Variables should be private as well, looks like no need to keep them package-private.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+public class Storage {
+    private File file;
+
+    /**
+     * Class Constructor that takes in the file path of the storage file.
+     *
+     * @param filePath where the storage file is at.
+     */
+    public Storage (String filePath) {
+        File temp = new File(filePath);
```
koh-jx: temp can be declared in a smaller scope (the try-block)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+        ArrayList<Task> history = new ArrayList<Task>();
+        try {
+            FileReader fileReader = new FileReader(file);
+            BufferedReader bufferedReader = new BufferedReader(fileReader);
+            String currentLine;
+            while ((currentLine = bufferedReader.readLine()) != null) {
+                Task temp = parseTask(currentLine);
+                history.add(temp);
+            }
+        } catch (Exception e) {
```
koh-jx: There could be better exception handling here

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Function that parses the line that is stored in the storage file.
+     *
+     * @param task line in the file that represents a task.
+     * @return Task object corresponding to the line in the file.
+     */
+    public Task parseTask(String task) {
+        String[] temp = task.split(" \\| ");
+        if (temp.length == 3) {
```
koh-jx, line 60: Interesting way to find for Todo. Why not do the same as the Event? (temp[0].equalsIgnoreCase("T")) It may be better as you may be adding tasks in the future whose temp.length == 3 as well.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+            return toReturn;
+        } else if (temp[0].equalsIgnoreCase("E")) {
+            Event toReturn = new Event(temp[2], temp[3]);
+            if (temp[1].equalsIgnoreCase("1")) {
+                toReturn.markAsDone();
+            }
+            return toReturn;
+        } else {
+            Deadline toReturn = new Deadline(temp[2], temp[3]);
+            if (temp[1].equalsIgnoreCase("1")) {
```
koh-jx, line 74: "1" shouldn't need to IgnoreCase? I'm not too sure, correct me if I'm wrong hahaha

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Clears the file to avoid appending of the same tasks.
+     */
+    public void fileClear () {
+        try {
+            FileWriter fileWriter = new FileWriter(file.getAbsoluteFile(), false);
+            BufferedWriter bufferedWriter = new BufferedWriter(fileWriter);
+            bufferedWriter.close();
+        } catch (Exception e) {
```
koh-jx, line 95: Better exception handling could be done here

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param text text to be written.
+     */
+    public void writeToFile (String text) {
+        try {
+            FileWriter fileWriter = new FileWriter(file.getAbsoluteFile(), true);
+            BufferedWriter bufferedWriter = new BufferedWriter(fileWriter);
+            bufferedWriter.write(text);
+            bufferedWriter.newLine();
+            bufferedWriter.close();
+        } catch (Exception e) {
```
koh-jx: Same as above

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,60 @@
...
+    protected boolean isDone;
+
+    /**
+     * Class Constructor that creates a task based on the description.
+     *
+     * @param description
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
koh-jx, line 17: This is interesting to note. I'm personally fine with or without.
image

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+     * Returns a new Tasklist with tasks matching keyword.
+     *
+     * @param keyword word to filter tasklist by.
+     * @return Tasklist containing the word.
+     */
+    public TaskList findMatchingTasks(String keyword) {
+        ArrayList<Task> result = new ArrayList<>();
+        for (int i = 0; i < tasks.size(); i++) {
+            Task temp = tasks.get(i);
+            if (tasks.get(i).filterByKeyword(keyword)) {
```
koh-jx: Just being nitpicky here, it may be better to call temp.filterByKeyword(keyword), since you had already called tasks.get(i) earlier

koh-jx (commented):
There are a few glaring issues and there can be improvements made. Otherwise, clean and easy to understand. Comprehensive test list as well (as least more than me hahaha :P)

src/main/java/duke/MainWindow.java
```diff
@@ -0,0 +1,87 @@
...
+            fxmlLoader.load();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        this.initialize();
+    }
+
+    /**
+     * initiatises the
```
neojxuan: Is this JavaDoc a WIP?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+ * Class to represent parsing the commands
+ */
+public class Parser {
+    private enum Commands { list, done, todo, event, deadline, delete, find }
+
+
+    /**
+     * Method that parses instruction from the user.
+     *
+     * @param instruction user's instruction
```
neojxuan, line 13: Missing description of return value.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
+package duke;
+
+/**
+ * Class to represent parsing the commands
+ */
+public class Parser {
+    private enum Commands { list, done, todo, event, deadline, delete, find }
+
```
neojxuan, line 8: There seems to be an extra empty line here?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param index index of where to delete.
+     * @return the deleted task.
+     */
+    public Task delete(int index) {
+        return tasks.remove(index);
+    }
+
+    /**
+     * Gets the number of tasks in the tasklist.
+     * @return
```
neojxuan: Missing description of returned value in javadoc?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
+package duke;
+
+/**
+ * Class to represent parsing the commands
+ */
+public class Parser {
+    private enum Commands { list, done, todo, event, deadline, delete, find }
```
neojxuan: Enums contain a set of predefined constants. I think by convention, if you were to stick to using an enum for Commands, it would be something like 
```
private enum Commands { LIST, DONE, TODO, EVENT, DEADLINE, DELETE, FIND }
```

src/main/java/duke/Duke.java
neojxuan: Agreed with above. I would also recommend making Command an abstract class and the enum values children classes.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @param input The user's input.
+     * @return String that duke would reply with.
+     */
+    public String getResponse(String input) {
+        try {
+            if (checkBye(input)) {
+                return ui.sayFarewell();
+            }
+            try {
```
neojxuan: Would it be possible to avoid nested try-catch blocks?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,80 @@
...
+            temp = item.split("/by ");
+            date = temp[1];
+            description = temp[0];
+            toAdd = new Deadline(description, date);
+            tasks.addTask(toAdd);
+            return ui.addedTaskMessage(toAdd, tasks.getSize());
+        default:
+            break;
+        }
+        return "";
```
neojxuan: Can we just return "" in the default case?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,60 @@
...
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Marks a task as done.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
```
neojxuan, line 35: missing javadoc comment?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Reads the file of the storage and returns a ArrayList of Tasks from the storage.
+     *
+     * @return ArrayList of Task that is stored.
+     */
+    public ArrayList<Task> parseFile () {
+        ArrayList<Task> history = new ArrayList<Task>();
+        try {
+            FileReader fileReader = new FileReader(file);
+            BufferedReader bufferedReader = new BufferedReader(fileReader);
```
neojxuan: Not related to coding standard but remember to close the BufferedReader object.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,102 @@
...
+                ui.printErrorMessage(e);
+            }
+        }
+        ui.sayFarewell();
+    }
+
+    /**
+     * Function that checks whether the user wants to exit the program.
+     *
+     * @param instruction instruction given by the user
```
neojxuan: According to the module's Coding Standard, there should be punctuation behind each parameter description.
```
@param instruction Instruction given by the user.
```

neojxuan (commented):
I think your code follows the module's Coding Standard well. 😀 I mostly nitpicked on small errors like format of/missing JavaDoc comments and enum conventions. However, as the other reviewer pointed out, the scope of some variables could be changed.

PR #114, opened 2021-08-21
[Deng Huaiyu] iP

NoraYUuu:
# Duke Update1

As you might be wondering *why you need to read this*:

> Add description for Pull Request for easy reference. ([source](https://nus-cs2103-ay2122s1.github.io/website/schedule/week4/project.html))

That’s it.

## **functions changed**
* CRUD
1. `add`
1. `delete`
1. `list`
1. `complete`
* Error message for users
1. wrong input
1. complete absent task
1. delete absent task
* local saving for task list

```Java
public void loadFile() {
    storedList = new File(FILE_PATH);
    try {
        if (!storedList.exists()) {
            if (!storedList.getParentFile().exists()) {
                storedList.getParentFile().mkdirs();
            }
            storedList.createNewFile();
            System.out.println("Local file created.");
        }
        TaskList.loadList(storedList);
    } catch (IOException e) {
        System.out.println("error when loading file: "+ e);
    }
}
```

## **demo of functions** :star_struck:
* Task List
- [x] [T][ ] read book (a todo item)
- [x] [X] this is a complete item
- [x] [ ]this is an incomplete item

**Thanks for reading!**


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+        SEARCH("find"),
+        DONE("done");
+
+        private final String k;
+
+        KeyWord(String k) {
+            this.k = k;
+        }
+
+        public String getK() {
```
jianh0ng: Perhaps the naming could be more descriptive than K? I think it would increase readability

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,76 @@
...
+            fw.close();
+        } catch (IOException e) {
+            System.out.println(":( Error when re-writing file");
+        }
+        ArrayList<Task> arr = TaskList.getTaskList();
+        for (Task t: arr) {
+            int c = t.getStatusIcon().equals("X") ? 1 : 0;
+            String now = t.getType() + " | " + c + " | " + t.getDescription();
+            if (!t.getType().equals("T")) {
+                now += " | "+t.getTime();
```
jianh0ng: Minor issue with the missing spaces.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     * Adds a new task according to instruction
+     *
+     * @param t input from the user
+     * @throws DukeException if the input is invalid
+     */
+    public void addTask(String t) throws DukeException{
+        String[] ss = t.split(" ");
+        switch (ss[0]) {
+            case("todo"):
```
jianh0ng: Not sure about the indentation for switch case. I believe case should be 4 spaces to the left.

jianh0ng (commented):
Naming for some variables could be clearer, like "ind" and "getK", but otherwise looks good. :)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,87 @@
...
+            executeNext.execute();
+            break;
+        case 5:
+            executeNext = new AddCommand(myTasks, input);
+            executeNext.execute();
+            break;
+        default:
+            break;
+        }
+    }
```
Li-Lehao: The `processInput` function is very clear and elegant.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+public class Parser {
+    private enum KeyWord {
+        END("bye"),
+        DELETE("delete"),
+        LIST("list"),
+        GET("get"),
+        SEARCH("find"),
+        DONE("done");
+
+        private final String k;
```
Li-Lehao: Since `String k` is an attribute of `KeyWord`, maybe we need to make its name more precise?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+            return 3;
+        } else if (next.length() > 2 && next.substring(0, 3).equals(KeyWord.GET.getK())) {
+            return 4;
+        } else if (next.length() > 3 && next.substring(0, 4).equals(KeyWord.SEARCH.getK())) {
+            return 6;
+        } else {
+            return 5;
+        }
+    }
+}
```
Li-Lehao: I really like this clean and elegant `parse` function. The implementation of `parser` is always tricky but you really did a good job. Maybe we can consider returning an `enum` type to enhance the readability of the code?

src/main/java/duke/command/SearchCommand.java
```diff
@@ -0,0 +1,36 @@
...
+
+import duke.TaskList;
+import duke.Ui;
+
+/**
+ * Command class to hand search instruction from user.
+ */
+public class SearchCommand extends Command {
+    private TaskList taskList;
+    private String s;
```
Li-Lehao: Maybe we need to make its name more meaningful?

src/main/java/duke/TaskList.java
NoraYUuu: Thanks for the comment! From coding standard, the case clauses should not be indented.

src/main/java/duke/Parser.java
NoraYUuu: Would change that!

PR #116, opened 2021-08-21
[Nay Lin] iP

NayLin-H99:
### DukePro

> "All you need is the plan, the road map, and the courage to press on to your desination." - Earl Nightingale ([Source](https://www.goodreads.com/quotes/606982-all-you-need-is-the-plan-the-road-map-and#:~:text=Quotes%20%3E%20Quotable%20Quote-,%E2%80%9CAll%20you%20need%20is%20the%20plan%2C%20the%20road%20map%2C,press%20on%20to%20your%20destination.%E2%80%9D))

- text-based
- easy to learn
- **~~FAST~~** ***SUPER* FAST** to use

All you need to do is,

1. download it from [here](https://github.com/NayLin-H99/ip/releases/tag/A-Jar).
2. double-click it
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing Tasks
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too! Here's the ``HelloWorld`` code:
```java
// Your First Program

class HelloWorld {
        System.out.println("Hello, World!"); 
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+
+import duke.command.AddCommand;
+import duke.command.AddCommandType;
+import duke.command.Command;
+import duke.command.DeleteCommand;
+import duke.command.DoneCommand;
+import duke.command.ExitCommand;
+import duke.command.FindCommand;
+import duke.command.ListCommand;
+import duke.task.TaskList;
```
LZher385: I like how neat the import statements are!

LZher385 (commented):
Good job! I really struggled to find any violations 👍 . Overall, I really like how neat your code is and also the extensive  amount of tests!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.util.Scanner;
+
+/**
+ * The type Ui that handles string formatting and basic interactions with the user.
+ */
+public class Ui {
+
+    /** line separator */
+    private static final String LINE_BREAK = "\t____________________________________________________________\n";
```
jennibearduit: I like that you made the line break a constant field, it makes the parts where you use line break look so much more readable

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+    }
+
+    /**
+     * Checks if the exit command is to be executed.
+     *
+     * @return boolean.
+     */
+    private boolean checkBye() {
```
jennibearduit: According to the coding standard, boolean variables/methods should be named to sound like booleans. Perhaps `isBye()` or `isByeKeyword()` will be a better method name?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+        return this.userInput.equals("bye");
+    }
+
+    /**
+     * Checks if the list command is to be executed.
+     *
+     * @return boolean.
+     */
+    private boolean checkList() {
```
jennibearduit: Similar to the `checkBye()` comment, perhaps can consider renaming the method to `isListKeyword()`? Same for the subsequent ones.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,67 @@
...
+        result += this.description;
+        return result;
+    }
+
+    /**
+     * Returns a string to be used for saving.
+     *
+     * @return string with delimiters for saving and loading.
+     */
+    public abstract String databaseString();
```
jennibearduit: According to the coding standard, names representing methods must be verbs. Perhaps can consider naming to `generateDatabaseString()` or perhaps `toDatabaseString()` instead, similar to Java's `toString()` method?

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,88 @@
...
+            case deadline: {
+                String[] args = inputs[1].split(" /by ", 2);
+                String[] datetimeArgs = args[1].split(" ", 2);
+                if (datetimeArgs.length == 1) {
+                    newTask = new Deadline(args[0], args[1]);
+                } else {
+                    newTask = new Deadline(args[0], datetimeArgs[0], datetimeArgs[1]);
+                }
+                break;
+            }
```
jennibearduit: According to the coding standard, the format of a case block does not have curly braces, and only requires the colon. Your other switch block in `Storage.java` line 58 is correct, I would suggest following that as well.
Screenshot 2021-09-02 at 10 09 52 PM

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,95 @@
...
+        tasks.get(ind).markDone();
+    }
+
+    /**
+     * Gets a task from the storage.
+     *
+     * @param ind index of the task to get from storage.
+     * @return the task.
+     */
+    public Task get(int ind) {
```
jennibearduit: This is minor and subjective, but I think replacing `ind` with `index` will make the code more readable, and does not significantly change the length of the variable name.

jennibearduit (commented):
LGTM, code is very easy to read and understand overall!

src/main/java/duke/Parser.java
NayLin-H99: Thank you! 😃

src/main/java/duke/Ui.java
NayLin-H99: Thank you! 😃

src/main/java/duke/Parser.java
NayLin-H99: Thanks for pointing this out! I've adjusted accordingly. 👍

src/main/java/duke/command/AddCommand.java
NayLin-H99: I initially included blocks because switch case does not allow same variables to be declared in different case, and putting a block fixes that. I'll try abstracting the variables out then 😅

src/main/java/duke/task/TaskList.java
NayLin-H99: Adjusted it to a more conventional idx to make it more readable. Thanks pointing this out 👍

src/main/java/duke/task/Task.java
NayLin-H99: Done! Thanks for pointing this out as well 👍

src/main/java/duke/Parser.java
NayLin-H99: Adjusted all the methods to fit the coding standards. Thanks!

PR #128, opened 2021-08-21
[Jay Aljelo Saez Ting] iP

jayasting98:
# Duke CHADbot
> "Refers to a person of the highest standards and that is perfect at everything with no flaws, will act upon what they wish to without hesitation or regard for authority, highly respected by others" -[Urban Dictionary Entry on "Absolute Chad"](https://www.urbandictionary.com/define.php?term=Absolute%20chad)

Duke CHADbot :robot: can _**easily**_ handle your tasks, including
* To-do tasks
* Tasks with deadlines
* Tasks that occur at a particular point in time such as events

Simply do the following to start
1. [Download the JAR file.](https://github.com/jayasting98/ip/releases/tag/v0.1)
1. Run it.
1. Add your tasks.
1. Let Duke CHADbot handle your tasks

If you know programming, it is _**highly extensible**_ too! Simply fork the repo and start modifying the code as you wish. For instance, if you'd like to create more types of commands, simply add a new `CommandType` instance and extend the `Command` class. For example:
```java
public enum CommandType {
    // ...
    YOUR_COMMAND(commandName, commandDescription);
    // ...
}
```
```java
public class YourCommand extends Command {
    // ...

    // define constructor
    public YourCommand(String command) {
        super(command);
        // ...
    }

    // implement abstract methods execute, parseCommand, getCommandType
    @Override
    public void execute(TaskHandler taskHandler, Ui ui) {
        // ...
    }

    @Override
    void parseCommand(String[] tokens) {
        // ...
    }

    @Override
    CommandType getCommandType() {
        // ...
        return CommandType.YOUR_COMMAND;
    }

    // ...
}
```

Features (including planned features):
- [x] Commands
- [x] Tasks
- [ ] GUI (coming soon)
- [ ] Solves halting problem (coming ~~never~~ soon using [this library](https://youtu.be/dQw4w9WgXcQ))


src/main/java/duke/ui/DialogBox.java
```diff
@@ -0,0 +1,63 @@
...
+import javafx.scene.Node;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+import javafx.scene.text.Text;
+import javafx.scene.text.TextAlignment;
+
+/**
+ * Controller class for a single dialog box.
+ * Adapted/reused from https://se-education.org/guides/tutorials/javaFxPart4.html written by Jeffry Lum.
```
andrewbaruch: Well done on citing your adaptation! :100:

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,20 @@
...
+import javafx.application.Application;
+
+/**
+ * Main class of Duke
+ *
+ * @author Jay Aljelo Saez Ting
+ */
+public class Duke {
+    /**
+     * Initialises and runs the Duke chatbot.
```
andrewbaruch: Should this be "Initialize and runs.."? (or Initialise for British English)

andrewbaruch (commented):
HI Jay!

I absolutely love your code! It's neat, seems to be thorough at catching the unexpected input, and is definitely easy to look through! I have commented a few thing which may not be in line with the coding standard. Nevertheless, great job!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,20 @@
+package duke;
+
+import javafx.application.Application;
+
+/**
+ * Main class of Duke
```
andrewbaruch: Should this have a punctuation after it?

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,202 @@
...
+        if (listLength != 1) {
+            return String.format("Now you have %d tasks in the list.", listLength);
+        } else {
+            return "Now you have 1 task in the list.";
+        }
+    }
+
+    private void printFormattedMessage(String message) {
+        String formattedMessage = messageFormatter.getFormattedMessage(message);
+        dialogContainer.getChildren().add(DialogBox.getDukeDialog(formattedMessage, dukeImage));
```
andrewbaruch: Would this be better if it is broken down to different lines? (to improve readability)

src/main/java/duke/Duke.java
charliemoweng: yes it does look a bit odd.

src/main/java/duke/DukeChatbot.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.IOException;
+
+import duke.command.Command;
+import duke.command.CommandParser;
+import duke.command.DukeInvalidCommandException;
+import duke.storage.StorageHandler;
+import duke.task.TaskHandler;
+import duke.ui.Ui;
+import javafx.application.Application;
```
charliemoweng: maybe can separate them into groups to make it clearer?

src/main/java/duke/Duke.java
jayasting98: It is intended to be British English, though I guess this module prefers American spelling, so I could just change it. Thanks!

src/main/java/duke/Duke.java
jayasting98: Yes, you're right, I missed that. Thanks for helping me spot the mistake!

src/main/java/duke/ui/Ui.java
jayasting98: Yes, I agree. I must have overlooked that. Thank you for pointing it out!

src/main/java/duke/DukeChatbot.java
jayasting98: I agree. I normally would separate these if it was my own project and I could set my own coding style. However, CheckStyle considers separating these as an error. I believe it is because separating them is not really a part of CS2103T's Java style guide. So, I will not separate them into groups even though I agree with you and I want to do that.

PR #132, opened 2021-08-21
[Lin Fangyuan] iP

LMAOboxhack:
# Bruh 
A *chatbot* for **21st century** task management.
## Features
* Task management
  * Example:
  * [x] Ace this mod
  * [ ] Take over the world 💪 
* Add & remove different types of tasks
...and many more!
## Installation
1. Download from [here](https://github.com/LMAOboxhack/ip/releases)
2. Double-click
3. Profit 💸 
## Printing a line in Java
Create a class & add the following `main` method:
```java
public class Main {
        System.out.println("Hello world!");
    }
}
```


src/main/java/bruh/tasklist/TaskList.java
```diff
@@ -0,0 +1,128 @@
...
+        return sb.toString();
+    }
+
+    /**
+     * Returns a description of the number of tasks in the list.
+     *
+     * @return A description of the number of tasks in the list.
+     */
+    public String getTaskCountDesc() {
+        return tasks.isEmpty() ? "There are no " + (isFiltered ? "matching " : "") + "tasks."
```
felissaf: I believe it would be better to separate this into a few lines to avoid confusion!:smile: For the conditional expressions, maybe you can put it in a separate variable such as 
String plurality = (tasks.size() == 1) ? "" : "s";

src/main/java/bruh/parser/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        } else if (input.equals(STOP_SIGNAL)) {
+            return new ExitCommand();
+        } else if (input.equals("list")) {
+            return new ListTasksCommand();
+        }
+
+        String[] inputs = input.split(" ", 2);
+        String keyword = inputs[0];
+
+        switch (keyword) {
```
felissaf: I like the way you divide the different cases:smile:

src/main/java/bruh/task/LocalDateTimeOrString.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public String toString() {
+        return (dateTime != null) ? OUT_FORMATTER.format(dateTime) : dateTimeString;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        } else if (o == null) {
```
felissaf: Maybe it would be better to combine line 44 and 46 (o == null || !(o instanceof LocalDateTimeOrString)) as the return statement is the same?:smile:

src/main/java/bruh/task/Task.java
```diff
@@ -0,0 +1,85 @@
...
+import bruh.exception.InvalidArgumentException;
+
+import java.io.Serializable;
+
+/**
+ * Represents a task that can be mark as done.
+ */
+public abstract class Task implements Serializable {
+    private final String description;
+    private final char symbol;
```
felissaf: I believe constant variable should be in all uppercase?

felissaf (commented):
You have very good code quality! I'm impressed:smile:

src/main/java/bruh/command/AddTaskCommand.java
```diff
@@ -0,0 +1,29 @@
...
+import bruh.ui.Ui;
+
+/**
+ * Represents a command which adds a task to the task list.
+ */
+public class AddTaskCommand extends Command {
+    private final Task taskToAdd;
+
+    /**
+     * Constructor for a command to add a task to the task list.
```
Xiaoyunnn, line 17: According to the coding standards, I suppose the first sentence should start with something like "Instantiates" or "Constructs" instead of "Constructor...".

src/main/java/bruh/parser/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        case "todo":
+        case "deadline":
```
Xiaoyunnn: According to the coding standards,

> The explicit //Fallthrough comment should be included whenever there is a case statement without a break statement.

So perhaps, you may want to add `// Fallthrough` at the end of line 43?

src/main/java/bruh/parser/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+     * @return The command to be run by the chatbot.
+     */
+    public static Command parseInputToCommand(String input) throws BruhException {
+        if (input.isEmpty()) {
+            throw new InvalidArgumentException();
+        } else if (input.equals(STOP_SIGNAL)) {
+            return new ExitCommand();
+        } else if (input.equals("list")) {
+            return new ListTasksCommand();
+        }
```
Xiaoyunnn: Perhaps you can consider adding the `else` branch to detect errors as mentioned in the code quality section of the textbook [here](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#use-the-default-branch).

src/main/java/bruh/parser/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    /**
+     * Checks for missing arguments & throws an exception, if found.
+     *
+     * @param sections     An array consisting of the input, split into the first word and the rest of the input
+     * @param errorMessage The error message to be displayed, if an exception is thrown.
+     * @throws MissingArgumentException if a missing argument is found.
```
Xiaoyunnn: I think you missed out the upper case for "if a missing ..." according to the coding standard XD

src/main/java/bruh/parser/Parser.java
michael-lee-sk: Agree with @Xiaoyunnn on adding a //Fallthrough comment

src/main/java/bruh/task/LocalDateTimeOrString.java
michael-lee-sk: Agree with @felissaf, that you should just merge both else if statements, unless you intend to print a different statement for each case.

michael-lee-sk (commented):
Didn't have much to comment on on top of @Xiaoyunnn and @felissaf. Thought u have pretty clean code 👍

src/main/java/bruh/tasklist/TaskList.java
LMAOboxhack: Thanks for the suggestion! Agreed, it would be better to split this into multiple lines for more readability. Will fix!

src/main/java/bruh/task/LocalDateTimeOrString.java
LMAOboxhack: Thanks for spotting this! Will fix.

src/main/java/bruh/task/Task.java
LMAOboxhack: Thanks for pointing this out! For these variables, they are not exactly constants; I added the `final` keyword for immutability, to reduce complexity and hopefully catch bugs, so I'm not too sure if these should be in all uppercase.

src/main/java/bruh/command/AddTaskCommand.java
LMAOboxhack: Thanks for pointing this out! I took a look at the coding standards and this doesn't seem to be explicitly specified, so I went on to take a look at the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment) (the coding standard recommends this for topics not covered) and it seems that it is okay for the summary fragment of the Javadoc to begin with a noun phrase.

src/main/java/bruh/parser/Parser.java
LMAOboxhack: Thanks for catching this! Will fix.

src/main/java/bruh/parser/Parser.java
LMAOboxhack: Thanks for raising this up! My rationale for not adding an else branch here is that I want to allow the control to simply proceed to the subsequent `switch` statements, where I included a `default` case. I believe this will catch any inputs that do not match any of the cases & simply throw an `InvalidArgumentException`.

src/main/java/bruh/parser/Parser.java
LMAOboxhack: Thanks for pointing this out! Will fix.

src/main/java/bruh/parser/Parser.java
LMAOboxhack: Thanks for pointing this out! Will fix.

src/main/java/bruh/task/LocalDateTimeOrString.java
LMAOboxhack: Thanks for raising this! Agreed, will merge both cases.

PR #141, opened 2021-08-21
[Gokul Rajiv] iP

gok99:
Duke helps you keep track of your tasks and todos. It does the following:
- Barely works
- 

To get started with Duke:
1. Download Duke from [here](https://github.com/gok99/ip/releases/tag/v0.1)
2. ~~Wrestle issues with your `jdk`~~
3. Write your own version of Duke that actually does what you want

Features:
- [x] Bugs
- [ ] Be useful (in progress)

---

Here's a code snippet that does something:
```java
public class Main {
        main(args);
    }
}
```


src/main/java/bot/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param index index of task in list
+     * @return success boolean
+     */
+    public String[] removeTask(int index) {
+        if (index < 0 || index >= taskList.size()) {
+            return new String[0];
+        }
+        Task task = getTaskAt(index);
+        this.taskList.remove(index);
+        return new String[]{
```
zacharylwy: { should have space before

src/main/java/commands/DoneCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public String[] run(Bot bot, String[] args) {
+        int index = Integer.parseInt(args[0]) - 1;
+        if (index < 0 || index >= bot.getTaskList().get().size()) {
+            throw new InvalidTaskException(Ui.ERROR_SIGNATURE + "This task does not exist in the task list!");
+        }
+        Task task = bot.getTaskList().getTaskAt(index);
+        task.markDone();
+        return new String[]{
```
zacharylwy: { should have space before

zacharylwy (commented):
Very consistent work and style. Keep up the good work

src/main/java/commands/EventCommand.java
```diff
@@ -14,7 +14,7 @@
 public class EventCommand extends Command {
 
     @Override
-    public void run(Bot bot, String[] args) {
+    public String[] run(Bot bot, String[] args) {
         if (args[0].equals("")) {
             throw new InvalidArgumentsException(Ui.ERROR_SIGNATURE + "The description of an event cannot be empty.");
         }
```
sj1999-BIT: This two seems like mutually exclusive events, perhaps you could use if else to connect them? 
For example, you could create the splitArgs array at the start. Then if splitArgs[0].equals(""), give first message, else if splitArgs.length < 2, give second message. 

I personally recommend avoid single if statements recurring in one function, as you may missed out certain logically flaws that the current if condition could not catch, but could lead to runtime exceptions.

src/main/java/bot/Storage.java
```diff
@@ -0,0 +1,76 @@
...
+import tasks.Task;
+
+/**
+ * Handles saving and loading bot data
+ */
+public class Storage {
+
+    public static final String SAVE_FILE_PATH = "data/save.txt";
+
+    private static void write(List<String> texts) {
```
sj1999-BIT, line 25: I would recommend a comment above this function as its seems to be quite a key function for this class. Perhaps just a simple description? I had a similar logic in my own work and I had to update my write-alike function several times as I keep updating the input format.

So perhaps it would be better to have a function descriptor for it?

src/main/java/bot/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return Bot's list of tasks
+     */
+    public List<Task> get() {
+        return this.taskList;
+    }
+
+    /**
+     * Set the taskList
+     *
+     * @param taskList list of tasks
```
sj1999-BIT, line 38: I think that it would be better to avoid the setters as it would reduce the point of having a private tasklist. 
I personally prefer to only allow the private attributes be modified only via the public methods provided by the wrapper class. For example in this case I only allow the tasklist to be modified by the add and deleteTask functions, whereas the getter will create a new arraylist and copy the private tasklist over, so as to return this copy instead of the actual tasklist. 
This is prevent others from modifying the tasklist at will, which could lead to runtime exceptions

src/main/java/bot/TaskType.java
```diff
@@ -0,0 +1,25 @@
+package bot;
+
+/**
+ * Types of Tasks
+ */
+public enum TaskType {
+
+    Deadline("D"), Event("E"), Todo("T"), Blank("B");
+
```
sj1999-BIT, line 9: This is quite cool. I did not know how to exactly use the functions in enum, so my work just used them like a pointer in C class.

src/main/java/commands/EventCommand.java
gok99: Point well taken! In this case, I don't think there are any functional differences but combining the two definitely makes the mutual exclusivity more obvious.

src/main/java/bot/TaskList.java
gok99: Also a good point, was definitely being lazy with setters in various places. Some 'tell-don't-ask' seems due!

PR #149, opened 2021-08-22
[Liew Jian Hong] iP

jianh0ng:
# Duke
Duke keeps tracks of the things you have to do, key features are: 
* text based
* simple commands
* efficient

How to use:
1. Choose 1 of 3 types of tasks, Todos, Deadlines and Events
2. Add the task in
3. Mark it as done after completion or delete it

Features
- [x] Managing tasks
- [ ] Reminders (coming soon)

Source code can be found [here](https://github.com/jianh0ng/ip)

Duke uses Java, so you can use it to practice too. A sample of the `main` method to run Duke is as follows
```java
 public static void main(String[] args){
     Duke duke = new Duke();
     duke.run();
}

```
**GLHF** :smile:
> It's something that was given to me, but I have a swimsuit calender hanging on the wall of my workplace. In short, I have 12 pictures of girls in bathing suits in my office. Nothing naughty. But anyways, every time the month changes, I have to flip the page to the next girl, and somehow, that makes me feel incredibly melancholy. 'Goody bye, September girl. I guess we'll never meet again.' That's how it makes me feel. And then I *flip*, and think, 'Wow, October's not bad either!", and that makes me think of how I'm already betraying the September girl... So, what I'm trying to say here is, that a year goes by really quickly. - Hirohiko Araki


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public String[] parse(String item) {
+        String[] result = new String[]{"", "", "", "false"};
+        String[] split = item.split(" ");
+        int firstWordLength = split[0].length();
+        result[0] = split[0];
+
+        String remainder = "";
+        int separator = 0;
+        if (split.length > 1) {
```
Droffilc13: I like how the Parser code is very readable as there is no deep nesting. 😄

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Class for parsing input of user for duke.Duke.
+ * @author Liew Jian Hong
+ */
+
+public class Parser {
+
+    /**
+     * Returns the result of parsing the user input in an array of string where index 0 is command, index 1 is
```
Droffilc13, line 14: I like that this Javadoc comment explains clearly the use of arbitrary integers 0, 1, 2 in the method body.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+                    separator = i;
+                    break;
+                }
+            }
+        }
+        if (separator == 0) {
+            result[1] = remainder;
+        } else {
+            result[1] = remainder.substring(0, separator - 1);
+            result[2] = remainder.substring(separator + 4);
```
Droffilc13: Perhaps a named constant instead of ```4```?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Class for duke.TaskList, containing a list of Tasks.
+ * @author Liew Jian Hong
+ */
+
+public class TaskList {
+    /**
+     * ArrayList to store Tasks.
+     */
+    private ArrayList<Task> al;
```
Droffilc13: Perhaps ```al``` could be renamed as ```tasks```?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+    }
+
+    /**
+     * Echoes the given command.
+     * @param t Command to be given.
+     */
+    public void echo(String t) {
```
Droffilc13: Any reason why you decided to keep method ```echo``` instead of deleting it?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Starts Duke to allow for inputs.
+     */
+    public void run() {
+        this.ui.showGreetMessage();
+        this.storage.load();
+        String userInput;
+        boolean isExit = false;
```
meixuanjin: This boolean can be renamed to follow the coding standards more closely, for example hasExited.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+                return this.ui.showSearchResults(result);
+            default:
+                throw new DukeException("Invalid input. Perhaps you've made a mistake.");
+            }
+        } catch (Exception e) {
+            return e.getMessage();
+        }
+    }
+
+    public String getResponse(String input) {
```
meixuanjin: Seems to be missing javadocs for this method?

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Constructor for a Todo task.
+     *
+     * @param desc String array consisting of parsed description.
+     */
+    public Todo(String[] desc) {
+        super(desc[1], Boolean.valueOf(desc[3]));
+    }
+
+    public String toWrite() {
```
meixuanjin, line 23: Missing javadocs for this method

src/main/java/duke/Parser.java
meixuanjin: There seem to be quite a few conditions here, maybe they could be abstracted out?

meixuanjin (commented):
Everything looks good generally, and is very readable. The parts I commented were very tiny details but other than that, looks good!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+        return this.ui.showMarkDoneMessage(this.getTaskByIndex(itemNum - 1));
+    }
+
+    /**
+     * Deletes a task from the list.
+     *
+     * @param items Parsed delete command from user.
+     * @throws DukeException Exception for wrong user inputs.
+     */
+    public String deleteTask(String[] items) throws DukeException {
```
Droffilc13, line 110: Any reason why you use an if-else block instead of just an if-block for the guard clause?

Droffilc13 (commented):
Overall, the code is easy to read, well-encapsulated and logic flows nicely. There are a few parts where I suggested changes, but all are very minor details for your consideration.

PR #155, opened 2021-08-22
[Rui Yan] iP

chowRuiYan:
# Dukey
> If you have a task, I will handle it for you!
> -Dukey

Dukey is a Todo-List that helps manage your tasks. Here are some basic features:
* Add ~~task~~ different type of tasks like (Todo, Deadline and Event)
* Mark tasks as done
* Delete tasks
* Find tasks with keywords
* Display all the tasks you have

All you need to do is,
1. Download it from [here](https://github.com/chowRuiYan/ip/releases).
2. Click on **Assets**, then click on **iP.jar**.
3. Run and add your tasks.
4. I'll do the rest for you!

It is *EASY* to use and most importantly, it is **FREE**  :grin:

Future updates:
- [x] Finding tasks with keywords
- [ ] Introducing calendar (coming real soon) 
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's a the `main` method.
```Java
public class Main {
        Duke dukey = new Duke(filePath);
        dukey.run();
    }
}
```


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,258 @@
+package duke;
+
+import java.util.ArrayList;
+
+/**
+ * Deals with the Command given by user and implement the corresponding actions.
+ */
+public abstract class Command {
+    /**
+     * Abstract method meant to indicate the actions taken for different command.
```
GnohChengYi: > In method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)

Would it be better this way?

```suggestion
     * Indicates the actions taken for different command.
```

A client who uses this class does not need to know that the method is abstract. A subclass of `Command` would know it is abstract, otherwise the code will not compile.

[Here](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/swing/Action.html) is an example from Java 11 API. Notice that `getValue(String key)` is listed under abstract methods, but its Javadoc does not mention anything about it being `abstract`.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param input String representation of the input given by user.
+     * @return The corresponding Command.
+     * @throws DukeException Throw DukeException.
+     */
+    public static Command parse(String input) throws DukeException {
+        input = input.strip();
+        String[] inputs = input.split(" ");
+        String action = inputs[0];
+
+        switch (action.toLowerCase()) {
```
GnohChengYi: I like how you made the commands case insensitive 👍

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Constructor for a UI.
+     */
+    public Ui() {
+        sc = new Scanner(System.in);
+    };
+
+    /**
+     * UI when a task has been added.
```
GnohChengYi: > In method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)

Would it be better this way?
```suggestion
     * Displays UI to show a task has been added.
```

I noticed the same issue in several other places within this file.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,125 @@
...
+        System.out.println("Got it. I've added this task:\n" + "  " + task + "\n" +
+                "Now you have " + taskList.size() + " tasks in the list.");
+    }
+
+    /**
+     * UI when a task has been marked as done.
+     *
+     * @param task The Task that has been marked as done.
+     */
+    public void doneUi(TaskList.Task task) {
```
GnohChengYi: Another issue is that 

> Names representing methods must be verbs and written in camelCase.

Would it be better to rename the methods in the following way?
```suggestion
    public void displayDoneUi(TaskList.Task task) {
```

This issue applies to several other Ui methods within this file.

GnohChengYi (commented):
LGTM, almost. Maybe just a few nits to fix.

src/main/java/duke/Command.java
chowRuiYan: Many thanks for all your comments, I'll change it accordingly.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,258 @@
...
+     * @return If current command is an exit command.
+     */
+    public boolean isExit() {
+        return false;
+    }
+
+    /**
+     * Handles the adding of new task into Duke.
+     */
+    public static class AddCommand extends Command {
```
zhing22: In the code quality guidelines [here](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#guideline-name-well) under naming section, theres this line that says "Use __nouns__ for classes/variables and verbs for methods/functions". 
Perhaps it would be nice to make your class name a noun (maybe something like ```CommandAdder extends CommandImplementer```? but i'll leave that to you)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,68 @@
...
+                        "Please indicate the task number of the completed task.");
+            } else {
+                try {
+                    int index = Integer.parseInt(inputs[1]);
+
+                    return new Command.DoneCommand(index);
+                } catch (NumberFormatException e) {
+                    throw new DukeException("Invalid input! Please enter the task number after 'done'.");
+                }
+            }
```
zhing22: Perhaps it would be nice to reduce nested coding. Here I removed the else block so the try-catch block is at the same indentation as the other lines in the ```done``` block. 

```suggestion
            if (inputs.length == 1) {
                throw new DukeException("Insufficient input received! " +
                        "Please indicate the task number of the completed task.");
            } 
            
            try {
                int index = Integer.parseInt(inputs[1]);
                return new Command.DoneCommand(index);
            } catch (NumberFormatException e) {
                throw new DukeException("Invalid input! Please enter the task number after 'done'.");
            }
```

zhing22 (commented):
Nice code! I really had nothing much to comment. I like how you split your code into multiple subclasses and methods which made it really clear and easy to read! 👍

PR #156, opened 2021-08-22
[Ng Wee Han] iP

beefham:
# Duke + A bit more
Duke has been upgraded to manage your schedule. These include:
* Your things **to do** anytime
* Your ~fearsome~ **deadlines**
* and your **events** that are happening

Then you simply just:
1. Add tasks using commands like `todo` or `event`
2. Tasks show up in Duke (use `list` to view them)
3. ❓ ❔ ❓ 
4. Profit

### Features
- [x] Manage tasks
- [x] Manage deadlines
- [x] Manage events
- [x] Save to disk
- [ ] ~Play games~

### Sample code
``` Java
// please use Duke
if (!user.isUsingDuke) {
    user.setIsUsingDuke(true);
}
```


src/main/java/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,29 @@
+package commands;
+
+import java.io.IOException;
+import java.time.LocalDate;
+
+import duke.*;
+import tasks.*;
+import exceptions.*;
```
AllardQuek: Should the specific imports be used here instead?

src/main/java/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,29 @@
...
+    LocalDate deadline;
+
+    public AddDeadlineCommand(String desc, boolean isDone, LocalDate deadline) {
+        super(desc, isDone);
+        this.deadline = deadline;
+    }
+
+    @Override
+    public void execute(Ui ui, TaskList taskList, Storage storage) {
+        Deadline newDeadline = new Deadline(this.desc, this.isDone, this.deadline);
```
AllardQuek: Might be nit-picking, but you could consider not using the `this` keyword here? 

My tutor mentioned mentioned to avoid unnecessary usage of `this`, but I would still also prefer using `this` haha

src/main/java/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,29 @@
+package commands;
+
+import java.io.IOException;
+import java.time.LocalDate;
+
```
AllardQuek, line 4: I like how the different types of imports are separated here 👍

src/main/java/commands/AddEventCommand.java
```diff
@@ -0,0 +1,29 @@
+package commands;
+
+import java.io.IOException;
+
+import duke.*;
+import tasks.*;
+import exceptions.*;
+
+public class AddEventCommand extends AddTaskCommand {
+
```
AllardQuek, line 6: Not sure if this is an unnecessary blank line here?

src/main/java/commands/AddEventCommand.java
```diff
@@ -0,0 +1,29 @@
+package commands;
+
+import java.io.IOException;
+
+import duke.*;
+import tasks.*;
+import exceptions.*;
```
AllardQuek: I notice the same issue above and in other files as well

src/main/java/commands/AddTaskCommand.java
```diff
@@ -0,0 +1,36 @@
...
+        taskList.addTask(newTask);
+        ui.printAddTask(newTask);
+        try {
+            storage.writeTasksToFile(taskList, storage.getTaskFile());
+        } catch (IOException e) {
+            ui.printFileWriteFail(storage.getTaskFile());
+        }
+    }
+
+    public boolean isQuit() { return false; }
```
AllardQuek: Should the body be on multiple lines instead? I think it's a coding standard for if-else statements, but not sure about methods. Could be easier to debug when line numbers come into play 🤔

src/main/java/commands/MarkDoneCommand.java
```diff
@@ -0,0 +1,40 @@
...
+                System.out.println("Invalid index given, enter a number from 1 to " + taskList.numberOfTasks());
+            } else if (taskList.numberOfTasks() == 0) {
+                System.out.println("You cannot mark any task as done because you have no tasks!");
+            }
+        } catch (IOException e) {
+            ui.printFileWriteFail(storage.getTaskFile());
+        }
+    }
+
+    public boolean isQuit() {return false;}
```
AllardQuek: Should there be spacing between the braces and the method body?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,61 @@
...
+
+import commands.Command;
+import tasks.*;
+import exceptions.*;
+
+public class Duke {
+
+    private Ui ui;
+    private TaskList taskList;
+    private Storage store;
```
AllardQuek: Would `store` be a bit vague since `store` might have multiple meanings? (textbook page 62) 

Feel free to ignore !

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,61 @@
...
+    private boolean isRunning;
+
+    Duke() {
```
AllardQuek: LGTM!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,130 @@
+package duke;
+
+import java.io.File;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
```
AllardQuek, line 5: Perhaps the imports could be ordered alphabetically?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,130 @@
...
+     * Reads the file of the current Storage object and stores the Tasks
+     * saved in the file into a ArrayList of Tasks.
+     *
+     * @return The ArrayList of Tasks read from the file.
+     */
+    public ArrayList<Task> readFile() {
+        try (BufferedReader br = new BufferedReader(new FileReader(this.taskFile))) {
+            ArrayList<Task> taskList = new ArrayList<>();
+            String line;
+            while ((line = br.readLine()) != null && !line.equals("")) {
```
AllardQuek: Not sure if there is a clearer way to express this? Not sure if it's considered a complicated expression (textbook: page 57-58)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,104 @@
...
+        System.out.println(helpMessage);
+    }
+
+    public void printSepLine() {
+        System.out.println(SEP_LINE);
+    }
+
+    /**
+     * Prints out the message when a task has been successfully added to the TaskList in Duke.
+     * @param t The task that was added to the TaskList.
```
AllardQuek: Maybe add a new line after the header?

AllardQuek (commented):
I realised we should only comment on coding standards but some points here are more of code quality and are quite nit-picky. Looks good overall!

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,32 @@
+package tasks;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+
+    LocalDate dueDate;
```
danielsimre, line 10: I noticed that in a lot of classes, many class variables are missing their access modifier. Perhaps you should explicitly state these as private? (Unless they are meant to be package private)

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Sets the task as not done.
+     */
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+    String completionStatus() {
```
danielsimre: Should the access modifier for this function be explicitly stated? (Either public or protected, from what I see of its usage)

src/main/java/duke/DukeParser.java
```diff
@@ -0,0 +1,76 @@
...
+        } catch (NumberFormatException e) {
+            System.out.println("It seems like you have entered an invalid task number");
+            System.out.println("Please enter the task number as shown in the list.");
+            return new ErrorCommand("An exception occurred! See details above.");
+        } catch (ArrayIndexOutOfBoundsException e) {
+            return new ErrorCommand("Missing fields in command. Type 'help' for more info.");
+        } catch (Exception e) { //TODO DANGEROUS! please change
+            return new ErrorCommand("An exception occurred!");
+        }
+    }
```
danielsimre: Currently this function is about 50 lines long, so perhaps you could find a way to reduce it to about 30 lines? (One possible way would be for the caller of `parse` to contain the try-catch block to handle errors)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,63 @@
...
+import tasks.*;
+import exceptions.*;
+
+/**
+ * This class encapsulates a duke.Duke task list.
+ * It is responsible for maintaining the tasks in an instance of Duke.
+ *
+ */
+public class TaskList {
+    private ArrayList<Task> taskList;
```
danielsimre, line 15: Maybe this variable could be called `tasks` instead since this variable is a collection of tasks?

danielsimre (commented):
Overall, the quality of your code is good! I only managed to find a few minor errors, plus a few that Allard has already caught. (Explicit import statements and usage of `this` keyword)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,104 @@
+package duke;
+
+import java.io.File;
+
+import commands.*;
+import tasks.*;
```
stephanie-csy, line 6: Should import everything explicitly instead of using wildcard import

stephanie-csy (commented):
Everything looks neat and readable, cannot find any major coding standards violations. Maybe you can add javadocs for all public classes and methods so that others reading your code can easily understand what everything does at one glance. Also, consider naming your packages `duke.task`, `duke.command` etc. instead of `tasks`, `commands`.

PR #159, opened 2021-08-22
[Kevin Nathanael] iP

kevinmingtarja:
HAL9000 frees your mind by remembering the things you need to do. It's
- text-based
- easy to use
- SUPER FAST

All you need to do is,
1. download it
2. run it
3. add your tasks

And it is **FREE!**

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events

---

You can also use it to practice Java too. Here's the `main` method:

```java
        new Duke("HAL9000").initialize();
}
```


src/test/java/duke/DukeTest.java
```diff
@@ -0,0 +1,12 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
imposterble, line 3: Perhaps the position of these 2 statements should be reversed (so that the ordering is consistent with that of other files)?

src/test/java/duke/DukeTest.java
```diff
@@ -0,0 +1,12 @@
...
+
+public class DukeTest {
+    @Test
+    public void dummyTest(){
```
imposterble: Should there be whitespace before "{"?

src/test/java/duke/ui/UiTest.java
```diff
@@ -0,0 +1,39 @@
...
+    public void print_validFormat_correctString() {
+        Ui ui = new Ui();
+        ui.print("%d, %s, %.2f", 1, "two", 3.14212322);
+        assertEquals("1, two, 3.14", outputStreamCaptor.toString().trim());
+    }
+
+    @Test
+    public void print_invalidFormat_exceptionThrown() throws Exception {
+        Ui ui = new Ui();
+        assertThrows(Exception.class, () -> ui.print("%d",  "two"));
```
imposterble: Should there be only a single space to separate non-whitespace characters (e.g., "two")?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,47 @@
+package duke;
```
imposterble, line 1: I **like** how you packaged the relevant classes into sub-packages

imposterble (commented):
LGTM! Good work overall and all the best with the remaining iP tasks :smile:

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Returns true if the program should still keep going on (e.g., when input is not equal to "bye")
+     *
+     * @param input full raw user input line.
+     * @return true if input is not equal to "bye".
+     */
+    public boolean parseInput(String input) {
+        ui.printDivider();
+        try {
```
jeremy-7007, line 38: Why not try using a switch statement! It might make your code a bit more clean, but otherwise, it is fine as is.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Parses add command from the user.
+     *
+     * @param input full raw user input line.
+     * @throws DukeException if the user input is of an invalid format.
+     */
+    private void addItem(String input) throws DukeException {
+        Task newItem = null;
+        if (input.contains("todo")) {
```
jeremy-7007: Unless I am misunderstanding your code, if you leave it as `input.contains("todo")`, wouldn't there be a bug where the user types, for example, "bla bla todo homework"? The if-condition is fulfilled, but the outcome might not be as expected. Same for other else if statements below.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+        try {
+            FileReader fr = new FileReader(path);
+            BufferedReader br = new BufferedReader(fr);
+            String line = br.readLine();
+            while (line != null) {
+                String type = Character.toString(line.charAt(line.indexOf("[") + 1));
+                boolean isDone = Character.toString(line.charAt(line.indexOf("[", 2) + 1)).equals("X");
+                String name = line.substring(line.indexOf("]", line.indexOf("]") + 1) + 2);
+
+                if (type.equals("T")) {
```
jeremy-7007, line 52: A switch statement can be used here too!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            newItem = new ToDo(parsedInput[1]);
+        } else if (input.contains("deadline")) {
+            String[] parsedInput = input.split(" ", 2); // Splits input into array of [deadline, ...]
+            if (isIncomplete(parsedInput)) {
+                throw new DeadlineException();
+            }
+
+            String[] description = parsedInput[1].split("/");
```
jeremy-7007, line 83: If you want to be even more strict with the user inputs, it might be a good idea to check that the "/" is indeed followed with "by" before splitting the input into event name and time. Same for event below.

jeremy-7007 (commented):
Made some suggestions to the Parser class. Otherwise, looks good to me. Good luck for the rest of the iP!

src/main/java/duke/parser/Parser.java
rushilramesh: Agreed, a switch statement would be useful here since there are a significant number of cases.

src/main/java/duke/parser/Parser.java
rushilramesh: Agreed, perhaps it would be better if you only check the first word?

src/main/java/duke/parser/Parser.java
rushilramesh: Agreed, perhaps it would be better to also specify in your DukeException the format of acceptable user input.

rushilramesh (commented):
Code makes sense. All the best!

src/main/java/duke/Main.java
```diff
@@ -0,0 +1,32 @@
...
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private final Duke duke = new Duke();
+
+    @Override
+    public void start(Stage stage) {
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
```
kdeasymoneysniper, line 22: maybe separate into multiple lines :-)

src/main/java/duke/parser/Parser.java
kdeasymoneysniper: Yup, a switch statement would make this part cleaner, although I would say it is still very readable the way it is now!

kdeasymoneysniper (commented):
Your code is really well thought out and the OOP demonstrated is pretty sick! :-)

src/main/java/duke/parser/Parser.java
kdeasymoneysniper: Yeah, theres a .startsWith(<string>) method for String classes in Java!

kdeasymoneysniper (commented):
Your code is really well thought out and the OOP demonstrated is pretty sick! :-)

PR #160, opened 2021-08-22
[Keerthi Vihashini] iP

kvihashini:
# 🌊  dory 

> "just keep swimming" - dory 

dory helps you to remember everything so that you don't have to! 

## 🦭  what's there to sea-anemonemonee 
 - add tasks 
 - mark tasks as done 
 - delete tasks 
 - find tasks
 - list tasks 

## ⚓  where is it? p. sherman 42 wallaby way?
 1. download it from [*here*](https://github.com/kvihashini/ip/releases/tag/A-Jar)
 2. double-click it
 3. have fun!

## 🐟  fish are friends, not food

`showIntro()` is probably the cutest method of the entire code: 

```javascript
 String fish = "                              ....\n"
         + "                             /.... \\\n"
         + "   hi my name is dory    .-`\\ \\   `...')\n"
         + "   and i can help you   ( o   | |      (\n"
         + "    remember things      '-_ / /_./``-._)\n"
         + "                             '\\___\\";
 System.out.println(fish);
```

## ✨  anything left? 

 - [X] adding tasks
 - [X] adding deadlines
 - [X] adding events 
 - [X] deleting everything
 - [ ] pretty GUI


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class Duke {
+
+    private Storage storage;
+    private TaskList taskList;
+    private Ui ui;
+
+    /**
+     * Constructor for the class.
+     * @param filePath filePath for the data file.
```
bharathcs: Maybe there should be an empty line between description and parameter section? I think thats what they mean [here](https://se-education.org/guides/conventions/java/intermediate.html#comments)

I noticed this in the other javadocs too, so remember to fix those too if this comment makes sense.

```suggestion
     * Constructor for the class.
     *
     * @param filePath filePath for the data file.
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,71 @@
...
+                ui.showError(e);
+            } finally {
+                ui.endOfResponse();
+            }
+        }
+    }
+
+    /**
+     * Generates response for user's input.
+     * @param input
```
bharathcs: I'm not sure, but maybe its better for javadocs to include a description for input too, to help ppl immediately grasp the kind of parameters you are looking for? (e.g. input user command typed in console) I'm looking [here](https://se-education.org/guides/conventions/java/intermediate.html#comments)

I noticed this in the other javadocs too, so remember to fix those too if this comment makes sense.

```suggestion
     * Generates response for user's input.
     *
     * @param input user command typed in console
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+        } else if (nextInput.contains("find")) {
+            String inputToMatch = nextInput.replace("find", "").trim();
+            return new FindCommand(inputToMatch);
+
+        } else if (nextInput.contains("bye")) {
+            return new ExitCommand();
+
+        } else if (nextInput.contains("todo") || nextInput.contains("event") || nextInput.contains("deadline")) {
+            return new AddCommand(nextInput);
+
```
bharathcs: Might be neater if you remove the newline before every `else if`?

( Not sure myself, looking at (this)[https://se-education.org/guides/conventions/java/intermediate.html#layout] )

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Deals with loading tasks from the data file
+ * and saving tasks in the data file.
+ */
+public class Storage {
+
+    private String filePath;
+
+    public Storage(String filePath) {
```
bharathcs, line 29: Should we add a description for this constructor (only asking because it has parameters)? I think its possible `filePath` could be explained in javadocs so that it can be used easily.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,190 @@
...
+
+        // introduction to chat bot
+        System.out.println(fish);
+        System.out.println(dory);
+    }
+
+    /**
+     * Displays deleted task.
+     * @param removed Deleted Task.
+     * @param taskList TaskList containing remaining tasks.
```
bharathcs: Really nitpicking here (code too clean haha), but i think these descriptions don't need capitalisations if its not a name, its probably just deleted task right?

src/main/java/duke/exceptions/DukeException.java
```diff
@@ -0,0 +1,10 @@
+package duke.exceptions;
+
+/**
+ * Duke Exception.
+ */
+public class DukeException extends Exception {
```
bharathcs, line 6: Since this is a parameter to a public function is this a large scope (can't use short forms)? here
(And for its subclasses)

src/main/java/duke/tasktypes/Task.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+    public boolean containsMatch(String input) {
+        return this.description.contains(input);
+    }
+
+    /**
+     * Returns in a format to save to file.
+     */
+    public String hardDiskSave() {
```
bharathcs: I think this needs to be renamed to match the CS2103t function names (starting with a verb). (and also in the other `.../tasktypes/*` files)

bharathcs (commented):
LGTM! Its really hard to find errors, good job!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Returns the Command to be executed after reading
+     * in user's input.
+     * @param fullCommand User's input.
+     * @return Command.
+     */
+    public static Command parse(String fullCommand) {
+        String nextInput = fullCommand.toLowerCase();
+        if (nextInput.contains("delete")) {
```
zhenxuantan: I think that Switch statements would make the code more readable as compared to multiple if else statements. What do you think?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+            if (doesFileExist) {
+                try {
+                    List<String> data = Files.readAllLines(dataFilePath);
+                    for (int i = 0; i < data.size(); i++) {
+                        String[] lineTask = data.get(i).split("/");
+                        String taskType = lineTask[0];
+                        String isDone = lineTask[1];
+                        String taskDesc = lineTask[2];
+
+                        boolean doneOrNot = (isDone.equals("true"));
```
zhenxuantan: Wouldn't this be better ```String isDone = lineTask[1].equals("true");``` instead of delaying the evaluation to doneOrNot?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+                        if (taskType.equals("T")) {
+                            ToDos toDo = new ToDos(taskDesc);
+                            toDo.changeStatus(doneOrNot);
+                            dataForDory.add(toDo);
+                        } else if (taskType.equals("D")) {
+                            String finishBy = lineTask[3];
```
zhenxuantan: Wouldn't "date" as variable name be better? What do you think?

src/main/java/duke/tasktypes/Deadlines.java
```diff
@@ -0,0 +1,50 @@
+package duke.tasktypes;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Deadline class.
+ */
+public class Deadlines extends Task {
```
zhenxuantan: I didn't like the fact that Task is a singular noun but the child class extending from it is plural. Do you have any rationale for naming these classes in their plural form?

src/main/java/duke/tasktypes/Deadlines.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Returns in a format to save to file.
+     */
+    @Override
+    public String hardDiskSave() {
+        return "D/" + this.getBooleanStatus() + "/" + this.getDescription() + "/" + this.by;
+    }
+
+
```
zhenxuantan: Is the empty line deliberate?

zhenxuantan (commented):
Overall, I feel that the code is clean enough, but I think it can be made more readable! 😄

PR #164, opened 2021-08-22
[Ng Xiang Jun] iP

xiangjunn:
- text-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use

All you need to do is,
1. download it from [here](https://github.com/nus-cs2103-AY2122S1/ip/pull/164).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😁 

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+                path = Files.createDirectory(path);
+            }
+
+            File file = path.resolve("duke.txt").toFile();
+            this.file = file;
+            if (!file.exists()) {
+                file.createNewFile();
+            }
+        } catch (IOException e) {
+            // this exception should not occur because the input(path) is fixed
```
YoYoCiti: May still be safer to handle this exception since there may be unforeseen issues with the file on another system?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,158 @@
...
+public class Ui {
+    private Scanner sc;
+
+    /**
+     * Constructs an Ui instance. Instantiate a scanner object to take in user's input.
+     */
+    public Ui() {
+        this.sc = new Scanner(System.in);
+    }
+    private void printMessage(String message) {
```
YoYoCiti:
```suggestion
    
    private void printMessage(String message) {
```

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Convert the deadline to a string that can be saved to a file and converted back to itself.
+     *
+     * @return The string to be stored.
+     */
+    @Override
+    public String stringToStore() {
+        return Deadline.TASK_LETTER + " | " + this.getStatusIcon() + " | " + this.description + " | "
+                + this.date.format(Deadline.DATE_TIME_FORMATTER) + " | " + this.time + "\n";
```
YoYoCiti:
```suggestion
        return String.format("%s | %s | %s | %s | %s \n", Deadline.TASK_LETTER, this.getStatusIcon(), 
                this.description, this.date.format(Deadline.DATE_TIME_FORMATTER), this.time);
```

Could consider for better readability?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+                throw new IncompleteTaskDescriptionException("event");
+            }
+        default:
+            // checked for command validity in duke.Parser class, so this should not execute at all
+            throw new InvalidCommandException();
+        }
+        tasks.add(task);
+        this.ui.showAddTask(task, this.tasks.size());
+        return this.tasks;
+    }
```
YoYoCiti: This method is pretty long. Maybe it would be cleaner to catch and throw the exceptions in the different Task classes (i.e Event, Deadline etc.) as the responsibility to do so seems to be more of theirs rather than the TaskList.

YoYoCiti (commented):
Overall, easy to read and good quality code :smile: Only a few nitpicks. Thanks.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+import duke.task.Task;
+import duke.task.TaskList;
+
+/**
+ * Deals with making sense of the user command and executing it.
+ */
+public class Parser {
+
+    /**
+     * Execute user command and returns a boolean that represents the continuation of the program.
```
ganhongyao: Should the comment start with "Executes" instead of "Execute"? I noticed this in several other places too.
```suggestion
     * Executes user command and returns a boolean that represents the continuation of the program.
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,93 @@
...
+                break;
+            } else {
+                throw new InvalidCommandException();
+            }
+        case FIND:
+            tl.find(restOfCommand);
+            break;
+        default:
+            /* will never be executed because the error would have been caught in run() method
+               if the user input a command that is invalid */
```
ganhongyao: I like how you explicitly state this behaviour which may not be obvious at first glance 👍

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+        try {
+            String dir = System.getProperty("user.dir");
+            Path path = Paths.get(dir, "data");
+
+            // create the data folder if it does not exist
+            if (!Files.exists(path)) {
+                path = Files.createDirectory(path);
+            }
+
+            File file = path.resolve("duke.txt").toFile();
```
ganhongyao: I like how you separate the different logical units using line breaks. 👍

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Add the task to the task list.
+     *
+     * @param description The description of the task.
+     * @param keyword The command keyword to identify the type of task to be added.
+     * @return The updated task list after the task is added.
+     * @throws DukeException If the task fails to be added to the list.
+     */
+    public ArrayList<Task> addTask(String description, CommandKeyword keyword) throws DukeException {
+        Task task = null;
```
ganhongyao: Would it be better if you do the declaration and initialisation together in the respective case blocks, and doing `tasks.add(task)` in the case block as well? I think that this has the benefit of ensuring that the task will never be `null`, especially if the logic is changed in some way in future.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+    private Ui ui;
+
+    /**
+     * Constructor for TaskList.
+     *
+     * @param tasksInfo Arraylist of tasks loaded from storage.
+     * @param ui An Ui instance to deal with interactions with the user.
+     */
+    public TaskList(ArrayList<String> tasksInfo, Ui ui) {
+        List<Task> list = tasksInfo.stream().map(Task::convertToTask).collect(Collectors.toList());
```
ganhongyao, line 27: Would this be a more descriptive variable name?
```suggestion
        List<Task> tasks = tasksInfo.stream().map(Task::convertToTask).collect(Collectors.toList());
```

ganhongyao (commented):
Overall, your code looks well-organised and easy to follow! 👍 Added some minor comments.

PR #171, opened 2021-08-22
[Liu Zhilan] iP

LiuZhi1an:
Duke Task Manager

- text-based
- easy to learn
- ~~fast~~ _super fast_ to use

All you need to do is,

1. download it from [here](https://github.com/LiuZhi1an/ip.git).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😏

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing todos
- [ ] Reminders (coming soon)

```
}
```


src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,12 @@
+package duke;
+public class DukeException extends Exception{
+    protected String message;
+
+    /** Constructs a new duke exception with the specified detail message.
+     *
```
trash-bin99: Most likely an accident, the JavaDocs description should be 1 line after the initial /** .

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Parses user input and responds with corresponding commands.
+     * 
+     * @param userInput input by user
+     * @return corresponding command of the input
+     * @throws DukeException
+     */
+    public static Command parse(String userInput) throws DukeException {
+        int firstBlankIndex = userInput.indexOf(" ");
+        String commandWord = firstBlankIndex == -1 ? userInput : userInput.substring(0, firstBlankIndex);
```
trash-bin99: This format is completely valid and compact. I think it would be easier for others to understand your code
if you expanded this statement, to something like:

```

String commandWord;
if (firstBlankIndex == -1) {
    commandWord = userInput;
} else {
    commandWord = userInput.substring(0, firstBlankIndex);
}
```

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,89 @@
+package duke.tasks;
+// import java.sql.Date;
+// import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task{
```
trash-bin99: According to the coding standards, we need descriptive header comments for all public classes as well as their methods. I noticed this with other public classses as well.

trash-bin99 (commented):
Hello!! After looking through your code, I found very little coding standard violations, and I also liked how everything was neatly packaged into different classes (e.g. the Command subclasses). I do think that adding some comments (especially for more complicated parts such as your parse() method) would greatly help both you and others reading your code to understand and spot potential errors. All in all, great job!

src/main/java/duke/Parser.java
BananaTechs: I think this format would also have a similar effect.
```
String commandWord = firstBlankIndex == -1
    ? userInput
    : userInput.substring(0, firstBlankIndex);
```

src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Adds the task to the taskslist.
+     *
+     * @param task the task to be added
+     */
+    private String addTask(Task task, TaskList tasks) {
+        String message1 = ("Got it. I've added this task: \n");
+        String message2 = task.toString();
+        tasks.getTaskList().add(task);
```
BananaTechs, line 61: Should a method like add() could be added to the TaskList class instead of asking it for the ArrayList inside each time? It would better follow the tell-don't-ask principle.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    /**
+     * Constructs a new empty task list
+     */
+    public TaskList() {
+        this.taskList = new ArrayList<>();
+    }
+
+    public ArrayList<Task> getTaskList() {
```
BananaTechs, line 27: Perhaps this class could have additional functionality to operate on the ArrayList and avoid having to return it?

src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param storage given storage object
+     * @throws IOException
+     * @throws DukeException
+     */
+    public String execute(TaskList tasks, MainWindow gui, Storage storage) throws IOException, DukeException {
+        if (this.task instanceof Todo) {
+            storage.saveTaskToFile(task.toString() + "\n");
+            return addTask(this.task, tasks);
+        } else if (this.task instanceof Deadline) {
+            Deadline deadline = (Deadline) this.task;
```
BananaTechs, line 42: Is this casting needed? this.task remains as type task, and when it is passed to addTask() it is treated of type Task.

src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,69 @@
...
+            return addTask(event, tasks);
+        }
+    };
+
+    /**
+     * Adds the task to the taskslist.
+     *
+     * @param task the task to be added
+     */
+    private String addTask(Task task, TaskList tasks) {
```
BananaTechs, line 58: Perhaps this could use this.task instead of taking in a Task argument?

BananaTechs (commented):
I like how neat and organised your code is! I just suggested some minor optimisations to the TaskList and AddCommand classes.

PR #172, opened 2021-08-22
[Ng Wei En] iP

wei2912:
# Duke

> Quote text here

## Lorem ipsum blah blah

Here's a *bullet* list:
* abc1
* def2
* ghi3

Here's a **numbered** list:
1. foo
2. bar
3. foobar

```java
public class Main {
        System.out.println("hello world!");
    }
}
```

Todo:
- [ ] Feature 1
- [ ] Feature 2
- [ ] Feature 3

Reference to `duke`: https://github.com/nus-cs2103-AY2122S1/ip :P


src/main/java/duke/TaskType.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final String taskIcon;
+    private TaskType(String taskIcon) {
+        this.taskIcon = taskIcon;
+    }
+
+    /**
+     * Getter for task icon.
+     * @return Task icon corresponding to type of task.
+     */
```
sunjc826, line 24: I'm not so sure if documentation for getters is necessary. To compare it with your `TaskList` class, there are trivial methods there that are undocumented. Hence, to be consistent, I suggest, either adding documentation to those methods as well, or removing documentation for methods whose purpose is clear from the name.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,153 @@
...
+            "Hello from\n" + LOGO + "What can I do for you?";
+    private static final String BYE_MSG = "Bye. Hope to see you again soon!";
+
+    /**
+     * Renders a task into a string.
+     * @param task Task.
+     * @return Input task rendered as a string.
+     */
+    public static String renderTask(Task task) {
+        var taskType = TaskType.identifyTask(task);
```
sunjc826: Interesting use of `var`. :) (and elsewhere in your code)
To give another perspective of this however, without an ide, say I'm viewing the codebase on github, I think it would be harder to tell what is the type of the variable. Hence, I'm not so sure if this is the "Java" way. I suppose this is still way better than reading raw code from many dynamically typed languages out there.
Regardless, viewing code without an ide probably isn't a major usecase these days so this shouldn't be an issue.

src/main/java/duke/TaskType.java
```diff
@@ -0,0 +1,56 @@
...
+    public String getTaskIcon() {
+        return this.taskIcon;
+    }
+
+    /**
+     * Identify type of task.
+     * @param task Task.
+     * @return Type of input task.
+     */
+    public static TaskType identifyTask(Task task) {
```
sunjc826, line 34: Interesting decoupling of domain logic and cli presentation.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+import java.time.LocalDate;
+
+/**
+ * Class for Deadline tasks.
+ */
+public class Deadline extends Task {
+    private LocalDate deadline;
+
+    /**
+     * Constructor for Deadline object.
```
sunjc826:
```suggestion
     * Constructor for Deadline object.
 
```
And for other methods as well. There should be a newline between description and params.

sunjc826 (commented):
Excellent work! Other than some newlines here and there, there aren't many nitpicks I can find.

Also, this is a first I've seen the use of var in a Java project. Very interesting.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+        } else if (command.equals("find") && parameters.length >= 2) {
+            String query = input.replace("find", "").strip();
+            Ui.printFoundTasks(tasks.find(query));
+        } else {
+            throw new Exception(
+                "I'm sorry, but I don't know what that means :-("
+            );
+        }
+        return false;
+    }
```
feliciaivane, line 98: I like how there is no deep nesting present in this method, but it is rather long. Would it be possible to split this method up into several shorter ones? :D

src/main/java/duke/TaskType.java
feliciaivane: I suggest adding documentation to the methods in the other classes as well 😄

src/main/java/duke/Deadline.java
feliciaivane: Yes, I agree with this.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,153 @@
...
+        printBanner(
+            Stream.concat(
+                Stream.of("Here are the tasks in your list:"),
+                IntStream.range(0, tasks.size())
+                    .mapToObj(i -> String.format(
+                        "%d. %s",
+                        i + 1,
+                        renderTask(tasks.get(i)))
+                    )
+            ).toArray(String[]::new)
```
feliciaivane: I think this method body is a bit confusing, would it be better to split it up into several intermediate steps instead? :D

feliciaivane (commented):
LGTM, I feel that overall the code is quite easy to read :D Good work!

src/main/java/duke/TaskType.java
wei2912: I'll work on that in future iterations after I reorganise my code for the GUI, thanks for the suggestions!

src/main/java/duke/Ui.java
wei2912: I'm not sure if `var` is permissible nowadays and the points you mentioned are certainly valid. I don't think it's a coding standard violation and I still can't find any standards that recommend for/against it, so I figured I'd just use it whenever I can to avoid having to put the types in 😂 Let me know if you find out more!

src/main/java/duke/Deadline.java
wei2912: Will note that when reworking my documentation, thanks for pointing it out!

src/main/java/duke/Parser.java
wei2912: Good point! I plan to add some additional functions to reduce the amount of repetition, and perhaps create `Command` classes for this purpose.

src/main/java/duke/Ui.java
wei2912: This makes sense, I'll create an intermediate variable for constructing the array of task strings. Thanks for the suggestion!

PR #174, opened 2021-08-22
[Tanishq Sharma] iP

Tanishq4331:
Getting overwhelmed by things you need to do? Fret not - Duke is here for you. It's,
- **here for you** (can't say this enough)
- smart
- understanding 
- supportive 

Here's how you summon `this` (Duke's preferred pronoun - although `this` doesn't mind being called Duke): 

1. download `this` from [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ).
2. double-click `this`.
3. add your tasks.
4. let `this` manage your tasks for you.

You don't need features - all you need is Duke. But if you insist `this` also has features!!!
  - [x] Managing tasks
  - [x] Managing deadlines
  - [x] Managing events
  - [ ] Life advice ~~if you ask nicely~~

If you are a Java programmer, you can use Duke to practice Java too. Here's the `main` method:
```java
        String filePath = System.getProperty("user.dir") + File.separator + "tasks.txt";
        new Duke(filePath).run();
    }
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Starts the assistant.
+     */
+    public void run() {
+        String input = ui.getNextInput();
+
+        while (true) {
+            try {
+                boolean keepParsing = parser.parse(input);
```
kevin9foong: Here's a minor nit, but boolean variable names should be named to sound like booleans. Perhaps something like `shouldKeepParsing` may be better?

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+                break;
+        }
+        storage.write(taskList);
+        return true;
+    }
+
+    private Task selectTask(String taskNumString) {
+        try {
+            int taskNum = Integer.parseInt(taskNumString);
+            Task inFocus = taskList.getTask(taskNum);
```
kevin9foong: perhaps this variable `inFocus` could be changed to `taskInFocus` so that it'll be easier to remember what this variable name is when used later on.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,92 @@
+package duke.task;
+
+import duke.main.DukeException;
+
+import java.time.LocalDate;
+import java.util.Objects;
+
+public class Event extends Task {
```
kevin9foong, line 12: it might be a good idea to add a header javadoc comment for public classes

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,98 @@
...
+        System.out.println("\t Nice! I've marked this task as done:");
+        System.out.println("\t\t " + this + "\n");
+    }
+
+    /**
+     * Generates a String for storing the Task.
+     *
+     * @return String for storing the Task.
+     */
+    abstract public String storageString();
```
kevin9foong: Based on the coding guidelines, this method name could be further improved to use a verb as the name instead. Maybe something like `generateStorageString`?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,157 @@
...
+     *
+     * @param query for matching Tasks.
+     * @return List of matching Tasks.
+     */
+    public List<Task> find(String query) {
+        if (query.isEmpty()) {
+            throw new DukeException("\tYou haven't specified any keywords\n");
+        }
+        String[] keywords = query.split(" ");
+        List<Task> matches = new ArrayList<>();
```
kevin9foong, line 89: I like how you used plural form for variable names which represent multiple values.

kevin9foong (commented):
Hi Tarnishq, I've been assigned to make a PR review on code quality guidelines for your code. Hopefully my comments will be useful! 😃

src/main/java/duke/Duke.java
Tanishq4331: You're right, it should follow the boolean naming scheme.

src/main/java/duke/main/Parser.java
Tanishq4331: Thanks for the suggestion. I'll make the change.

src/main/java/duke/task/Event.java
Tanishq4331: Right, I totally missed the javadoc.

src/main/java/duke/task/Task.java
Tanishq4331: I must have missed this. Thanks for pointing out the violation.

src/main/java/duke/main/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        this.taskList = taskList;
+    }
+
+    /**
+     * Parses user inputs.
+     *
+     * @param input String from user.
+     * @return true if input is not "bye", else false.
+     */
+    public boolean parse(String input) {
```
IrvingHe000: Should the name of the method be named like a boolean? Because this method returns a boolean variable :)

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,157 @@
...
+        System.out.println("\t \t " + task);
+        System.out.println(taskSummary());
+    }
+
+    /**
+     * Generate a summary of the TaskList.
+     *
+     * @return String summary of TaskList.
+     */
+    public String taskSummary() {
```
IrvingHe000: Perhaps it's better to change the method name to getTaskSummary since it's a method?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,157 @@
...
+        System.out.println(taskSummary());
+    }
+
+    /**
+     * Finds tasks that match any keyword.
+     *
+     * @param query for matching Tasks.
+     * @return List of matching Tasks.
+     */
+    public List<Task> find(String query) {
```
IrvingHe000: Maybe can change find to a more meaningful name?

src/main/java/duke/main/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+    public Storage(String filePath) {
+        this.filePath = Path.of(filePath);
+    }
+
+    /**
+     * Loads tasks from text storage.
+     *
+     * @return Loaded TaskList.
+     */
+    public TaskList load() {
```
IrvingHe000: Maybe is better to change to a more meaningful name such as loadTaskList?

src/main/java/duke/main/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+            throw new DukeException("\t☹ OOPS!!! Your tasks might be corrupted.");
+        }
+    }
+
+    /**
+     * Write taskList to Storage
+     *
+     * @param tasklist that is written to Storage
+     */
+    public void write(TaskList tasklist) {
```
IrvingHe000: The same as above.

IrvingHe000 (commented):
Overall, I think your code is well written and easy to read 👍. Just some nits related to naming standards can be improved :)

src/main/java/duke/main/Storage.java
Tanishq4331: Thanks for the suggestions. I'll make these method names more descriptive.

src/main/java/duke/task/TaskList.java
Tanishq4331: Thanks for pointing this out. I totally missed it.

src/main/java/duke/main/Parser.java
Tanishq4331: Good catch! I've made the change.

PR #177, opened 2021-08-22
[Felix Ong] iP

felix-ong:
# Duke
> "The best task management tool."

Duke is a tool designed to manage your tasks. It's:
* text-based
* fast
* easy to use
* **FREE**

To start using Duke,
1. Download it from [here](https://github.com/felix-ong/ip)
2. Run the jar file
3. Start typing in commands! 😄 

Features:
- [x] Managing todos, events and deadlines
- [ ] Reminders

Duke can also be used for learning Java. Here's the `main` method.
```java
}
```


src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class AddCommand extends Command {
+    private Task task;
+    private String taskCountMessage;
+
+    public AddCommand(Task task) {
+        this.task = task;
+    }
+    /**
+     * Executes the specific actions for this command.
```
xiangjunn:
```suggestion
        this.task = task;
    }
    
    /**
     * Executes the specific actions for this command.
```

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Returns false.
+     *
+     * @return false
+     */
+    @Override
+    public boolean isExit() {
+        return false;
+    }
```
xiangjunn, line 42: Maybe you can provide a description of how this method works? I can't really understand why isExit() should return false. Also maybe you can name it shouldExit() instead of isExit()?

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    public DeleteCommand(String index) {
+        this.index = index;
+    }
+
+    /**
+     * Executes the specific actions for this command.
+     *
+     * @param tasks   Handles the list of tasks.
+     * @param storage Handles the saving and loading of tasks.
+     */
```
xiangjunn: Perhaps you could stick to the same style for the javadoc comments of the param?
```suggestion
     * Executes the specific actions for this command.
     *
     * @param tasks Handles the list of tasks.
     * @param storage Handles the saving and loading of tasks.
     */
```

src/main/java/controller/MainWindow.java
```diff
@@ -0,0 +1,52 @@
...
+import duke.Duke;
+import javafx.fxml.FXML;
+import javafx.scene.control.Button;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.TextField;
+import javafx.scene.image.Image;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.VBox;
+/**
+ * Controller for MainWindow. Provides the layout for the other controls.
```
xiangjunn:
```suggestion
import javafx.scene.layout.AnchorPane;
import javafx.scene.layout.VBox;

/**
 * Controller for MainWindow. Provides the layout for the other controls.
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param filePath Relative path to the storage location.
+     */
+    public Duke(String filePath) {
+        this.storage = new Storage(filePath);
+        try {
+            this.tasks = new TaskList(this.storage.loadData());
+        } catch (DukeException e) {
+            this.tasks = new TaskList();
+        }
+    }
```
xiangjunn: There seems to be a duplicate of codes here. Just a recommendation, perhaps you can create a method for this? How about this?
```suggestion
    public Duke() {
        this.setUp("data/tasks.txt);
    }
    
    public Duke(String filePath) {
        this.setUp(filePath);
    }
    
    private void setUp(String filePath) {
        this.storage = new Storage(filePath);
        try {
            this.tasks = new TaskList(this.storage.loadData());
        } catch (DukeException e) {
            this.tasks = new TaskList();
        }
    }
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+public class Parser {
+
+    /**
+     * Parses the user input to determine the command.
+     *
+     * @param userInput Command and description entered by the user.
+     * @return Command to be executed.
+     * @throws DukeException If missing arguments for commands or unknown commands.
+     */
+    public static Command parse(String userInput) throws DukeException {
```
xiangjunn, line 32: In my opinion, this method seems to be a little too long. If possible, maybe you can try to divide it into more methods.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,116 @@
...
+        this.tasks = new ArrayList<>();
+    }
+
+    /**
+     * Returns the size of the list of tasks with the appropriate grammar.
+     * @return A String describing the size of the list of tasks with the appropriate grammar.
+     */
+    public String getTaskCount() {
+        int taskCount = this.tasks.size();
+        return String.format("You have %d %s in the list.%n", taskCount, taskCount > 1 ? "tasks" : "task");
```
xiangjunn: Would it be less readable this way? Maybe you can try this?
```suggestion
        int taskCount = this.tasks.size();
        String name = taskCount > 1 ? "tasks" : "task";
        return String.format("You have %d %s in the list.%n", taskCount, name);
```

xiangjunn (commented):
LGTM. Overall, the code is neat and easy to understand. I like how the variable and method names are well-thought of and there is not much ambiguity. Most of the code follow the coding standard. Just some minor issues which are not significant and easy to fix. Good Job! 😄

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public void execute(TaskList tasks, Storage storage) throws DukeException {
+        this.task = tasks.deleteTask(storage, this.index);
```
ganhongyao, line 30: Should we avoid the unnecessary use of `this` as mentioned [here](https://se-education.org/guides/conventions/java/index.html#variables)?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd kkmm");
+
+        switch (command) {
+        case "bye":
+            return new ExitCommand();
+        case "todo":
+            if (partsLength < 2) {
+                throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+            }
+            Task todo = new Todo(parts[1]);
```
ganhongyao: Is there a reason why you're declaring a `Task` here instead of a `Todo`? Same for the other cases.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+        String command = parts[0];
+        int partsLength = parts.length;
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd kkmm");
+
+        switch (command) {
+        case "bye":
+            return new ExitCommand();
+        case "todo":
+            if (partsLength < 2) {
+                throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
```
ganhongyao: Would it make sense to create a subclass of `DukeException` that is meant for invalid descriptions/arguments (more general)? This would help to reduce the seemingly repeated error messages.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+                    + "\nPlease enter one of the following commands:\n todo <task>"
+                    + "\n deadline <task> /by <deadline(in yyyy-MM-dd kkmm format)>"
+                    + "\n event <event> /at <date(in yyyy-MM-dd kkmm format)>"
+                    + "\n list\n bye(to quit)");
+        }
+    }
+
+    private static String getDescription(String[] parts, int partsLength) {
+        return String.join(" ", Arrays.copyOfRange(parts, 1, partsLength - 3));
+    }
```
ganhongyao: Even though this is a private method, would it be good to add in some documentation for this method? Since it may not be obvious what it does at first glance.

src/main/java/duke/TaskList.java
ganhongyao: I agree with @xiangjunn on this, though maybe a better variable name can be used in place of `name` (I can't think of a better name too, I used "pluralised" for mine 🤣 )

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Returns a list of tasks with description containing the keyword.
+     *
+     * @param keyword Word to search for in task description.
+     * @return List of tasks with description containing the keyword.
+     */
+    public List<Task> findTask(String keyword) {
+        List<Task> matches = new ArrayList<>();
+        for (Task task : this.tasks) {
+            if (task.getDescription().contains(keyword)) {
```
ganhongyao: Not code quality related, but would it be better to do a case-insensitive search?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Returns a string representation of the task to be saved in.
+     *
+     * @return The string representation of the task.
+     */
+    @Override
+    public String toSaveString() {
+        return "D" + super.toSaveString() + String.format(",%s", this.by);
```
ganhongyao, line 43: Since you're already using `String.format`, would it be better to just use it for the entire string instead of concatenation?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Returns the string representation of a Deadline Task.
+     *
+     * @return The string representation of a Deadline Task.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: "
+                + this.by.format(DateTimeFormatter.ofPattern("dd MMM yyyy hh:mma")) + ")";
```
ganhongyao, line 54: Would it be more readable if we used `String.format` here?

ganhongyao (commented):
Great work! I could not find many things to correct, mostly just minor nits. 👍

PR #179, opened 2021-08-22
[Wong Yun Rui Chris] iP

wyrchris:
### DUKE

> _"All quotes are made-up"_ - [Chris Wong](https://github.com/wyrchris) 

Duke is better as it's,

- Text-based
- Fast
- **Fun**😃

How to use it,

1. Download it.
2. Double-click it.

Features:

- [X] Adding/Deleting task
- [X] Marking task as done
- [ ] Searching for task by keyword

Example of the `code`:
```
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+        String[] inputArray = input.split(" ",2);
+        String reply = "";
+
+        switch (inputArray[0]) {
+        case "list":
+            reply = list.displayTask();
+            break;
+
+        case "done":
+            // Fallthrough
```
Jacobnbh, line 39: Hello! I like how you made use of `// Fallthrough` here to handle the marking and deleting of tasks. Reduces the number of times the code has to be written (making it succinct) for both types of commands since they require only an index 👍

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+
+        case "find":
+            if (inputArray.length == 1) {
+                throw new DukeException("The keyword is missing.");
+            }
+            reply = list.findTask(inputArray[1]);
+            break;
+
+        case "todo":
+            // Fallthrough
```
Jacobnbh, line 47: Great use of `// Fallthrough`, same comments as above! 👍

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Saves the data of the task in the list into a text file formatted.
+     *
+     * @param list The TaskList containing the tasks that is to be saved into
+     */
+    public void saveData(TaskList list) {
+        try {
+            // get file directory string
+            String[] dir = this.filepath.split("/");
```
Jacobnbh: I thought `dir` was a little too general for a variable name. Since it's an array, perhaps a plural form could be used? Or naming it differently altogether? 🤔 (e.g. `filePathArray`, even `dirArray` could be better)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+/**
+ * A class that encapsulates the saving/reading data for Duke.
+ */
+public class Storage {
+    private String filepath;
```
Jacobnbh: Not sure if "filepath" is one or two words😅 But if it is made up of two words, then as a variable, it should be in CamelCase: `filePath`

Jacobnbh (commented):
LGTM! Just a few nits to fix. Other than that, code is succinct and easy to understand! 😄

src/main/java/duke/task/Todo.java
```diff
@@ -0,0 +1,42 @@
+package duke.task;
+
+/**
+ * A wrapper for a Todos which is a task without date/time
+ *
+ * @author Wong Yun Rui Chris
+ */
+public class Todo extends Task{
```
jianoway: Possibly redundant suggestion but in accordance with the prof's recommendation for this Task type to be "ToDo", maybe change to the camelcase version?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A public constructor to initialise a duke.task.Deadline duke.task.Task.
+     *
+     * @param description The String description/name of the task
+     * @param by The String describing when the duke.task.Deadline need to be done by
+     */
+    public Deadline(String description, String by, Boolean isDone) {
+        super(description, isDone);
+        this.by = by;
```
jianoway, line 25: Maybe make this 'by' variable a little more specific? The byLd is the LocalDate version of it so it would be slightly clearer to be more specific with by.

jianoway (commented):
Overall, one of the cleaner iP repos I've seen. Your implementation of OOP is very good and I'm quite impressed actually!! :D Keep up the good work :)

src/main/java/duke/gui/DialogBox.java
```diff
@@ -0,0 +1,79 @@
+package duke.gui;
```
meerian, line 1: Nice use of subpackages!

src/test/java/duke/DukeTest.java
```diff
@@ -0,0 +1,52 @@
...
+            Scanner sc = new Scanner(new File("./data/duke.txt"));
+            Scanner sco = new Scanner(new File("./data/output.txt"));
+            while (sc.hasNext()) {
+                assertEquals(sc.nextLine(), sco.nextLine());
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println(e.getMessage());
+        }
+    }
+}
```
meerian, line 52: I like how rigorous your JUnit testings are!

meerian (commented):
Good job! Keep up the good work!

PR #182, opened 2021-08-22
[Pua Li Xue] iP

pualixue:
# EightBit
> “Life is too complicated not to be orderly.” – Martha Stewart ([source](https://dansilvestre.com/productivity-quotes/#10_Life_is_too_complicated_not_to_be_orderly_-_Martha_Stewart))

EightBit frees your mind of having to remember things you need to do. It's
* text-based
* easy to learn
* ~~fast~~ *SUPER* FAST to use

All you need to do is:
1. download it from [here](https://github.com/pualixue/ip/releases/tag/v0.1)
2. double-click it
3. add your tasks
4. let it manage your tasks for you 😄 

And it is **FREE**!

Features:
- [x] List your tasks
- [x] Add/remove tasks
- [x] Mark tasks as done
- [x] Search for tasks
- [ ] Reminders (coming soon)

#
```java
    EightBit eightBit = new EightBit(FILEPATH);
    eightBit.run();
}
```


src/main/java/eightbit/EightBitException.java
```diff
@@ -0,0 +1,25 @@
...
+public class EightBitException extends Exception {
+    private String errorMessage;
+
+    /**
+     * Constructs the exception with the error message.
+     *
+     * @param errorMessage Error message to be shown to user.
+     */
+    public EightBitException(String errorMessage) {
+        this.errorMessage = errorMessage;
```
pbthang: I think you should use the constructor of the parent class.
```suggestion
        super(errorMessage);
```

src/main/java/eightbit/task/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+        return this.by;
+    }
+
+    /**
+     * @return String representation of a Deadline.
+     */
+    @Override
+    public String toString() {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd MMM yyyy HH:mm");
+        return "[D]" + super.toString() + " (by: " + by.format(formatter) + ")";
```
pbthang, line 51: In this case, should we use `String.format()` instead, since for me, it's more readable and less verbose?

src/main/java/eightbit/task/Task.java
```diff
@@ -0,0 +1,55 @@
+package eightbit.task;
+
+/**
+ * Represents a task that can be completed.
+ */
+public class Task {
```
pbthang: Should this class be abstract, because it's only instantiated in the forms of Todo, Event, etc?
```suggestion
public abstract class Task {
```

pbthang (commented):
Looks good. Some minor changes requested.

src/main/java/eightbit/task/Task.java
pualixue: Thanks for your suggestions!

src/main/java/eightbit/command/FindCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param storage  Storage responsible for reading/writing the file.
+     * @return The response after executing the command.
+     */
+    @Override
+    public String execute(TaskList taskList, Storage storage) {
+        // Get tasks containing the keyword
+        ArrayList<Task> results = new ArrayList<>();
+        for (int i = 0; i < taskList.size(); i++) {
+            Task task = taskList.get(i);
+            if (task.getDescription().toLowerCase(Locale.ROOT).contains(keyword)) { // case-insensitive search
```
chongjunwei, line 41: The comment should be indented relative to its position in the code, in this case in line with the if (task.getDescription()...) statement.

src/main/java/eightbit/command/EventCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Adds the event into the user's list.
+     *
+     * @param taskList User's list of tasks.
+     * @param storage  Storage responsible for reading/writing the file.
+     * @return The response after executing the command.
+     */
+    @Override
+    public String execute(TaskList taskList, Storage storage) {
```
chongjunwei, line 31: I think it would be better to name the method after the functionality, instead of execute() which may not be the most intuitive. Perhaps something like addEventToList()?

chongjunwei (commented):
Looks good, just a few minor comments based on the code quality and standard guidelines given. Good job!

PR #195, opened 2021-08-22
[Wilbur Fong] iP

wilburrito:
## Hello! Duke is here to help you organise your day!

> “Whether you think you can, or you think you can't--you're right.” - Henry Ford

Duke will help free some RAM in your brain by taking note of and tracking tasks that you tell Duke about. It's:

- Easy to use
- **~Fun~** **Exciting** to use
- Text-based

All you have to do is:

1. Download it from [here](https://github.com/wilburrito/ip/releases/tag/v0.1).
2. Open Duke up.
3. Add your tasks according to the specified formats and types of tasks that Duke can track
4. Let it manage the tasks for you!

Best part is, Duke is **FREE**!

Features:

- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders

If you know a little Java, you can use Duke to practice it too. Here's the `main` method:

`public static void main(String[] args) {
        Duke.awaken();
    }`


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,126 @@
...
+public class Storage {
+
+    private static boolean appendMode = true;
+    private static String filePath = "./data/dukeData.txt";
+
+    /**
+     * loads data about the task list from dukeData.txt.
+     * @return an ArrayList of TaskItems that will be used to create Duke's TaskList.
+     */
+    public static ArrayList<TaskItem> loadData() {
```
kaldius, line 16: Your `loadData()` method is rather long and deeply nested. Abstracting out certain parts to helper methods/classes could make it simpler!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,126 @@
...
+                return new ArrayList<TaskItem>();
+            } else {
+                BufferedReader bufReader = new BufferedReader(new FileReader("./data/dukeData.txt"));
+                ArrayList<TaskItem> taskItems = new ArrayList<>();
+                String task = bufReader.readLine();
+                while (task != null) {
+                    String completed = task.substring(4, 5);
+                    String taskType = task.substring(1, 2);
+
+                    if (taskType.equals("D")) {
```
kaldius, line 40: This section where you check for each type of task and add them to the list could be abstracted away and placed in the `TaskItem` class or your `TaskList` class to reduce bulk in this method.

src/main/java/duke/TaskItem.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public void completeTask() {
+        this.completed = true;
+    }
+
+    /**
+     * toFileString returns the TaskItem in a string.
+     * @return returns the Deadline in a string format that is specifically made for the dukeData.txt file.
+     */
+    public String toFileString() {
```
kaldius, line 44: This method seems redundant since you already have a `toString()` method that contains the same lines of code.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+        this.arrayList.remove(index);
+
+    }
+
+    /**
+     * finds tasks that contain a given keyword.
+     * @param keyword if a task contains this, it will be printed to the user.
+     * @return the TaskList containing tasks matching the keyword, if any.
+     */
+    public TaskList findTask(String keyword) {
```
kaldius, line 62: Is there any particular reason why you implemented this method this way? It seems to me a simple `for` loop iterating through the ArrayList checking if each task's string representation `.contains()` a certain string would be simpler.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Prints a message to the user about the task that the user just deleted.
+     * @param task The task that got deleted.
+     * @param taskListSize The updated TaskList size.
+     */
+    public static String deletedTaskMessage(TaskItem task, int taskListSize) {
+        String body = "";
+        if (taskListSize == 1) {
+            /* System.out.println("Now you have " + 1 + " task in the list."); */
+            body = "Now you have " + 1 + " task in the list.\n";
```
kaldius, line 48: It might be simpler to just type the entire thing as one string instead of concatenating multiple.

kaldius (commented):
Hi there! I had a good time reading your code. You have clearly put in a lot of effort and thought into your design. I made a few comments in areas where I have a few suggestions on how you could make your code even better. Overall I think this is a well implemented program! Good work!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+    }
+
+    /**
+     * deletes a task from the list that duke.Duke creates.
+     * @param taskToBeDeleted the index of the task to be deleted.
+     * @throws DukeException An exception stemming from incorrect or awkward input to duke.Duke.
+     */
+    public static String deleteTask(String taskToBeDeleted) throws DukeException {
```
aishh12, line 71: Perhaps you may want to consider defining the deleteTask method in the TaskList class (in order to maintain encapsulation and shortening the code for this class)?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+        return output;
+    }
+
+    /**
+     * Finds tasks based on the keyword the user inputs and prints them to the user
+     * @param keyword keyword used to find matching tasks
+     * @return a TaskList with matching tasks to be printed.
+     */
+    public static TaskList findTask(String keyword) {
```
aishh12, line 153: As mentioned earlier, you may want to consider adding all the methods related to the task list (adding, deleting, marking and finding tasks as well as printing the list) under Tasklist.java instead?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,55 @@
...
+            int hour = byDateAndTime.getHour();
+            if (byDateAndTime.getHour() > 12) {
+                timeOfDay = "PM";
+                hour -= 12;
+            } else if (byDateAndTime.getHour() == 12) {
+                timeOfDay = "PM";
+            } else {
+                timeOfDay = "AM";
+            }
+            this.dateAndTimeInString = this.byDateAndTime.getDayOfWeek().toString() + " " + hour + timeOfDay;
```
aishh12, line 32: Would you consider formatting the date and time to be stored in the "dateAndTimeInString" by using the methods under the DateTimeFormatter class? This way you can make your code cleaner and print in a specified format too :)

src/main/java/duke/Storage.java
aishh12: Yup, I agree with this :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,350 @@
...
+                        //System.out.println(number + "." + matchingTasks.get(i).toString());
+                        response += number + "." + matchingTasks.get(i).toString() + "\n";
+                        number++;
+                    }
+                }
+            } catch (DukeException e) {
+                Ui.customErrorMessage(e.getMessage());
+            }
+        } else {
+            response = "____________________________________________________________\n"
```
aishh12, line 363: Since there are multiple instances to draw the lines for formatting, you could consider having a method to print the line to make your code cleaner!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Prints a message to the user about the task that the user just deleted.
+     * @param task The task that got deleted.
+     * @param taskListSize The updated TaskList size.
+     */
+    public static String deletedTaskMessage(TaskItem task, int taskListSize) {
+        String body = "";
+        if (taskListSize == 1) {
+            /* System.out.println("Now you have " + 1 + " task in the list."); */
```
aishh12, line 47: Commented code can be removed since it is no longer needed.

aishh12 (commented):
Overall, good job! Some methods were pretty long and abstracting away parts of it could significantly reduce code length and enhance readability. Besides that it's good to go, all the best :)

PR #197, opened 2021-08-22
[Yeo Yiheng] iP

yeo-yiheng:
Ever wanted a personal assistant to keep track of tasks but people are too costly and notebooks are too cumbersome?

DukePro is your **free** and **lightweight** solution which is :
* text-based
* ~~quick~~ _SUPER QUICK_ to access
* simple yet **powerful**

You can own your very own DukePro by:
1. Downloading it from [here.](https://github.com/yeo-yiheng/ip)
2. Run in a terminal.
3. Add your tasks
4. Let it remember and manage your tasks for you 🤩 

Have we mentioned it yet? Or if we did, we shall do so again. It is absolutely **FREE**!

Core Features :
- [X] Managing tasks
- [X] Finding tasks by keyword
- [ ] Reminders (coming soon)

If you're a Java programmer and you'd like to inspect the code, you can inspect the `main` method:
```java
public class Duke {
    }
}
```


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,300 @@
...
+            try {
+                int index = Integer.parseInt(stringIndex) - 1;
+                if (task.markDone(index)) {
+                    USER_INTERFACE.taskComplete(TaskList.getTask(index));
+                }
+                break;
+            } catch (NumberFormatException e) {
+                USER_INTERFACE.integerInputWarning();
+                break;
+            }
```
Py0000, line 56: Perhaps you can consider extracting the `break` instead of repeating it in both `try` and `catch` block?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,300 @@
...
+                if (deletedTask == null) {
+                    break;
+                }
+                task.delete(index2);
+                USER_INTERFACE.taskDeleted(deletedTask);
+                break;
+            } catch (NumberFormatException e) {
+                USER_INTERFACE.integerInputWarning();
+                break;
+            }
```
Py0000, line 74: Perhaps you can consider extracting the `break` instead of repeating it in both `try` and `catch` block?

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,77 @@
+package duke.parser;
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Todo;
+import org.junit.jupiter.api.Test;
```
Py0000, line 6: Perhaps 'import org.junit.jupiter.api.Test;` should be separated from the 'duke' import statements? Please do check if this is correct as I am not too sure on this as well.

src/test/java/duke/storage/StorageTest.java
```diff
@@ -0,0 +1,49 @@
+package duke.storage;
+
+import duke.task.TaskList;
+import duke.task.Todo;
+import org.junit.jupiter.api.Test;
```
Py0000, line 5: Similar to the previous suggestion, perhaps import `org.junit.jupiter.api.Test` should be separated from the `duke` import statements?

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,244 @@
...
+        if (index < 0 || index >= TASKS.size()) {
+            USER_INTERFACE.invalidIntegerWarning();
+            return null;
+        } else {
+            return TASKS.get(index);
+        }
+    }
+
+    /**
+     * This method finds the tasks that matches the input keyword and subsequently
```
Py0000, line 195: Not sure if this is necessary, but perhaps you can consider starting the Javadoc comments with `Finds the tasks...` ?
I have noticed it in some other files as well.
I have also noticed that there is also 2 different style used for the Javadoc comments; one that starts with "This method ..." and other without "This method". Perhaps you can considering to standardize it?

Py0000 (approved):
I like your program! It was well-coded and easy to read and follow through. 👍 
Minor nitpick: Perhaps you can consider adding some line break in between some statements in some of your codes to further improve readability? 

Other than a few minor nitpicks, it LGTM. 😄 

So sorry that this comment is separated from the rest below 😨

src/main/java/duke/parser/Parser.java
Py0000: Perhaps you can consider to extract `break` out instead of writing it in both `try` and `catch` block?

src/main/java/duke/parser/Parser.java
Py0000: Similar to the previous suggestion, perhaps you can consider to extract `break` out instead of writing it in both `try` and catch block?

src/test/java/duke/parser/ParserTest.java
Py0000, line 6: Perhaps there should be a line break before `import org.junit.jupiter.api.Test;` ? Please do check on this as I am not sure about this as well.

src/test/java/duke/storage/StorageTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.jupiter.api.io.TempDir;
```
Py0000, line 6: Similar to the previous suggestion, perhaps there should be a line break before `import org.junit.jupiter.api.Test;` ? Please do check on this as I am not sure about this as well.

src/main/java/duke/task/TaskList.java
Py0000: Not sure if this is required/necessary. Perhaps you can consider to start the Javadoc comments with `Finds the tasks ...` instead of `This method...` ? I have also noticed it in some other Javadoc comments as well.
Also, I have noticed that 2 styles are used for Javadoc comments, namely starting with "This method" and not starting with "This method". Perhaps you can consider to standardize it ?

Py0000 (approved):

src/main/java/duke/parser/Parser.java
yeo-yiheng: Thank you for this suggestion. I will look into extracting the `break` statement.

src/main/java/duke/parser/Parser.java
yeo-yiheng: Thanks! Will do 👍

src/test/java/duke/parser/ParserTest.java
yeo-yiheng: You're right, we should separate related imports. I will change this, thanks for noticing 😄

src/test/java/duke/storage/StorageTest.java
yeo-yiheng: Noted on this! Thank you.

src/main/java/duke/task/TaskList.java
yeo-yiheng: I agree with you. It should be phrased in a different manner. Thanks for pointing that out and I will make the necessary changes.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,36 @@
...
+
+import duke.parser.Parser;
+import duke.task.TaskList;
+import duke.ui.UserInterface;
+
+import java.util.Locale;
+import java.util.Scanner;
+
+public class Duke {
+    private static final TaskList TASK = new TaskList();
```
Yttruire, line 15: I think that maybe this should also just be named `TASKLIST` or at least `TASKS` (since it is a container for many objects). `TASK` by itself might sound a bit misleading I think.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public static void main(String[] args) {
+
+        USER_INTERFACE.greet();
+        TASK.loadArrayList(); // Loads the array list based on our file on hard disk
+        boolean breakWhile;
+
+        while (SCANNER.hasNext()) {
+            String firstWord = SCANNER.next().toLowerCase(Locale.ROOT);
+            breakWhile = PARSER.firstCommandParser(firstWord, SCANNER);
```
Yttruire: According to coding standards, method names should be verbs, so perhaps this method should be named as `firstCommandParse`?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,300 @@
...
+        "APRIL", "MAY", "JUNE", "JULY", "AUGUST", "SEPTEMBER", "OCTOBER", "NOVEMBER",
+        "DECEMBER" };
+    private static final int OFFSET = 7;
+
+    /**
+     * A default constructor for the Parser class.
+     */
+    public Parser() {}
+
+    private final static UserInterface USER_INTERFACE = new UserInterface();
```
Yttruire: It seems that you have to create more than one UserInterface in your program, which sounds a little weird. Perhaps a suggestion would be for the Parser to have a reference to it rather than create its own one?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,300 @@
...
+                double timingEnd = Double.parseDouble(parsedTime[6]) * 100
+                        + (parsedTime[7].equals("PM") ? 1200 : 0);
+                int flattenedTimingStart = (int) timingStart;
+                int flattenedTimingEnd = (int) timingEnd;
+                startTime = Integer.toString(flattenedTimingStart);
+                endTime = Integer.toString(flattenedTimingEnd);
+            }
+            ld = LocalDate.parse(timeFormat);
+        } catch (ArrayIndexOutOfBoundsException e) {
+            USER_INTERFACE.nullFunction();
```
Yttruire: This method might be better as a verb too. At the glance I'm confused what a `nullFunction` is supposed to do

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * This class represents the storage of the task list and handles all the
+ * interaction between the user and the file contents.
+ */
+public class Storage {
+
+    private final File file;
+    private final String filePath;
+    private final static UserInterface USER_INTERFACE = new UserInterface();
```
Yttruire: Same as per Parser, a bit strange for every component to have their own user interface

Yttruire (commented):
Very well done! Can see that you have put a lot of effort into this, with very detailed unit tests, very robust features such as regex as well. Apart from the small points that were brought up, looks great! :D

PR #201, opened 2021-08-22
[Gnoh Cheng Yi] iP

GnohChengYi:
# BluePro

> “Don't comment bad code ― rewrite it” 
> — Brian W. Kernighan and P. J. Plaugher ([source](https://en.wikipedia.org/wiki/The_Elements_of_Programming_Style))

BluePro aims to help you simplify stuff. It is

* text-based
* easy to learn
* _~~SUPER~~_ FAST to use

What are you waiting for? Do the following now!

1. Download BluePro from [here](https://github.com/GnohChengYi/ip/releases)
2. Open a command window in the folder where the jar file is
3. Run the command `java -jar ip.jar`

And it is **_FREE_**!

Features:

* [x] Manage tasks
* [x] 3 types of tasks: ToDo, Deadline, Event
* [ ] Reminders (coming soon)

If you know Java, feel free to fork and create your unique BluePro! Here's the `main` method:

```java
public class Blue {
        new Blue("data/tasks.txt").run();
    }
}
```

Happy coding! 😁


src/main/java/blue/Blue.java
```diff
@@ -0,0 +1,138 @@
...
+import blue.handler.ToDoHandler;
+
+/**
+ * Entry point of the Blue application.
+ * Initializes the application and interacts with the user.
+ */
+public class Blue {
+    private static final String CONFUSED_RESPONSE =
+            "☹ OOPS!!! I'm sorry, but I don't know what that means :-(";
+    private final Storage storage;
```
xyliew25, line 21: Constants should be in caps.

src/main/java/blue/Blue.java
```diff
@@ -0,0 +1,138 @@
...
+    private TaskList tasks;
+    private final Ui ui;
```
xyliew25: Same here, constants should be in caps. Generally, variable names with the `final` keyword should be in caps. I saw there is a few more such violation in the other files.

src/main/java/blue/Blue.java
```diff
@@ -0,0 +1,138 @@
...
+        ListHandler listHandler = new ListHandler(tasks);
+        ToDoHandler toDoHandler = new ToDoHandler(tasks);
+        DeadlineHandler deadlineHandler = new DeadlineHandler(tasks);
+        EventHandler eventHandler = new EventHandler(tasks);
+        DoneHandler doneHandler = new DoneHandler(tasks);
+        DeleteHandler deleteHandler = new DeleteHandler(tasks);
+        FindHandler findHandler = new FindHandler(tasks);
+        ExitHandler exitHandler = new ExitHandler(tasks);
+
+        // put the handlers into HashMap
```
xyliew25: Not sure if this is a coding standard violation, but I think the first letter of the first word of a comment can be in caps to ensure consistency. This part is probably overlooked.

src/main/java/blue/Parser.java
```diff
@@ -0,0 +1,32 @@
+package blue;
+
+import java.util.Arrays;
+
+/**
+ * Helps in making sense of the user input.
+ */
+public class Parser {
+    static String getCommand(String input) {
```
xyliew25: Not sure if this is a coding standard violation as well and not sure if you intended to make this method package-private, else I think you can add in the appropriate access modifier. If it is the former, I saw there are a few such cases in the other files as well.

src/main/java/blue/Blue.java
```diff
@@ -0,0 +1,138 @@
...
+        commandHandlers.put(Command.TODO, toDoHandler);
+        commandHandlers.put(Command.DEADLINE, deadlineHandler);
+        commandHandlers.put(Command.EVENT, eventHandler);
+        commandHandlers.put(Command.DONE, doneHandler);
+        commandHandlers.put(Command.DELETE, deleteHandler);
+        commandHandlers.put(Command.FIND, findHandler);
+        commandHandlers.put(Command.EXIT, exitHandler);
+    }
+
+    private boolean canHandle(String input) {
```
xyliew25: This method name sounds, to me, like a boolean rather than a verb. Perhaps a suggestion eg. `checkCanHandle`

xyliew25 (commented):
Overall, it was a job well done! Just a few minor issues. Do correct me if I misunderstood or commented something wrongly :)

src/main/java/blue/Blue.java
xyliew25: Regarding this, I reviewed and I think it's alright to name a method returning boolean to sound like a boolean, as per the sample code given under MoreOOP where there is a `isExit()` method in the `Command` class.

src/main/java/blue/Command.java
```diff
@@ -0,0 +1,15 @@
...
+public class Command {
+    public static final String LIST = "list";
+    public static final String TODO = "todo";
+    public static final String DEADLINE = "deadline";
+    public static final String EVENT = "event";
+    public static final String DONE = "done";
+    public static final String DELETE = "delete";
+    public static final String EXIT = "bye";
+    public static final String FIND = "find";
+}
```
xyliew25: Perhaps can consider using enums?

src/main/java/blue/task/Task.java
```diff
@@ -0,0 +1,49 @@
...
+    public String toString() {
+        String string = "[";
+        if (isDone) {
+            string += "X";
+        } else {
+            string += " ";
+        }
+        string += "] " + title;
+        return string;
+    }
```
xyliew25:
```suggestion
    public String toString() {
        String taskStatus = isDone ? "X" : " ";
        return "[" + taskStatus + "] " + title;
    }
```

src/main/java/blue/task/ToDo.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * Returns a String that represents this class.
+     *
+     * @return String that represents this class.
+     */
+    public static String getClassRepr() {
+        return "T";
+    }
```
xyliew25: Perhaps can replace with a public static constant property eg. `public static final CLASS_REPR = "T";`, as there is no function logic present? Similar for Event and Deadline class.

src/main/java/blue/task/Task.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Constructs a Task.
+     *
+     * @param title  Title of task.
+     * @param isDone Whether the task is done.
+     */
+    public Task(String title, boolean isDone) {
+        this.title = title;
+        this.isDone = isDone;
+    }
```
xyliew25, line 23: Hm may I know what is this overloaded constructor for? As I noticed this overloaded constructor is not called anywhere, except the other constructor. Also, personally, I don't think users will add a task that is already done.

src/main/java/blue/handler/FindHandler.java
```diff
@@ -0,0 +1,36 @@
...
+    public FindHandler(TaskList taskList) {
+        super(taskList);
+    }
+
+    @Override
+    public String handle(String input) throws BlueException {
+        List<Task> foundTasks = new ArrayList<>();
+        String keyword = Parser.getArguments(input)[0];
+        for (Task task : taskList.getAll()) {
+            if (task.getTitle().contains(keyword)) {
```
xyliew25: This is a functionality issue rather than a code quality issue. It seems that the keyword is only searched on the task title. Perhaps you can just search the whole string so as to allow users to search for dates as well ie. `at` from `Event` and `by` from `Deadline`.

src/main/java/blue/handler/ExitHandler.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public ExitHandler(TaskList taskList) {
+        super(taskList);
+    }
+
+    @Override
+    public String handle(String input) {
+        return EXIT_CONTENT;
+    }
+}
```
xyliew25, line 16: If I'm not wrong, I think this class will never be used as if the `EXIT` command is being detected in the `canHandle()` method in `Blue` class, `ui` will `sayGoodbye()` and `shouldContinue` is `false` thus breaking out of the while loop.

xyliew25 (commented):
Very sorry! I thought I was supposed to review the coding standards on your pull request. I've added the comments for coding quality as well.

src/main/java/blue/Blue.java
GnohChengYi: I think constants refer to only `static final`. This is not specifically mentioned in the coding style guide of SE-EDU, but it asks us to follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) for  anything not mentioned. 

In the guide, it states that:
> Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects. This includes primitives, Strings, immutable types, and immutable collections of immutable types. If any of the instance's observable state can change, it is not a constant. Merely intending to never mutate the object is not enough.

So, I think the naming of `storage` is fine here.

src/main/java/blue/Blue.java
GnohChengYi: I can't find this mentioned anywhere in SE-EDU coding standard and Google Java Style Guide, so capitalizing the first letter in a comment is probably not a coding standard. (If you can find it, let me know).

But I think you have a valid point, that capitalizing the first letter makes it more consistent.

src/main/java/blue/Parser.java
GnohChengYi: Yup, I intend to make everything as private as possible, that's why there are many methods with only package-level access.

src/main/java/blue/Blue.java
GnohChengYi: I am actually not quite sure about how to name this method. `canHandle` sounds like only checking whether it can handle, but it is actually also handling it, which makes the name misleading.

The `boolean` was needed to determine whether it is time to exit the application. I am not sure whether it is better to handle the exit command at the GUi level or at the level of the bot logic. I might improve the design if I found a better way.

PR #204, opened 2021-08-22
[Michael Lee] iP

michael-lee-sk:
**Duke**
> Be regular and orderly in your life, that you may be violent and original in your work. - Clive Barker [Source](https://www.brainyquote.com/quotes/gustave_flaubert_109857)

Duke creates your lists for you, so you do not have to
* text-based
* *SUPER* easy to learn
* ~FAST~ SUPER FAST

All you need to do is,
1. Download it here for a low price of $0
2. let it manage your tasks for you 😉

Features include
- [x] Managing tasks


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+     * Parses the user command and inputs the task accordingly.
+     *
+     * @throws DukeException If task input by user is incomplete.
+     * @throws DeleteException If delete is incomplete.
+     * @throws IOException If an input or output operation is failed or interpreted.
+     */
+    public void parseCommand() throws DukeException, DeleteException, IOException, FindException {
+
+        Activity activity;
+        if (command.equals("bye")) {
```
Ongzl: This method / while loop might be too long. This [coding quality website](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods)  stated the following:

>Be wary when a method is longer than the computer screen, and take corrective action when it goes
>beyond 30 LOC (lines of code). The bigger the haystack, the harder it is to find a needle.

This is a little longer than 30 LOC , though I too am not sure how to condense this part of logic haha, maybe implementing a 'command' class?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,126 @@
...
+import java.io.IOException;
+import java.util.Scanner;
+
+/**
+ * Deals with loading tasks from the file and saving tasks in the file.
+ */
+public class Storage {
+    private static String filePath;
+
+    private boolean exit = false;
```
Ongzl: I think the boolean variable should be name to something that sounds like booleans? Maybe shouldExit or something?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Public constructor for an event.
+     *
+     * @param description Description of the event.
+     * @param dateAndTime Date and time of the event.
+     */
+    public Event(String description, String dateAndTime) {
+        super(description);
+        this.description = description;
```
Ongzl: I am not sure if this double description assignment is necessary since there is a description variable for Task already and the super(description) already assigns the description

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Event class is a task. The input must be in such a format
+ * "Event <event name> /at <event date>".
+ */
+public class Event extends Task{
+    protected String description;
```
Ongzl: I am not sure if having two String description for the children of Task is a good idea since it looks like you are assigning description twice and have 2 description variables. I think you did the same for Event, Deadline and ToDo, I am not sure if this is wise.

Ongzl (commented):
Coding standard looks good to me. I am not sure if the double description thing is under coding standards or not, but just thought to let you know, cheers

src/main/java/duke/Parser.java
Xiaoyunnn: I agree with @Ongzl. Perhaps, you can consider moving the content in the switch cases to separate methods to shorten this method body?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Public constructor for a task.
+     *
+     * @param description Description of the event.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
```
Xiaoyunnn: I think this JavaDoc should begin with a verb? e.g. "Instantiates" or "Constructs". I believe this applies for your other constructors as well.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,79 @@
...
+     * Checks if a task is done, or not.
+     *
+     * @return Particular string if task is done, or not done.
+     */
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]");
+    }
+
+    /**
+     * Set boolean if task is done, or not.
```
Xiaoyunnn: I think this line should start with "Sets"?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Parses the user command and inputs the task accordingly.
+     *
+     * @throws DukeException If task input by user is incomplete.
+     * @throws DeleteException If delete is incomplete.
+     * @throws IOException If an input or output operation is failed or interpreted.
+     */
+    public void parseCommand() throws DukeException, DeleteException, IOException, FindException {
```
Xiaoyunnn: I think you missed out the description for `FindException`?

src/main/java/duke/Event.java
Xiaoyunnn: I agree with @Ongzl , you can still get the subclass instance's `description` (from parent class) without the subclass having a private field for `description`.

src/main/java/duke/Storage.java
michael-lee-sk: Thanks for pointing this out, will do the changes!

src/main/java/duke/Event.java
michael-lee-sk: Thanks for pointing this out, will do the changes!

src/main/java/duke/Task.java
michael-lee-sk: Thanks for pointing this out, did not know this haha. Will do the changes accordingly :)

src/main/java/duke/Parser.java
michael-lee-sk: hmm, may i clarify something? 30 LOC is for one method right? Technically I have many different switch cases, by which each case has << 30 LOC (the longest switch case has 15 LOC inclusive of empty lines). So a method refers to one switch case right, not the entire switch?

Ongzl: sftlr, I am not too sure too because the site itself is quite vague, but I think as long as measures are put in to minimize the size of the method then it should be fine.

PR #206, opened 2021-08-22
[Felissa Faustine] iP

felissaf:
# MyJournal 
MyJournal __*stores your busy schedule*__ and __*keeps all of your important deadlines and upcoming events*__!:laughing:

It is:
* Text-based
* Easy to use

What MyJournal can do:
- [x] Show your tasks (todo/event/deadline0
- [X] Mark completed tasks
- [X] Find tasks
- [X] Delete tasks

All you need to do is,
1. Download [here](https://github.com/felissaf/ip/releases)!
1. Add your tasks
1. You're all set!

As Benjamin Franklin said:
> For every minute spent in organising,
> an hour is earned.

This is the `main` method:
```java
    try {
        new MyJournal("./tasks.txt").run();
    } catch (IOException e) {
        e.printStackTrace();
    }
}
```


src/main/java/myjournal/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                    tasks.add(event);
+                } else {
+                    Deadline deadline = new Deadline(description, time);
+                    deadline.setState(isDone);
+                    tasks.add(deadline);
+                }
+            }
+        }
+        return tasks;
+    }
```
Xiaoyunnn: This `load()` method might be too long. This [coding quality website](https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods) stated the following:

> Be wary when a method is longer than the computer screen, and take corrective action when it goes beyond 30 LOC (lines of code). The bigger the haystack, the harder it is to find a needle.

It also suggests that you should not have more than 3 levels of indentations. Maybe you can move the content in the `else` block into a separate method to make the code look cleaner?

src/main/java/myjournal/Ui.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Creates the Ui object.
+ *
+ * @author felissafaustine
+ */
+public class Ui {
+    /**
+     * Prints the welcome message.
+     * @return
```
Xiaoyunnn: I think you forgot to add the description for `@return` here XD

src/main/java/myjournal/Ui.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Prints out the statement after a task is added.
+     *
+     * @param taskList The list of all tasks.
+     */
+    public String taskAddPrint(TaskList taskList) {
+        return "Okay!! I've added the following task:\n"
+                + taskList.getTask(taskList.getSize() - 1) + "\n"
+                + "Now you have " + taskList.getSize() + " in the list";
+    }
```
Xiaoyunnn: I realised you seem to forget to add `@return` for most of your JavaDoc for methods that return something in this Ui class?

src/main/java/myjournal/Parser.java
```diff
@@ -0,0 +1,292 @@
...
+            throw new InvalidTaskNumberException("OOPS!!! Please specify the task "
+                    + "that needs to be marked as done!");
+        }
+        int index = line.nextInt() - 1;
+        if (index >= tasks.getSize() || index < 0 || tasks.getTask(index) == null) {
+            throw new InvalidTaskNumberException("OOPS!!! Please enter a valid task number!");
+        }
+        tasks.getTask(index).setState(true);
+        return tasks.getTask(index);
+    }
```
Xiaoyunnn: Since the Exception thrown is not caught in the method body, shouldn't the method signature include `throws InvalidTaskNumberException` or `throws EmptyDescriptionException, InvalidWordException`

src/main/java/myjournal/Parser.java
```diff
@@ -0,0 +1,292 @@
...
+                String beforeOrAfterNoon = time.getHour() >= 12 ? "pm" : "am";
+                int hour = time.getHour() == 0
+                        ? 12
+                        : time.getHour() >= 12
+                        ? time.getHour() - 12
+                        : time.getHour();
+                int min = time.getMinute();
+                String timeFormatted = (String.valueOf(hour).length() == 1 ? "0" + hour : hour)
+                        + ":" + (min < 10 ? "0" + min : min) + beforeOrAfterNoon;
+                parsed = parsed + " " + timeFormatted;
```
Xiaoyunnn: Perhaps you can consider using [DateTimeFormatter](https://www.baeldung.com/java-datetimeformatter) here to simplify your code?  😊

src/main/java/myjournal/Storage.java
felissaf: Thank you for the suggestion!! I do feel like it’s too long but wasn’t really sure how to shorten it. I’ll try taking out the else block!:)

src/main/java/myjournal/Ui.java
felissaf: Thanks for pointing it out!! I forgot to change the javadocs after changing my code oops XD

src/main/java/myjournal/Parser.java
felissaf: Okay I’ll try!! Thanks for the suggestion!:)

src/main/java/myjournal/Parser.java
MGRL2201: Perhaps this could apply to parseDelete too?

src/main/java/myjournal/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+        ArrayList<Task> tasks = new ArrayList<>();
+        while (sc.hasNextLine()) {
+            currLine = sc.nextLine();
+            String taskDescription = currLine.substring(5);
+            boolean isDone = currLine.charAt(2) == '1';
+            if (currLine.charAt(0) == 'T') {
+                Todo todo = new Todo(taskDescription);
+                todo.setState(isDone);
+                tasks.add(todo);
+            } else {
```
MGRL2201, line 74: Perhaps to shorten your code, you could try using `.split(" \\| ")` which will split the line at the character |?

MGRL2201 (commented):
Your code is well written and clean!

src/main/java/myjournal/Storage.java
felissaf: Thanks for the suggestions!!

Agentum07 (commented):
Overall, the code looks great, good job!

I just have one suggestion - maybe create a "dukeException" class and have your other exception classes extend from that.

PR #214, opened 2021-08-22
[Gary Yansen] iP

g4ryy:
# Nyx Chatbot :robot:

>“Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes/))

Nyx isn't an ordinary chatbot. It helps you to manage your tasks and events, and most importantly it's
*  Easy to use  
* Responsive 
* ~~FAST~~ **SUPER FAST** to use :zap:

With Nyx, you no longer have to worry about forgetting important tasks or events! 

Nyx supports the following functionalities:

- [x] Add up to 3 different types of tasks: _Todo_, _Deadline_, _Event_
- [x] Mark task as done
- [x] Delete task
- [x] List all tasks
- [x] Search for tasks using keyword 

### Getting Started

1. Download the latest version from [here](https://github.com/g4ryy/ip/releases/tag/v0.1)
2. Follow the instructions [here](https://github.com/g4ryy/ip)
3. Start chatting with Nyx :zany_face:

``` Java
    new Nyx("data", "nyx.txt").run();
}
```


src/main/java/Nyx.java
```diff
@@ -0,0 +1,65 @@
...
+            + "|   \\   | |__    __ __     __\n"
+            + "| |\\ \\  | |\\ \\  / / \\ \\   / /\n"
+            + "| | \\ \\ | | \\ \\/ /   \\ \\ / / \n"
+            + "| |  \\ \\| |  \\  /    /    /\n"
+            + "| |   \\   |  / /    / / \\ \\\n"
+            + "|_|    \\__| /_/    /_/   \\_\\\n";
+    private final Ui ui;
+    private TaskList taskList;
+    private final Storage storage;
+
```
ryantianj: Great use of final modifiers! Makes it really clear that these Objects will not be changed.

src/main/java/nyx/Task.java
```diff
@@ -0,0 +1,54 @@
...
+
+    String getContent() {
+        return this.content;
+    }
+
+    /**
+     * Returns a String representation of the task in the format required for saving into hard disk.
+     * @return String representation of the task in the format required to save into text hard disk.
+     */
+    public abstract String dataFormat();
```
ryantianj: Perhaps naming it with a verb, such as formatData,  would be better.

src/main/java/nyx/TaskList.java
```diff
@@ -0,0 +1,113 @@
...
+            return true;
+        }
+    }
+
+    /**
+     * Returns a combined String representation of all the tasks in the list
+     * in the format required for saving into hard disk.
+     * @return String representation of all the tasks in the list in the format required for saving into hard disk.
+     */
+    public String genSaveFormat() {
```
ryantianj: Personally, 'gen' was not very clear to me (generate?), may be better to write in full (generateSaveFormat)

src/main/java/nyx/TaskList.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Returns a String representation of the list.
+     * @return String representation of the list.
+     */
+    @Override
+    public String toString() {
+        if (tasks.isEmpty()) {
+            return "You do not have any task currently";
```
ryantianj: Minor grammar, perhaps it should be "You do not have any task**s** currently"

src/main/java/nyx/Task.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    protected int getStatusInt() {
+        return isDone ? 1 : 0;
+    }
+
+    String getContent() {
```
ryantianj: Perhaps adding an access modifier here would make this method easier to read!

ryantianj (commented):
Honestly, your code formatting and quality was excellent. It was very difficult to find faults within your code and I learnt quite a bit (like OOP design) reading your code. I had to really nitpick your code to find faults with it. Hope it will be useful for you!

src/main/java/nyx/Task.java
rohit0718: perhaps `formattedData` might be more appropriate here since it returns a string. what do you think @ryantianj?

src/main/java/nyx/TaskList.java
rohit0718: I agree. Perhaps can even use something simpler like `getSaveFormat`

src/main/java/nyx/Task.java
rohit0718: i think the author intended it to be package private here, but i might be mistaken. @g4ryy is this method being used elsewhere in your package?

src/main/java/Nyx.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final String logo = " __      _\n"
+            + "|   \\   | |__    __ __     __\n"
+            + "| |\\ \\  | |\\ \\  / / \\ \\   / /\n"
+            + "| | \\ \\ | | \\ \\/ /   \\ \\ / / \n"
+            + "| |  \\ \\| |  \\  /    /    /\n"
+            + "| |   \\   |  / /    / / \\ \\\n"
+            + "|_|    \\__| /_/    /_/   \\_\\\n";
+    private final Ui ui;
+    private TaskList taskList;
+    private final Storage storage;
```
rohit0718: super small thing, but perhaps can grp the variables with the similar prefix together! in this case that would be:
```suggestion
    private final Ui ui;
    private final Storage storage;
    private TaskList taskList;
```

src/main/java/Nyx.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Begins interactions with the user until a bye command is encountered.
+     */
+    public void run() {
+        ui.displayStart(logo);
+
+        String input;
+
```
rohit0718: perhaps can remove the spacing here since its not needed.

src/main/java/Nyx.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Creates a Nyx object to initialize the chatbot and run it.
+     * @param args The command line arguments.
+     */
+    public static void main(String[] args) {
+        new Nyx("data", "nyx.txt").run();
+    }
+}
```
rohit0718: perhaps you can standardize all files to include a new line!

src/main/java/nyx/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a task with a deadline.
+ * This class inherits from the parent abstract Task class.
+ */
+public class Deadline extends Task {
+    private static final String DATETIME_FORMAT = "yyyy-MM-dd H:m";
+    private final LocalDateTime by;
```
rohit0718: perhaps a better name for the variable might be `endDateTime`. it is not immediately obvious what `by` refers to!

src/main/java/nyx/Event.java
```diff
@@ -0,0 +1,53 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents an event that occurs at a specific date and time.
+ * This class inherits from the parent abstract Task class.
+ */
+public class Event extends Task {
+    private static final String DATETIME_FORMAT = "yyyy-MM-dd H:m";
+    private final LocalDateTime at;
```
rohit0718: same for this, perhaps can be `startDateTime`!

src/main/java/nyx/Storage.java
```diff
@@ -0,0 +1,96 @@
...
+        this.data = new File(Paths.get(folderName, fileName).toString());
+    }
+
+    private Task lineToTask(String line) throws NyxException {
+        String[] splitLine = line.split(", ");
+
+        boolean isDone = splitLine[1].equals("1");
+
+        Task task;
+
```
rohit0718: same comment as above, perhaps can remove the spacing here.

rohit0718 (commented):
Overall coding style was great! Only have some minor comments!

src/main/java/nyx/Task.java
g4ryy: Hi @rohit0718, thanks for pointing this out! Indeed, my intention is to make this method package private since it is not used outside the package.

src/main/java/nyx/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Constructs a deadline task with its description, deadline and an indicator of whether it is marked as done.
```
justintanyf: The javaDocs convention said that we should leave a gap between the description and @param

justintanyf (commented):
Overall an excellent job!

PR #215, opened 2021-08-22
[Tey Hao Ze] iP

Haozeee:
# Duke

> The key to success is having a good task manager. - Unknown

Duke frees your mind from having to remember all your trivial tasks. It's,
* Text-based
* Lightweight
* Easy to learn
* ~~FAST~~ SUPER FAST to use
* **COMPLETELY FREE OF CHARGE!!!**

Here are some of the features of Duke:
- [x] Store different types of tasks such as deadlines or events.
- [x] Search for tasks based on keywords
- [x] Features to manage your task list easily
- [ ] Search for tasks by date (Coming soon)
- [x] Cool GUI (Coming soon)

If this sounds like something you want, all you need to do is,
1. Download it from [here](https://github.com/Haozeee/ip).
2. Double-click it.
3. Add some tasks.
4. Let it manage your tasks for you :satisfied:

---
If you are a Java programmer, you can use it to practice Java too. Here's the `main` method.
```java
public class Launcher {
    }
}
```


src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param userInput The user input that triggers the command.
+     * @param typeToAdd The type of task that should be added.
+     */
+    public AddCommand(String userInput, Task.Type typeToAdd) {
+        this.userInput = userInput;
+        this.type = typeToAdd;
+    }
+
+    public abstract boolean verifyAddCommand(String input);
```
peilinye, line 47: Perhaps a header comment could be added for this public method and the other 2 below?

src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,55 @@
+package commands;
+
+import duke.Ui;
+import tasks.Task;
+
+/**
+ * An abstract class that represents commands given by the
+ * user that involves adding a task to the taskList.
+ */
+public abstract class AddCommand extends Command {
```
peilinye: I like how you made the `AddCommand` class abstract and used `AddEventCommand`, `AddDeadlineCommand`, etc.

src/main/java/commands/AddDeadlineCommand.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param input The input by the user that triggers this command.
+     * @param taskList The taskList for Duke.
+     */
+    public AddDeadlineCommand(String input, TaskList taskList) {
+        super(input, Task.Type.DEADLINE);
+        this.taskList = taskList;
+    }
+
+    @Override
```
peilinye, line 26: Perhaps a header comment here to briefly describe what this method does would be good? Could add in header comments on other public methods as well.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.ArrayList;
+
+/**
+ * A class that abstracts a list of tasks set by the user. It also contains methods
+ * that pertain to updating the list of tasks.
+ */
+public class TaskList {
+
+    /** An arraylist that contains the tasks set by the user. */
+    private final ArrayList<Task> taskArrayList = new ArrayList<>();
```
peilinye: Perhaps a more intuitive variable name here? such as 'tasks'

src/main/java/commands/AddEventCommand.java
```diff
@@ -0,0 +1,64 @@
...
+            return false;
+        }
+        String[] inputParts = input.split(" -at | /at ");
+        if (inputParts.length != 2) {
+            this.setExecutionMessage(this.getInvalidArgumentsMessage());
+            return false;
+        }
+        return true;
+    }
+}
```
peilinye, line 67: I like how the different add commands parse the user input and check for input errors, the code is clean this way. :)

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,156 @@
...
+    /** The number of uncompleted tasks in the task list */
+    private int uncompletedTasks = 0;
```
peilinye: Perhaps a more intuitive variable name here such as 'numOfUncompletedTasks'? Same for 'totalTasks'

peilinye (commented):
Overall, your code is neat and easy to read. There were not many coding standard violations, but you could consider adding more header comments for public methods.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,119 @@
...
+     * Returns the string representation of the type of task.
+     *
+     * @return The required string representation.
+     */
+    public String eventTypeToString() {
+        if (this.type.equals(Type.TODO)) {
+            return "[T]";
+        } else if (this.type.equals(Type.EVENT)) {
+            return "[E]";
+        } else {
```
dhshah1: Very small suggestion: Perhaps it would be better to still keep an else if clause here, and then handle the else case when it doesn't match any type as well. It's unlikely that this will be an issue but maybe its better to be defensive here in case in the future there are more task types and this method is not updated by mistake

```suggestion
        } else if (this.type.equals(Type.DEADLINE) {
```

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,119 @@
...
+     * of date.
+     *
+     * @return The String format to save the task as.
+     */
+    public abstract String taskSaveString();
+
+    /**
+     * Marks a task as completed.
+     */
+    public void setAsFinished() {
```
dhshah1: Perhaps this method could be named as setAsDone or markAsDone? Just so that the terminology is consistent instead of having "finished" and "done".

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,119 @@
...
+        String typeOfTask = strComponents[0].strip();
+        boolean isDone = strComponents[1].strip().equals("1");
+        int dukeDateType = Integer.parseInt(strComponents[2].strip());
+        String taskName = strComponents[3].strip();
+        Task loadedTask;
+        if (typeOfTask.equals("T")) {
+            loadedTask = Todo.newTodoTask(taskName);
+        } else if (typeOfTask.equals("D")) {
+            loadedTask = new Deadline(taskName, DukeDate.getDukeDateFromType(strComponents[4].strip(), dukeDateType));
+        } else {
```
dhshah1: Same as above, perhaps to be defensive this can also be changed to an else if clause

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Returns a task from the String used to saved the task in local storage. When a task is saved to local
+     * storage it is saved as a formatted string that stores the information of the task. This methods takes
+     * that String and converts it back to the correct task.
+     *
+     * @param saveString The String that represents the saved task.
+     * @return The task converted from saveString.
+     */
+    public static Task stringToTask(String saveString) {
```
dhshah1: Maybe this method could be renamed to something like storageStringToTask, just to make it clear that its specifically talking about parsing the storageString format

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Marks a task in the task list as completed based on the index given.
+     *
+     * @param index The index of the task list. Note that the index provided starts from 1.
+     *              So the index 1 represents the first task in the taskList ArrayList.
+     * @return The message that should be shown to the user after a task has been marked as done.
+     */
+    public String markTaskAsCompleted(int index) {
```
dhshah1: Perhaps the terminology for a completed task could be standardised? Currently there's a few terms used to refer to this: "finished", "done", "completed".

dhshah1 (commented):
Overall, I felt the code was really well written and organised. Honestly, I learnt quite a lot from going through the code, so there weren't really any major coding quality issues I picked up on. Just left some minor comments below for your consideration.

PR #219, opened 2021-08-22
[Sourabh] iP

srj31:
# DukePro (v1.0)

> Java is to JavaScript what car is to Carpet. 
> – Chris Heilmann

DukePro frees your mind of having to remember things you need to do. It has 

* a **text based** interface
* an *intuitive* design

All you need to do to get it,

1. clone the repo or get the jar file (use [this](https://www.google.com/) to see how to use jar files)
2. double click
3. start adding tasks :smiley:

And ~pay $10~ it's **FREE!!!**

Features: 

- [x] managing tasks (todos, events, deadlines)
- [ ] Reminders (coming soon)

If you know how to program in Java, can edit the files in `Duke.java` and make changes in the `main`  method as per the changes needed: 

```java

public class Duke {
     public static void main(String[] args) throws DukeException, IOException {
        Duke duke = new Duke();
        duke.runDuke();
    }
}

```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,23 @@
+package duke;
+
+/**
+ *
+ */
```
willyamped: I think there is a missing javadocs header here? 😄

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,27 @@
...
+public class Event extends Task {
+    protected String time;
+
+    public Event(String description, String time) {
+        super(description);
+        this.time = time;
+    }
+
+    /**
+     * To get the time of the event stored in the list
```
willyamped: Javadoc headers should start with words in the form `Returns ...`, `Sends ...`, `Adds ...` (not `Return` or `Returning` etc.)
This applies to your other javadoc headers too :)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+            FileWriter writer = new FileWriter(file, false);
+            for(Task task: Tasklist.dukeList) {
+                String commandLine = "";
+                if(task instanceof Deadline) {
+                    commandLine = "D | " + (task.getIsDone() ? "1 | " : "0 | ") + task.getDescription() + " | " + ((Deadline) task).getTime() + '\n';
+                } else if(task instanceof Event) {
+                    commandLine = "E | " + (task.getIsDone() ? "1 | " : "0 | ") + task.getDescription() + " | " + ((Event) task).getTime() + '\n';
+                } else {
+                    commandLine =  "T | " + (task.getIsDone() ? "1 | " : "0 | ") + task.getDescription() + '\n';
```
willyamped: I think this chunk is not very reader-friendly 😬? You could try splitting them up into a few lines. Remember to follow the convention (e.g the number of tabs, etc)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,153 @@
...
+            printLines();
+        }
+
+    }
+
+    /**
+     * Remove a Task from the list
+     * @param taskIndex the index of the Task to be removed
+     */
+    void deleteTask(int taskIndex) {
```
willyamped: I like that you create these separate methods 👍

willyamped (commented):
LGTM. Overall, the code is good except for some minor things

src/main/java/duke/Deadline.java
calvintanwj: Agreed, seems like it's walked off somewhere. 😄

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,51 @@
+package duke;
+
+import duke.exceptions.DukeException;
+
+import java.io.*;
```
calvintanwj: When you import your classes, I think it would be better to list them explicitly.
```suggestion
import java.io.IOException;
import ...
```
This adheres to the coding standards, which should be the same for your tP. Additionally, it gives excellent documentation value for the team. I believe for IntelliJ, if 3 or more classes are imported from the same package, IntelliJ automatically converts it into a wilcard import, this limit can be toggled or turned off within your IDE settings.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,51 @@
...
+        duke.runDuke();
+    }
+
+
+    public void runDuke() throws DukeException, IOException {
+        storage.fetchData();
+        ui.displayWelcomeMessage();
+        String command = sc.next();
+        String description = sc.nextLine();
+        while(!command.equals("bye")) {
```
calvintanwj: I believe there should be a space between your while keyword and condition.
```suggestion
        while (!command.equals("bye")) {
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,51 @@
...
+            handleCommandExecution(command,description);
+            command = sc.next();
+            if(!command.equals("bye")) {
```
calvintanwj: Same as above.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,51 @@
...
+                description = sc.nextLine();
+            }
+            storage.saveData();
+        }
+        ui.displayByeMessage();
+    }
+
+    public void handleCommandExecution(String command, String description)throws DukeException {
```
calvintanwj: Same as above. Missing some spaces. 
```suggestion
    public void handleCommandExecution(String command, String description) throws DukeException {
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class Parser {
+
+    /**
+     * Splits the input line into description and time for the deadline
+     * or event
+     * @param string    the input line
+     * @return          array containing the description and time info
+     */
+    public static String[] splitDescriptionAndTime(String string) {
```
calvintanwj, line 20: Not too sure if I quite like the naming convention for the parameter, string. 
May I suggest something like "inputString".

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,42 @@
...
+        return string.split("/");
+    }
+
+    /**
+     * Extracts the description after splitting using splitDescriptionAndTime
+     * @param description_and_time  array after splitting the input line
+     * @return                      the description of the Task
```
calvintanwj: For javadoc comments, I believe there should be a spacing between your description of the method and the params, like:
```suggestion
     * Extracts the description after splitting using splitDescriptionAndTime
     *
     * @param description_and_time  array after splitting the input line
     * @return                      the description of the Task
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,85 @@
...
+
+        File savedFile = new File("./data/saved-tasks.txt");
+        if(!savedFile.exists()) {
+            savedFile.getParentFile().mkdirs();
+            savedFile.createNewFile();
+        }
+        Scanner scanner = new Scanner(savedFile);
+        while (scanner.hasNextLine()) {
+            String[] keywords = scanner.nextLine().split(" \\| ");
+            Task cur = null;
```
calvintanwj, line 62: Perhaps instead of "cur", "currTask" may be better.

src/test/java/duke/DukeTest.java
```diff
@@ -0,0 +1,22 @@
...
+import duke.exceptions.DukeEmptyTodoDescriptionException;
+import duke.exceptions.DukeUnknownCommandException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class DukeTest {
+
+    @Test
+    public void testWrongCommand() {
```
calvintanwj: I think that this method naming convention is fine, an possible alternative you and your team could use: is **featureUnderTest_testScenario_expectedBehavior()**

calvintanwj (commented):
Overall, really good stuff 👍 Just need to be a bit careful of those spaces and indentations, especially for your if and while blocks. I forget about those sometimes as well, perhaps you can configure your editor (File > Settings > Editor > Code Style > Java) to follow the coding standards. This way, you just need to run the reformat code (ctrl + alt + j) before you commit your changes. There are also a few other cosmetic issues like javadoc comments and naming conventions for methods and arguments, but these are rather minor. All the files are packaged nicely, with most of the methods having descriptive comments.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,27 @@
+package duke;
+
+/**
+ * duke.Event class to create Events for the duke list
```
luffingluffy: I think that you are missing full stops for the JavaDocs; similarly for the rest of the code.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,51 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    /**
+     * If the task is marked done X is shown
```
luffingluffy: Maybe it's better to write actively; "Shows X if the task is marked done"

src/main/java/duke/Tasklist.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Add a task to the list of tasks.
+     * @param task - the task which is to be added in the list
+     */
+    public static void add(Task task) {
+        dukeList.add(task);
+    }
+
+    /**
+     * Removing a task from the list of the tasks.
```
luffingluffy: Similar to the rest, maybe it is better to write as "Removes a task.."

luffingluffy (commented):
I think can improve some minor JavaDocs formatting.

PR #227, opened 2021-08-22
[Pham Chau Giang] iP

pcgiang:
### Duke

> "If you spend too much time thinking about a thing, you'll never get it done." - Bruce Lee ([source](https://dansilvestre.com/productivity-quotes/))

No longer need to remember things you need to do, Duke is:

- text-based
- super easy to learn
- ~~FAST~~ SUPER FAST to use!!

What you need to do is:
1. Download Duke from [here](https://nus-cs2103-ay2122s1.github.io/website/schedule/week4/project.html)
2. Click twice!
3. Add your tasks.
4. Trust Duke and we will manage your tasks for you! 👯 

And it is **FREE**!!

Features:

- [x] Manage tasks
- [x] Manage deadlines and time for events
- [ ] Reminders (coming soon)

If you wish to practise Java with our app, here's the our `main` method:
```
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+            storage.updateFile(list);
+            input = sc.nextLine();
+        }
+
+        ui.printByeMessage();
+        sc.close();
+    }
+}
```
over-fitted: The if and elseif chunks seem to make the code harder to read. Could the nesting be reduced, such as by using guard clauses and `break` to eliminate the need for `else`?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+            file.delete();
+            File newFile = new File(filePath);
+            newFile.createNewFile();
+            FileWriter writer = new FileWriter(file);
+
+            for (Task task : list.getList()) {
+                writer.write(task.writeTask() + "\n");
+            }
+            writer.close();
+        }
```
over-fitted: Could this block be refactored so as to remove the need for the else block and reduce indentation?
```suggestion
        if (!file.exists()) {
            file.createNewFile();
            return;
        }
        file.delete();
        File newFile = new File(filePath);
        newFile.createNewFile();
        FileWriter writer = new FileWriter(file);

        for (Task task : list.getList()) {
            writer.write(task.writeTask() + "\n");
        }
        writer.close();
```

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,46 @@
...
+        done = true;
+    }
+
+    /**
+     * A method that prints out details of a task.
+     *
+     * @return Details of task
+     */
+    public String writeTask() {
+        return (done ? "1" : "0") + " | " + title;
```
over-fitted: Could this be written more explicitly to improve readability?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,190 @@
...
+        int index = Integer.parseInt(i);
+        if (index < 0 || index > list.size()) {
+            throw new OutOfBoundException();
+        } else {
+            Task task = list.get(index - 1);
+            task.markAsDone();
+            System.out.println("Nice! I have marked this task as done!");
+            System.out.println(task);
+            return task;
+        }
```
over-fitted: Could this logic block be refactored using a guard clause to handle the exception throwing in order to reduce indentation caused by the else block?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,190 @@
...
+    public Task addTask(String str) throws EmptyDescriptionException
+            , InvalidTaskException, InvalidDeadlineException {
+        //first check if the task only contain 1 word
+        if (str.split(" ").length == 1) {
+            //check if task if valid
+            if (str.startsWith("todo") || str.startsWith("deadline") || str.startsWith("event")) {
+                throw new EmptyDescriptionException(str);
+            } else { //check for invalid task
+                throw new InvalidTaskException();
+            }
```
over-fitted: Could this else be removed?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                }
+
+                System.out.println("Got it. I've added this task.");
+                System.out.println(list.get(list.size() - 1));
+                System.out.println("Now you have " + list.size()
+                        + (list.size() == 1 ? " task in the list" : " tasks in the list."));
+                return task;
+            }
+        }
```
over-fitted: Could the else statements be removed to improve readability?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,190 @@
...
+            throw new OutOfBoundException();
+        } else {
+            Task task = list.get(index - 1);
+            list.remove(index - 1);
+            System.out.println("Noted. I've removed this task: ");
+            System.out.println(task);
+            System.out.println("Now you have " + list.size()
+                    + (list.size() == 1 ? " task in the list" : " tasks in the list."));
+            return task;
+        }
```
over-fitted: since the if statement to throw the exception is a guard clause already, could the else statement be removed?

over-fitted (commented):
I like the comprehensiveness of your functional tests!

Just need to take another look at reducing the complexity in your code through reducing else block usage in multiple areas, in addition to a couple other flagged areas where complexity could be reduced. Further, it may prove useful for future maintainability to follow git commit title conventions. Otherwise, your PR for the most part LGTM! :tada:

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Constructor for Deadline class.
+     *
+     * @param title Name of the task to be created.
+     * @param deadline Date of the deadline.
+     */
+    public Deadline(String title, String deadline) throws InvalidDeadlineException {
+        super(title);
+        this.type = "D";
```
ananya-ravi-3021: Perhaps you may not need the 'this' keyword here, since you are not assigning the instance variable 'type' to a parameter variable of the same name.

```suggestion
        type = "D";
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+                        task = new ToDos(stuff[2]);
+                        break;
+                case "D":
+                        task = new Deadline(stuff[2], stuff[3]);
+                        break;
+                case "E":
+                        task = new Events(stuff[2], stuff[3]);
+                        break;
+                default:
+                        throw new InvalidTaskException();
```
ananya-ravi-3021: Should there be a break statement here after throwing the InvalidTaskException?

```suggestion
                        throw new InvalidTaskException();
                        break
```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * A class that represents a list of Tasks.
+ *
+ * It contains a constructor, a getter,
+ * a method to print a list, add a task to a list,
+ * mark a task as done and delete a task.
+ *
+ */
+public class TaskList {
+    private ArrayList<Task> list;
```
ananya-ravi-3021: Perhaps it would be better to name this variable tasks? Since it is a collection of Task objects.

```suggestion
    private ArrayList<Task> tasks;
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,81 @@
...
+        } catch (IOException e) {
+            System.out.println("File failed to load.");
+        }
+    }
+
+    /**
+     * A method that allows user give input and runs the program.
+     * Contains a parser which reads command from the input.
+     */
+
```
ananya-ravi-3021, line 44: Perhaps there should be no spaces between the documentation block and the method.

src/test/java/duke/test/DukeTest.java
```diff
@@ -0,0 +1,15 @@
+package duke.test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
ananya-ravi-3021, line 2: Perhaps be more consistent with the ordering of imports, with static imports only above or below non-static imports.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
+package duke;
+
+import duke.exception.*;
```
ananya-ravi-3021, line 3: The asterisk in this import implies it is not explicit. Should it be changed to an explicit dukeException import?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * A method to mark a task as done.
+     *
+     * @param str User input which indicates details of the task to be marked as done.
+     * @return Task after it is marked as done.
+     * @throws OutOfBoundException Thrown when user inputs task number that is out of the given task range.
+     * @throws NumberFormatException Thrown when user gives a non-number as the task number.
+     */
+    public String markAsDone(String str) throws OutOfBoundException, NumberFormatException {
+        String i = str.substring(str.length()-1);
```
ananya-ravi-3021: Should the operator be surrounded with a space?
```suggestion
        String i = str.substring(str.length() - 1);
```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,189 @@
...
+                        + list.get(list.size() - 1).toString() + "\r\n"
+                        + "Now you have " + list.size()
+                        + (list.size() == 1 ? " task in the list" : " tasks in the list.");
+                return result;
+            }
+        }
+    }
+
+    /**
+     * A method to delete a certain in a TaskList.
```
ananya-ravi-3021, line 90: Perhaps an empty line between the parameter/return/throws section and the description?

Banana3021 (commented):
Overall, your program structure was excellent. The code is very easy to read and understand, and your documentation is very detailed. I found a small amount of coding standard violations, such as Javadoc comment spacing, and needing to include a 'break'  in the default component of the case statement.

Otherwise, really good job :)

src/main/java/duke/Storage.java
pcgiang: yep that's a good suggestion, thanks!

src/main/java/duke/Storage.java
pcgiang: I think there shouldn't be a break here since an exception is thrown? So the code wont be able to reach line 61 if im not wrong

PR #243, opened 2021-08-23
[Shee Hui] iP

sheehui:
# Duke

> Someday you will be old enough to read fairytales again ~C. S. Lewis ([more](https://www.reddit.com/r/books/comments/1jiop4/what_does_the_cs_lewis_quote_some_day_you_will_be/))

Duke frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn
- **super fast**

All you need to do it:
1. download it
2. run it in your local environment
3. add your tasks
4. start working towards your dreams!

It is all ***FREE***!!!! :confetti_ball: :partying_face: :tada:

Features:
- [X] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders

```java
public class Main {
    }
}
```


src/main/java/duke/commands/Command.java
```diff
@@ -0,0 +1,21 @@
...
+import duke.util.Ui;
+
+/**
+ * Command class that represents each user command.
+ */
+public abstract class Command {
+    /**
+     * Executes the user's command.
+     * @param tasks Current tasklist.
+     * @param ui User interface of Duke.
```
tau-bar: I think you should include an empty line between the description and parameter section based on [this](https://se-education.org/guides/conventions/java/intermediate.html#comments).

```suggestion
     * Executes the user's command.
     * 
     * @param tasks Current tasklist.
     * @param ui User interface of Duke.
```

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * The Deadline class that represents a task with a starting datetime.
+ */
+public class Deadline extends Task {
+
+    protected LocalDate when;
+
+    /**
+     * Constructor to create a Dedline task.
```
tau-bar:
```suggestion
     * Constructor to create a Deadline task.
```

src/main/java/duke/tasks/Event.java
```diff
@@ -0,0 +1,48 @@
...
+     * @throws UserInputError
+     */
+    public Event(String description, String when, boolean done) throws UserInputError {
+        super(description, Task.Type.EVENT, done);
+        try {
+            this.when = LocalDate.parse(when.trim());
+        } catch (DateTimeException e) {
+            throw new NoDescriptionException("Wrong datetime format");
+        }
+    }
```
tau-bar: I think that the parameter naming can be named to sound more like a boolean.
```suggestion
    public Event(String description, String when, boolean isEventDone) throws UserInputError {
        super(description, Task.Type.EVENT, isEventDone);
        try {
            this.when = LocalDate.parse(when.trim());
        } catch (DateTimeException e) {
            throw new NoDescriptionException("Wrong datetime format");
        }
    }
```

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,197 @@
...
+        this.isDone = done;
+    }
+
+    /**
+     * Creates a new task according task type.
+     *
+     * @param input String containing task details.
+     * @param type The enum type of the task.
+     * @return A Task object.
+     * @throws UserInputError
```
tau-bar: I think you can put the condition when the error is thrown
i.e. ```* @throws UserInputError If ...```

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,197 @@
...
+            SEPARATE,
+            done,
+            SEPARATE,
+            this.description,
+            SEPARATE
+        );
+    }
+
+    /**
+     * Return a formated icon to symbolise if task is completed.
```
tau-bar:
```suggestion
     * Return a formatted icon to symbolize if task is completed.
```

src/main/java/duke/tasks/TaskList.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Find the number of Tasks.
+     *
+     * @return
+     */
+    public int length() {
+        return taskList.size();
+    }
```
tau-bar: I think you should put what the method returns after the ```@return``` tag.
```suggestion
    /**
     * Find the number of Tasks.
     *
     * @return The number of tasks in the task list.
     */
    public int length() {
        return taskList.size();
    }
```

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,158 @@
+package duke.util;
+
+import duke.Duke;
+import duke.commands.*;
```
tau-bar: I think you should explicitly state the imports as stated [here.](https://se-education.org/guides/conventions/java/intermediate.html#package-and-import-statements)

tau-bar (commented):
LGTM, however I noticed some minor coding standard violations, which I have pointed out. I'm impressed at the depth of documentation in your code, and the abstraction of each Command into their own class. Good job! 👍🏽

PR #251, opened 2021-08-23
[Zhang Zhiyao] iP

zzybluebell:
**### Instruction** 😃

zzyBluebell 's Duke frees your mind of having to remember things you need to do. It's,

text-based
easy to learn

*FAST SUPER FAST to use*

All you need to do is,

1.   download GUI version V0.2.
2.   Open terminal.
3.   Use <cd> command to navigate to the directory of jar file.
4.   Enter java -jar duke.jar (Prerequisites: jdk installed).
5.   add your ### tasks.
6.   let it manage your tasks for you
7.   And it is FREE!

Features:

[ ]  Managing `Todo` Tasks
[ ]  Managing `Event` Tasks
[ ]  Managing `Deadline` Tasks
[ ]  `Find` specific Tasks

Examples:
-  todo read book 

-  event meeting /at 2021-08-23 17:00
-  deadline return book /by 2021-08-20 12:00
-  find book

If you Java programmer, you can use it to practice Java too. Here's the Launcher class and main method:
``` java
/**
 * A launcher class to workaround classpath issues.
 */
public class Launcher {

    /**
     * The main method for launch the Main class
     * 
     * @param args
     */
        launch(Main.class, args);
    }
}
```
 [Project details can be viewed here](https://github.com/zzybluebell/ip)


src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,72 @@
+package Duke;
```
ykwei7:
```suggestion
package duke;
```

src/main/java/Duke/Exceptions/DeleteWrongIndexException.java
```diff
@@ -0,0 +1,14 @@
+package Duke.Exceptions;
```
ykwei7: Might be easier to put all the DukeExceptions in a single Java file, and extend by doing public static class for sub-class exceptions

ykwei7 (commented):
Looks good overall. 👍

src/main/java/Duke/Tasks/Find.java
```diff
@@ -0,0 +1,32 @@
...
+import Duke.Tool.TaskList;
+import Duke.Ui;
+
+/**
+ * Represents the Find class
+ */
+public class Find extends Task {
+
+    protected String description;
+    /**
```
ramapriyan912001: Perhaps it would be better for the readability of the code if you left a line space between the end of the field declarations and the constructor?

text-ui-test/input.txt
```diff
@@ -0,0 +1,10 @@
+todo eat
+todo cs2103 tut
```
ramapriyan912001, line 2: Perhaps it would be good to test other commands such as delete too?

ramapriyan912001 (commented):
Overall looks very good!
Perhaps the only minor change I might suggest would be to incorporate an empty line space between the declaration of fields and the declaration of the first constructor.

text-ui-test/input.txt
mingyi456: `find()` would also be good to test I guess?

data/duke.txt
```diff
@@ -0,0 +1,5 @@
+T | 1 | read book
+D | 0 | return book | 2021-08-12 15:30
+E | 0 | project meeting | 2021-08-22 21:30
+T | 1 | join sports club
+T | 0 | work
```
mingyi456, line 5: Not sure why there seems to be 2 data files. Maybe one of them should be deleted?

src/main/java/Duke/Duke.java
mingyi456: I believe packages should be named with all lowercase or something, so `duke` and not `Duke`?

src/main/java/Duke/Exceptions/DeleteWrongIndexException.java
mingyi456: Also, just like your `Duke` package, your `Exceptions` package should be named in lower case, i.e. `exceptions`

src/main/java/Duke/Exceptions/DeleteWrongIndexException.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Represents DeleteWrongIndexException class
+ */
+public class DeleteWrongIndexException extends Exception {
+    /**
+     * The Constructor for DeleteWrongIndexException
+     * @param message
```
mingyi456: I think there needs to be a description for the parameter `message` in this JavaDocs line. This appears to be repeated many times, especially in all your `Exception` classes.

src/main/java/Duke/Exceptions/EmptyTaskListException.java
```diff
@@ -0,0 +1,14 @@
+package Duke.Exceptions;
+
+/**
+ * Represents EmptyTaskListException class
+ */
+public class EmptyTaskListException extends Exception{
+    /**
+     * The Constructor for EmptyTaskListException
+     * @param message
```
mingyi456: Please see previous comment regarding JavaDocs.

src/main/java/Duke/Tasks/Task.java
```diff
@@ -0,0 +1,82 @@
+package Duke.Tasks;
+
+import Duke.Tool.Storage;
+import Duke.Tool.TaskList;
+import Duke.Ui;
+
+
+/**
+ * Represents the Task class
```
mingyi456: It would be good to provide more descriptive details in this JavaDocs line, so that reviewers can understand this class better. Currently, I think this is rather vague.

src/main/java/Duke/Tool/Parser.java
```diff
@@ -0,0 +1,259 @@
...
+                    }
+
+                }
+
+            default:
+                throw new NoCommandException(instruction.name());
+            }
+        }
+
+        else {
```
mingyi456: I feel there are too many empty lines in this chunk of code, which affects readibility as a lot more scrolling is needed. Consider cutting down some of them? I noticed this pattern happening elsewhere, but this part is much more pronounced, so I believe you might be adhering to your own internal convention regarding how many blank lines to leave. Even so, I think breaking this convention is worth it if there are going to be this many empty lines.

src/main/java/Duke/Tool/Parser.java
```diff
@@ -0,0 +1,259 @@
...
+        } catch (NoDescriptionException | EmptyTaskListException | NoCommandException | NoTimeException |
+                DeleteWrongIndexException | WrongTimeFormatException e) {
+
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+
+
```
mingyi456: Again, I feel there are too many blank lines here.

mingyi456 (commented):
Overall, the code quality is very much decent, with a few minor nits. I see almost every JavaDocs is too skeletonised, with almost no description of parameters if any. It might be better to have fewer, but more developed JavaDocs, and this would also fit the module requirements better. Also, I think the number of empty lines can be cut down.

PR #254, opened 2021-08-23
[Tianran] iP

Nature711:
# DinoPro

DinoPro frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/Nature711/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```ruby
public class Main {
    }
}
```


src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,81 @@
...
+import dino.util.Storage;
+import dino.util.Ui;
+
+
+/**
+ * Represents a Personal Assistant ChatBot name Dino
+ * It helps the user to keep track of various things
+ */
+public class Dino {
+    private final TaskList taskList;
```
calvintanwj, line 15: Why not call it "tasks" instead of "taskList"? :)

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,81 @@
...
+            Command cmd = new FindCommand(command);
+            return cmd.execute(storage, taskList);
+        }
+        case BYE: {
+            return ui.exit(storage, taskList);
+        }
+        default: {
+            throw new InvalidInputException();
+        }
+        }
```
calvintanwj: Very nice! Easy to understand and readable.

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,81 @@
+package dino;
+
+import dino.command.*;
+import dino.exception.*;
```
calvintanwj, line 4: Perhaps it would be better to explicitly name your imports for better documentation.

src/main/java/dino/command/AddTaskCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    private final String taskString;
+    private final Parser.CMDTYPE taskType;
+
+    /**
+     * Constructs a AskTaskCommand object
+     *
+     * @param taskString the user input command
+     * @param taskType the type of the task, which can only be one of the type ToDo, Deadline, or Event
+     */
+    public AddTaskCommand(String taskString, Parser.CMDTYPE taskType) {
```
calvintanwj, line 31: hm, this is interesting. May I know why you decided to group the adding of tasks as one command, instead of separating them such as "AddListCommand", "AddDeadlineCommand"? ... I think as the code-base gets bigger, and if you do choose to implement more different types of tasks, this class and in particular the execute and getTaskDescription methods can get pretty messy.

src/main/java/dino/command/AddTaskCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Extracts the description of the task from a user input command to add task
+     *
+     * @param s the user input command
+     * @param type the type of the task
+     * @return the description of the task that is extracted from the user input command
+     * @throws EmptyTaskDescriptionException if the description of the task is empty
+     * @throws TimeNotSpecifiedException if the task is of type Deadline or Event but the time is not specified
+     */
+    public static String getTaskDescription(String s, Parser.CMDTYPE type) throws EmptyTaskDescriptionException, TimeNotSpecifiedException {
```
calvintanwj: Perhaps the parameter "s" can be named better, may I suggest something like "userInput" or "inputString"

src/main/java/dino/command/AddTaskCommand.java
```diff
@@ -0,0 +1,123 @@
...
+                return s.substring(type.toString().length() + 1, s.indexOf("/")).trim();
+            } else {
+                throw new TimeNotSpecifiedException(type.toString());
+            }
+        }
+        default: {
+            break;
+        }
+        }
+        return s;
```
calvintanwj: Why not shift the "return s;" into the default branch. It seems more readable to me. :) Additionally for switch statements, I believe you can remove the curly braces for each case.
```suggestion
        default: 
            return s;
            break;
        }

```

src/main/java/dino/command/AddTaskCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     *
```
calvintanwj, line 110: There seems to be a missing description here 😄

src/main/java/dino/command/Command.java
```diff
@@ -0,0 +1,15 @@
...
+import dino.exception.*;
+import dino.task.TaskList;
+import dino.util.Storage;
+
+/**
+ * Contains a set of instructions that can be understood and executed by the ChatBot
+ */
+public abstract class Command {
+
+    abstract public String execute(Storage storage, TaskList taskList) throws InvalidFormatException, TimeNotSpecifiedException, EmptyTaskDescriptionException, InvalidIndexException, TaskAlreadyDoneException, InvalidInputException, IndexNotSpecifiedException, EmptyListException, TaskNotFoundException;
```
calvintanwj, line 12: Perhaps you can have a parent class like DinoException, where all these exceptions extend from this DinoException class. Then the method signature could be throws DinoException instead, of having such a long method signature. I think it would be quite a headache, when you add more exceptions and will have to change the method signatures of your command classes.

src/main/java/dino/command/MarkCommand.java
```diff
@@ -0,0 +1,91 @@
...
+        switch (markType) {
+        case DONE: {
+            return taskList.markTaskDone(index);
+        }
+        case DELETE: {
+            return taskList.deleteTask(index);
+        }
+        default: {
+            throw new InvalidInputException();
+        }
```
calvintanwj: Same as before. I don't think there is a need for the curly braces.
```suggestion
        case DONE: 
            return taskList.markTaskDone(index);
        case DELETE: 
            return taskList.deleteTask(index);
        default: 
            throw new InvalidInputException();
```

src/main/java/dino/command/Command.java
calvintanwj: After reading further down, I realize you do have a DinoException, so why not just throws DinoException instead? :)

src/main/java/dino/task/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Prints out the task list in console, prefixed with index
+     *
+     * @return the output message after execution
+     * @throws EmptyListException if the current task list is empty
+     */
+    public String printTaskList() throws EmptyListException {
+        if (taskList.isEmpty()) throw new EmptyListException();
```
calvintanwj: Don't forget to adhere to egyptian brackets. This is readable to me, but best to adhere to coding standards :)
```suggestion
        if (taskList.isEmpty()) { 
            throw new EmptyListException();
        }
```

src/main/java/dino/task/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+        String list = "";
+        for (int i = 0; i < taskList.size(); i++) {
+            list += (i + 1) + ". " + taskList.get(i) + "\n";
+        }
```
calvintanwj: May I suggest using a StringBuilder instead, since String is immutable in Java, and concatenation creates a new String each time, resulting in O(n^2) time complexity. I think for simple string concatenation, using the '+' operator is fine, but because you are looping through all the tasks in your taskList, I think it can get quite costly.

src/main/java/dino/util/Storage.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+import dino.exception.*;
+import dino.task.*;
```
calvintanwj, line 15: Once again, I think it's better to use explicit import statements. I believe for IntelliJ, it automatically uses wildcard imports, if you have 3 or more imports from the same package. This limit can be changed or toggled off within your IDE settings.

calvintanwj (commented):
Overall, looks really good 👍  A few cosmetic issues here and there such as the curly braces in switch statements, and your import statements. Naming conventions for certain parameters can be improved as well (instead of using "s" and "t"). Other than that, the code is readable and structured really well.

src/main/java/dino/Dino.java
Moley456, line 4: Would it be better if your imported classes are explicitly stated for better documentation? This would also comply with the coding standards in the CS2103T website as well 😄

src/main/java/dino/command/AddTaskCommand.java
Moley456: Would it be better if the parameter "s" was named "userCommand" or "userInput"? This would allow others understand the code more easily.

src/main/java/dino/command/AddTaskCommand.java
Moley456: Hmm do you think it's better to just move the return into the default case and remove the break statement? I feel that it would make the code look cleaner! Also, the curly braces could be removed as well in each of the cases.

```suggestion
        default: 
            return s;
        }
```

src/main/java/dino/exception/DinoException.java
```diff
@@ -0,0 +1,30 @@
...
+
+
+
+
+
+
+
+
+
+
```
Moley456, line 30: You could probably remove the empty lines here 😅

src/main/java/dino/command/MarkCommand.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+
+    /**
+     * Extracts the index indicated by the done or delete command
```
Moley456: Small detail here but I think your description and parameter section should have an empty line separating them, according to the coding standards.

src/main/java/dino/task/TaskList.java
Moley456: According to the coding standards, I think there should be Egyptian brackets here instead of having everything on one line.

Moley456 (commented):
Everything looks good. Clean code with just some minor changes to be made to adhere to the coding standards. Keep it up! 👍🏻

PR #265, opened 2021-08-23
[Hern Ping] iP

hernpiblo:
# Pib

## What _Pib_ can do:

- ### Add tasks
  1. #### Todo : Normal tasks
      - `Todo read a book`
  2. #### Deadline : Tasks with a deadline
      - `Deadline SWE quiz /by 2021-09-01 2359`
  3. #### Events : Tasks with an event date
       - `Event hackathon /at 2021-10-10 1230`
- ### View entire list of tasks
- ### Mark tasks as completed
- ### Delete tasks
- ### Search for tasks
- ### Auto-save and load saved data

#### Snippet of code:
```java
public Pib() {
        this.ui = new Ui();
        this.list = new TaskList();
        try {
            this.storage = new Storage(DATA_FILE_PATH);
            storage.loadData(this.list);
        } catch (PibException e) {
            e.print();
        }
        this.parser = new Parser(this.list);
    }
```

## Get started
- [x] Download
- [x] Keep track of tasks
- [x] Enjoy
- [ ] Feedback

The source code is inspired by this quote:
> Keep it simple, stupid! - Kelly Johnson, lead engineer at the Lockheed Skunk Works


src/main/java/pib/Pib.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * pib is a Personal Assistant Chat-bot that is able to keep track of tasks (CRUD) and deadlines
+ */
+public class Pib {
+
+    /**
+     * String to locate data file to load/save data
+     */
+    public static final String DATA_FILE_PATH = "data/data.txt";
```
benedictchuajj, line 17: I like how there's a comment here to help understand this variable 👍

src/main/java/pib/Pib.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private Ui ui;
+    private TaskList list;
```
benedictchuajj, line 20: Perhaps consider changing the name to *tasks* so there's greater clarity of what this variable is supposed to be.

src/main/java/pib/gui/DialogBox.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static DialogBox getUserDialog(String text, Image img) {
+        return new DialogBox(text, img);
+    }
+
+    public static DialogBox getPibDialog(String text, Image img) {
+        var db = new DialogBox(text, img);
+        db.flip();
+        return db;
+    }
```
benedictchuajj, line 60: Perhaps consider adding some Javadocs here since they are public methods? 😄

src/main/java/pib/gui/MainWindow.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+    }
+
+    public void setPib(Pib p) {
+        pib = p;
+        dialogContainer.getChildren().add(DialogBox.getPibDialog(pib.loadData(), pibImage));
+    }
```
benedictchuajj, line 38: Small nit but perhaps you could add some Javadocs here too since they are public. Noticed the same issue in a few other places too.

src/main/java/pib/tasks/Deadline.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Public method to convert task to a string format used to save inside a .txt file
+     *
+     * @return string format of Deadline task to be saved
+     */
+    public String toDataString() {
+        return "D," + getIsDone() + "," + getDescription() + "," + date + "," + time + System.lineSeparator();
+    }
```
benedictchuajj, line 110: Perhaps you can improve readability here by using String.format()? 
Noticed the same issues in several other places that returns a task as a String.

src/main/java/pib/utility/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+        this.list = list;
+    }
+
+    /**
+     * Uses the scanner to read inputs from the user
+     */
+    public void readInput() {
+        scanner:
+        while (sc.hasNextLine()) {
+            try {
```
benedictchuajj, line 31: Any reason why you catch the exceptions in these Parser methods instead of throwing it to the main caller in Pib to catch? Since they all throw the same type of PibException, letting Pib catch these exceptions when it is calling these methods seems like a better control flow to me.

Just a nit from me and I might not be right about this 😃

src/main/java/pib/pibexception/PibException.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * Calls Ui.printError method to print the error message to user
+     *
+     * @return String containing response to be printed to user
+     */
+    public String print() {
+        return Ui.printError(errorType);
+    }
```
benedictchuajj, line 28: I think this might be breaking the encapsulation of PibException as Ui knows the types of errorTypes and has to refer to it to return the errorMessage when it is done this way. Perhaps consider moving the switch statement from Ui to here to pass in only the errorMessage to Ui?

benedictchuajj (commented):
Looks good with very few coding standard violations. 👍 
Just a few nits to further improve readability.

Well done! I appreciate the effort you have put in to your Javadocs and comments throughout the code that makes it easy for me to follow. 😃

rldefa (commented):
Very clean code! Good job, no coding standard violation spotted. Very impressive : )

PR #267, opened 2021-08-23
[Ravi Ananya] iP

Banana3021:
# Banana :banana:

## Welcome to Banana, the new and improved Duke chatbot!

> Why do you need to write down tasks when you've got a chatbot.

### Here's what Banana can do:
1. Store your task information
1. Retrieve your task information immediately
1. Access your task information
    1. Find a task
    1. Delete a task
    1. Edit a task to indicate that it has been completed
    1. Use a proper `<Tasklist>` element

### It's super cool! 
* User-friendly
* Really easy to run
* Checks off tasks in a satisfying way.

### The next list of features:
- [x] GUI
- [ ] More types of tasks.

### How to run the code
```javascript
    }
```


src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,352 @@
...
+ * The Duke class runs the duke/chatbot
+ * program.
+ *
+ * @author: Ravi Ananya
+ **/
+
+public class Duke {
+
+    private static final String WELCOME_LABEL = "Hello! I'm Banana \n" + "     What can I do for you?";
+    private static final String BYE_LABEL = "Bye. Hope to see you again soon!";
```
JunWei3112: Should these associated constants be prefixed by the common type name instead?
```suggestion
    private static final String LABEL_WELCOME = "Hello! I'm Banana \n" + "     What can I do for you?";
    private static final String LABEL_BYE = "Bye. Hope to see you again soon!";
```

src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,352 @@
...
+
+    private Storage store;
```
JunWei3112: Perhaps it will be better if a noun was used for this variable name, instead of using a verb (which is better suited for methods/functions) ? 
```suggestion
    private Storage storage;
```

src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,352 @@
...
+        }
+    }
+
+    public static void main(String[] args) {
+        /*String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);*/
```
JunWei3112: Could this chunk of commented out code (also known as **dead code**) be deleted, since they are redundant?

src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * The DukeException class
+ * throws specialised Duke exceptions.
+ *
+ * @author: Ravi Ananya
+ **/
+class DukeException extends Exception {
+
+    private String text;
```
JunWei3112: Perhaps it would be better if a more descriptive name (e.g. errorMessage) was used for this variable?

src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * This class handles deadline
+ * types of tasks.
+ *
+ * @author: Ravi Ananya
+ **/
+
+class Deadline extends Task {
+
+    protected String deadLine;
```
JunWei3112: Should this variable be renamed to **deadline** instead, since deadline is a single word and naming it deadLine may cause confusion?

src/main/java/banana/Parser.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * Checks how to handle the user's
+     * input command.
+     *
+     * @param tasks the list of tasks.
+     * @throws DukeException if the input was invalid.
+     */
+    public void useInput(TaskList tasks) throws DukeException {
```
JunWei3112: Since there are many ways in which input can be used, should a more descriptive name be used for this method so as to further specify the purpose of this method? One possible suggestion is **parseInput**

src/main/java/banana/Parser.java
```diff
@@ -0,0 +1,281 @@
...
+
+    /**
+     * Changes the date format to be
+     * parsed by LocalDate.
+     *
+     * @param date the old-format date.
+     * @return the date.
+     */
+    public String parseDates(String date) {
+        String[] sep = date.split("/");
```
JunWei3112: As readers may not be able to infer the purpose of this variable from the abbreviated variable name, perhaps it would be better if a lengthier name was used to describe this variable instead? I noticed this same issue in some other places as well.

src/main/java/banana/TaskList.java
```diff
@@ -0,0 +1,62 @@
...
+    public Task getTask(int index) {
+        return allTasks.get(index);
+    }
+    
+    /**
+     * Gets the size of the list.
+     *
+     * @return the size of the list.
+     */
+    public int size() {
```
JunWei3112: Should verbs be used in the naming of this method instead? One possible suggestion is 
```suggestion
    public int getSize() {
```

src/main/java/banana/Ui.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * The Ui class handles
+ * the user inputs.
+ *
+ * @author: Ravi Ananya
+ **/
+
+public class Ui {
+
+    protected Scanner userInput;
```
JunWei3112: Should another name be used in naming this Scanner variable instead? Naming it **userInput** may cause confusion to readers, since the purpose of this variable is to read and obtain user input, and readers may misinterpret that this variable is used to store user input instead.

JunWei3112 (commented):
Overall, I found that your code was quite easy to read and your methods also contain an appropriate number of lines of code (not too lengthy and nested). 
However, some of the classes may be getting too long. One example will be Duke.java, where classes that serve other purposes are also being declared in the same file. You can consider splitting it into different class files.
Also, do take extra note when naming variables or methods, so as to improve readability of the code. Otherwise, good job! 😁

src/main/java/banana/Duke.java
```diff
@@ -0,0 +1,251 @@
...
+     *
+     * @return the error message.
+     */
+    @Override
+    public String getMessage() {
+        return errorMessage;
+    }
+
+}
+
```
rgbpokka, line 99: Overall I feel the code is well organized into distinct methods, however it is a very large file. Perhaps the writeToFile method could be part of the Storage class instead?

src/main/java/banana/Parser.java
```diff
@@ -0,0 +1,294 @@
...
+     * @return the label.
+     */
+    public String addTaskCommand(String input, TaskList tasks) {
+        if (input.contains("todo")) {
+            String info = input.substring(5);
+            tasks.addTask(new ToDo(info));
+        } else {
+            tasks.addTask(new Task(input));
+        }
+        return displayLabel("Got it. I've added this task:  \n" +
```
rgbpokka: If i'm not wrong, the line should break before an operator (the "+" sign should go onto the start of the next line instead of being at the end of the line)

src/main/java/banana/Parser.java
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * Changes the date format to be
+     * parsed by LocalDate.
+     *
+     * @param date the old-format date.
+     * @return the date.
+     */
+    public String parseDates(String date) {
+        String[] sepIntoYearMonthDate = date.split("/");
```
rgbpokka: I like the clear variable name.

rgbpokka (commented):
Overall I think the code is really clean and easy to read! Keep up the good work :)

PR #283, opened 2021-08-23
[Lim Wei Chang] iP

weichang18:
### Duke is program that helps you record down your task similarly to writing in your diary. It's,

-  text-based
- easy to use and lean
- convenient

You just need to:

1. download it [here](https://github.com/weichang18/ip/releases/tag/A-Jar)
2. open it
3. input your task
4. and its saved

Best of all its **FREE**!😉

Features:

- [x]  Managing tasks

```ruby
public class Duke {
    public static void `main`(String[] args) {
        (new Duke("data/dukeTaskList.txt")).run();
    }
}
```

As David Allen said:
>Your mind is for having ideas, not holding them.


.gitignore
```diff
@@ -15,3 +15,32 @@ bin/
 
 /text-ui-test/ACTUAL.txt
 text-ui-test/EXPECTED-UNIX.TXT
+src/test/out/production/main/duke/Deadline.class
```
luminousleek, line 18: Not a code quality issue, but instead of having to manually add all the class files and whatnot into your `.gitignore` file, maybe can just do this:
```
/src/test/out/
/src/main/out/
```
and it should work, unless you have other stuff in the out subfolders that you want to get stored on Git. Then I guess you could do
```
src/test/out/production/main/duke/*.class
src/test/out/test/test/duke/*.class
src/main/out/production/main/duke/*.class
```

luminousleek: Might also want to add `/.idea/` to your `.gitignore`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,122 @@
...
+                    if (input.contains("todo")) {
+                        newTask = new Todo(parser.toDoInputParser(input));
+                    } else if (input.contains("deadline")) {
+                        if (!input.contains("/by")) {
+                            //for missing dateline
+                            throw new MissingDateException();
+                        } else {
+                            newTask = new Deadline(parser.deadlineInputTaskParser(input),
+                                    parser.deadlineInputDateParser(input));
+                        }
```
luminousleek: Quite deep nesting here - maybe you can put the block of code highlighted here into a new method called `deadlineInputParser`? Same thing with the event input below

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,122 @@
...
+                //Find tasks
+                String keyword = parser.findInputParser(input);
+                TaskList taskListWithKeyword = this.taskList.findTasks(keyword);
+                this.ui.findTaskMsg(taskListWithKeyword);
+            } else {
+                try {
+                    //Initialise the task if its a valid input.
+                    Task newTask = null;
+                    if (input.contains("todo")) {
+                        newTask = new Todo(parser.toDoInputParser(input));
```
luminousleek: You might want to consider making the various task parsers return a task? Might make your life easier.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Return the date of the deadline task when input "deadline... /by xxxx-xx-xx xxxx" command.
+     *
+     * @param input Inout of the user.
+     * @return The deadline task date.
+     */
+    public String deadlineInputDateParser(String input) {
+        String[] parts = input.split(" /by ");
+        return parts[1];
+    }
```
luminousleek: Wondering if you could combine both these methods together and return a String[], i.e.
```java
public String[] deadlineInputParser(String input) {
    String[] parts = input.split( " /by ");
    parts[0] = parts[0].replace("deadline ", "");
    return parts;
}
```
Similar thing for the event parser below

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class Storage {
+    private File taskListFile;
+    private Boolean doesFileExists;
```
luminousleek: Sorry to be a grammar Nazi, but I think it should be named `doesFileExist` or simply `fileExists`?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Returns the checkbox of the task.
+     *
+     * @return String representation of checkbox.
+     */
+    public String checkIsDone() {
+        return this.isDoneBox;
+    }
```
luminousleek: Does the method name sound like it would return `isDone` instead of `isDoneBox`? Maybe consider renaming it to `getIsDoneBox`?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Checks if the file specified exists.
+     *
+     * @return True if file exists and false if it does not.
+     */
+    public Boolean getDoesFileExists() {
+        return this.doesFileExists;
+    }
```
luminousleek: Is this method used anywhere else?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,100 @@
+package duke;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private int noOfTask = 0;
```
luminousleek: Consider changing to `numOfTask` or `numTasks`? The abbreviation No can be a bit confusing

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+                    + taskList.get(j).getTaskType()
+                    + taskList.get(j).checkIsDone()
+                    + " " + taskList.get(j).getDescription() + "\n";
+            str = str + listItem;
+        }
+        return str;
+    }
+    public int getNoOfTask() {
+        return this.noOfTask;
+    }
```
luminousleek: What's the difference between this method and the `taskListSize` method?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+    public void saveFile(TaskList taskList) throws IOException {
+        StringBuilder listBuilder = new StringBuilder();
+        for (int j = 0; j < taskList.getNoOfTask(); j++) {
+            String listItem = (j + 1)
+                    + "."
+                    + taskList.getTask(j).getTaskType()
+                    + taskList.getTask(j).checkIsDone()
+                    + " " + taskList.getTask(j).getDescription();
+            listBuilder.append(listItem).append(System.lineSeparator());
+        }
```
luminousleek: This seems like it would have the same output as `taskList.toString()`? Maybe can call that instead? Then change the `taskList.toString` method to use a `StringBuilder` instead.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    @Override
+    public String toString() {
+        String str = "";
+        for (int j = 0; j < taskList.size(); j++) {
+            String listItem = (j + 1)
+                    + "."
+                    + taskList.get(j).getTaskType()
+                    + taskList.get(j).checkIsDone()
+                    + " " + taskList.get(j).getDescription() + "\n";
```
luminousleek: Maybe implement a `Task.toString()` method so you won't have to do this?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,102 @@
...
+     * Prints add task message.
+     *
+     * @param task Task to be printed.
+     */
+    public void addTaskMsg(Task task, int noOfTask) {
+        System.out.print(LINEBREAK
+                + "Roger! I will add this task in: \n"
+                + task.getTaskType()
+                + task.checkIsDone()
+                + " "
```
luminousleek: Here you inserted a space in between the checkbox and the description, but in `deleteTaskMsg` you didn't?

src/main/java/duke/WrongInputException.java
```diff
@@ -0,0 +1,18 @@
+package duke;
+
+public class WrongInputException extends DukeException {
+
+    WrongInputException() {
```
luminousleek: No access modifier?

luminousleek (commented):
Good job on doing up the iP till this far! The only general comment I have is to try to avoid duplicated code - if you find yourself writing the same lines of code again you should probably wrap it in a method.

All the best with this week's tasks - do get started on JavaFX early because it'll take a while to do 😣

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Runs duke bot.
+     *
+     * @throws IOException If there is wrong input or output.
+     */
+    public void run() throws IOException {
+        this.isRunning = true;
+        this.ui.greetingMsg();
+        Scanner sc = new Scanner(System.in);
```
jiarong15: Maybe you can consider abstracting these reading of inputs into your UI class?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,122 @@
...
+                    Task newTask = null;
+                    if (input.contains("todo")) {
+                        newTask = new Todo(parser.toDoInputParser(input));
+                    } else if (input.contains("deadline")) {
+                        if (!input.contains("/by")) {
+                            //for missing dateline
+                            throw new MissingDateException();
+                        } else {
+                            newTask = new Deadline(parser.deadlineInputTaskParser(input),
+                                    parser.deadlineInputDateParser(input));
```
jiarong15: Agree that the method here is really long, abstract into a private method would be better 😄

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,50 @@
...
+    private LocalTime time;
+
+    /**
+     * Constructor for event task.
+     *
+     * @param description Description of event task.
+     * @param deadline Deadline of event task in "yyyy-mm-dd HHmm" format.
+     */
+    public Event(String description, String deadline) {
+        super(description, "[E]");
```
jiarong15, line 19: maybe for the "[E]" can make a private static final String attribute ?

jiarong15 (commented):
Generally okay just minor changes :)

PR #284, opened 2021-08-23
[Simon Julian Lauw] iP

simonjulianl:

* text-based
* easy to learn
All you need to do is,

1. download it from [here](google.com).
2. double-click it.
3. add your tasks.
And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)


src/main/java/dao/TaskDaoImpl.java
```diff
@@ -0,0 +1,127 @@
...
+import configuration.Setting;
+import model.Task;
+import storage.FileListStorage;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
```
SHEZADHASSAN22, line 14: I think your indentation levels are too big. Should be 4 spaces.

src/main/java/dao/TaskDaoImpl.java
```diff
@@ -0,0 +1,127 @@
...
+	}
+	
+	@Override
+	public int getSize() {
+		try {
+			ArrayList<Task> tasks = taskFileListStorage.readArrayListFromFile();
+			return tasks.size();
+		} catch (Exception e) {
+			logger.warning(e.getMessage());
+			return -1;
```
SHEZADHASSAN22: Perhaps used a named constant instead of the number.

src/main/java/dao/TaskDaoImpl.java
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+	
+	private boolean isInvalidIndex(int index) {
+		int size = getSize();
+		return size != -1 && (index >= getSize() || index < 0);
```
SHEZADHASSAN22: Maybe this expression can be simplified for easier understanding

auhc99: I agree, This expression might be confusing to people trying to understand your code.
Maybe you can replace it with:
```
private boolean isInvalidIndex(int index) {
    int size = getSize();
    if (size == -1) {
        return false;
    } else if (index >= size || index < 0) {
        return true;
    } else {
        return false;
    }
}
```
You should also consider moving the error checking for size out of this function.

src/main/java/Dude.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Runs Dude chatterbot.
+	 */
+	@SuppressWarnings("InfiniteLoopStatement")
+	public void run() {
+		String logo = " ____        ____      \n"
+				+ "|  _ \\ _   _|  _ \\____\n"
+				+ "| | | | | | | | |/ __ \\\n"
+				+ "| |_| | |_| | |_|   __/\n"
+				+ "|____/ \\__,_|____\\____|\n";
```
auhc99: Should this logo be extracted out out the function and be stored as a static final constant?

src/main/java/dao/TaskDao.java
```diff
@@ -0,0 +1,66 @@
+package dao;
+
+import model.Task;
+
+import java.util.List;
+
+/**
+ * Data access object (dao) for the Task model.
+ */
+public interface TaskDao {
```
auhc99, line 11: You may want to rename this class (and other classes containing the short form "Dao") to something more understandable.

src/main/java/dao/TaskDaoImpl.java
auhc99: Yes, you can change your IDE / text editor's settings to reconfigure your tab to be of 4 spacings.

src/main/java/parser/ParserImpl.java
```diff
@@ -0,0 +1,225 @@
...
+			if (timing.isBlank()) {
+				logger.warning("Empty deadline timing supplied");
+				throw new IllegalArgumentException("deadline timing cannot be empty");
+			}
+			
+			commandLogicUnit.processCommand(Command.DEADLINE, Map.of(
+					"description", desc,
+					"timing", timing
+			));
+		} catch (Exception e) {
```
auhc99: Perhaps you can specify which Exception you are handling, and not just generalise it to all exceptions.

auhc99 (commented):
Overall great job with complying with the java coding standards. I noted a few minor suggestions that you can consider. I like your use of java interfaces as it makes your code more abstract.

src/main/java/dao/TaskDaoImpl.java
simonjulianl: Yup, resolved, forgot to change tabs to spaces

src/main/java/dao/TaskDaoImpl.java
simonjulianl: Resolved

src/main/java/dao/TaskDaoImpl.java
simonjulianl: Sure, but I dont think using `if-else` is a good idea since I dont see any benefit from it, ill use more variables that represent better naming instead

simonjulianl:
> Overall great job with complying with the java coding standards. I noted a few minor suggestions that you can consider. I like your use of java interfaces as it makes your code more abstract.

Thank you so much!

src/main/java/Dude.java
simonjulianl: Yeah sure, I think that's a good idea

src/main/java/dao/TaskDao.java
simonjulianl: I think `Dao` is fine, because I think its pretty common pattern in Java, especially in Android

src/main/java/parser/ParserImpl.java
simonjulianl: Yeah sure! Resolved

PR #286, opened 2021-08-23
[Nguyen Ba Van Nhi] iP

nbvannhi:
# Bloom

> "It is better to light a candle than to curse the darkness." – Anonymous

Bloom frees your mind of having to remember things you need to do. It's,

 * text-based
 * easy to learn

All you need to do is,

1. download it from [here](https://github.com/nbvannhi/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Reminders (coming soon)

---

If you are Java programmer, you can use it to practice Java too. Here's the ```main``` method:

```java
public class Main {
    }
}
```


kevinmingtarja (commented):
LGTM! Good work overall and all the best with the remaining tasks :)

src/main/java/Bloom.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private Image user = new Image(this.getClass().getResourceAsStream("/images/user.jpg"));
+    private Image bloom = new Image(this.getClass().getResourceAsStream("/images/bloom.jpg"));
+
+    public static void main(String[] args) {
+        Application.launch(Bloom.class, args);
+    }
+
+    @Override
+    public void start(Stage stage) {
```
ErnestCuong: Perhaps you can consider breaking this method down by using helper methods to make it shorter?

src/main/java/bloom/constant/Message.java
```diff
@@ -0,0 +1,42 @@
+package bloom.constant;
+
+/**
+ * Contains messages to communicate with users.
+ */
+
+public enum Message {
```
ErnestCuong, line 6: I like how you make use of enum to group all the messages into one place.

src/main/java/bloom/task/Task.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Gets the status icon of the task.
+     *
+     * @return X if the task is done and a space otherwise
+     */
+
+    public String getStatusIcon() {
+        return (this.isDone ? "X" : " ");
```
ErnestCuong, line 61: Perhaps instead of returning "X" or " ", you can consider returning "[X]" and "[ ]".

src/main/java/bloom/task/Task.java
```diff
@@ -0,0 +1,68 @@
...
+        return "[" + this.getStatusIcon() + "] " + this.description;
+    }
+
+    /**
+     * Converts to the format used for database storage.
+     *
+     * @return a string representing the task
+     */
+
+    public String toDb() {
```
ErnestCuong, line 47: Perhaps you can make the method name more descriptive like `toDatabaseString`.

ErnestCuong (commented):
Looks good to merge. I don't see any major coding violations, perhaps you can consider shortening some of the methods. Great work.

src/main/java/Bloom.java
```diff
@@ -0,0 +1,155 @@
...
+        textToAdd.setWrapText(true);
+
+        return textToAdd;
+    }
+
+    /**
+     * Creates two dialog boxes, one echoing user input and the other containing
+     * Bloom's reply and then appends them to the dialog container.
+     * Clears the user input after processing.
+     */
```
kaixin-hc: I've noticed many of your comments have whitespace between the end of the comment and the start of the method. According to the coding standards, `Punctuation behind each parameter description -No blank line between the documentation block and the method/class` -- perhaps you could consider changing this in your next commit!

src/main/java/Launcher.java
```diff
@@ -0,0 +1,11 @@
+import javafx.application.Application;
+
+/**
+ * A launcher class to workaround classpath issues.
```
kaixin-hc, line 4: nit: work around (workaround is a noun)

kaixin-hc (commented):
Aside from these two nits, LGTM! I thought it was interesting that you used enums to separate out the messages for each command -- but also had individual classes for each command. Since each enum is only used in one place, I'm curious about why you didn't just include it as a string within each command. All the best!

PR #288, opened 2021-08-23
[Hoang Le Tri Cuong] iP

ErnestCuong:
DukePro Max Plus Ultra allows you to track your tasks with ease. Or is it? ([Vsauce](https://www.youtube.com/watch?v=IfX1hJS0iGM&t=37s))

This chat box is,
* text-based
* FAST FAST FAST
* ~~EASY~~ SUPERCALIFRAGILISTICEXPIALIDOCIOUSLY EASY to use

All you need to do is,
1. Download DukePro Max Plus Ultra [here](https://www.youtube.com/watch?v=dQw4w9WgXcQ) :wink:
2. Double-click it
3. Add your tasks
4. Magic!

Wait, did I forget to mention the best part? It's **FREE**! :partying_face:

Currently we have the following features:
- [x] Task Management
- [ ] Reminders (Coming soon!)

```java
public class Main {
    }
}
```


src/main/java/duke/FindCommand.java
```diff
@@ -0,0 +1,31 @@
...
+        this.userInput = userInput;
+    }
+
+    @Override
+    public String getResponse(TaskList tasks, Ui ui, Storage storage) throws DukeException {
+        try {
+            TaskList found = new TaskList();
+            for (int i = 0; i < tasks.numOfTasks(); i++) {
+                Task curr = tasks.getTask(i);
+                if (curr.checkDescription(userInput.substring(5))) found.addTask(curr);
```
rebeccacxy: Should conditionals be put on a separate line and wrapped by curly brackets?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+                    int end = curr.lastIndexOf(")");
+                    Task currentTask = new Deadline(
+                            curr.substring(7, by),
+                            LocalDateTime.parse(curr.substring(by + 6, end),
+                                    DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm"))
+                    );
+                    tasks.add(currentTask);
+                } else {
+                    // do nothing
+                }
```
rebeccacxy: Is this best practice to have an empty else clause? Perhaps an exception could be thrown here

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,27 @@
...
+        return description.contains(input);
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void markAsDone() {
+        isDone = true; // mark this task as done
+    }
```
rebeccacxy: Missing javadocs for this task class?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,46 @@
...
+        return tasks.get(i);
+    }
+
+    public Task removeTask(int i) {
+        return this.tasks.remove(i);
+    }
+
+    public int numOfTasks() {
+        return tasks.size();
+    }
```
rebeccacxy: Missing javadocs for this tasklist class?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,23 @@
+package duke;
+
+public class Parser {
+    public Parser() {
+
+    }
```
rebeccacxy: Is this empty constructor needed?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    Path currentRelativePath = Paths.get("");
+    String s = currentRelativePath.toAbsolutePath().toString();
+
+    public Storage() {
+
+    }
```
rebeccacxy: Is this empty constructor needed?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public ArrayList<Task> load() throws DukeException {
```
rebeccacxy, line 30: Since your save method is named `saveTasks`, would it be better to name this `loadTasks`? For standardisation

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public String showTaskDone(Task task) {
+        return "Nice! I've marked this task as done:\n  "
+                + task.toString();
+    }
+
+    public String showError(String s) {
+        return s;
+    }
```
rebeccacxy: Would it be better to name these methods `getXXX` instead of `showXXX`?

src/main/java/MainWindow.java
```diff
@@ -0,0 +1,54 @@
...
+    private Duke duke;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/images/DaUser.png"));
+    private Image dukeImage = new Image(this.getClass().getResourceAsStream("/images/DaDuke.png"));
+
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+
+        String greeting = "Hello! I'm Duke\nWhat can I do for you?";
```
rebeccacxy, line 33: Would it be better for this to be a global variable? And naming in caps, ie `private static final String GREET = ...` And also would it be better for this to be done in the Ui class and set a getter to it?

rebeccacxy (commented):
Overall, code quality is pretty good and I didn't find many violations. 
However, you are missing javadocs in many classes (all public methods must have them iirc) and there are some coding standard violations you can take note of. It is also better to put your classes in separate packages, ie. `storage`, `tasks`, `util` instead of all in one `duke` package. 

Good job and all the best for the rest of IP!

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+        try {
+            Command c = Parser.parse(input);
+            return c.getResponse(tasks, ui, storage);
+        } catch (DukeException e) {
+            return ui.showError(e.getMessage());
+        }
+    }
+
+
+    public void run() {
```
jayasting98: Perhaps you can consider adding a Javadoc header comment for this public method. Otherwise, maybe you can also consider making it private since, if I am not wrong, I do not think this is called anywhere else.

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
+import duke.Command;
+import duke.DukeException;
+import duke.Parser;
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
 public class Duke {
```
jayasting98, line 11: Should this have a Javadoc header comment as well since it is a public class?

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    public Duke() {
```
jayasting98, line 20: Maybe you should consider adding a Javadoc header comment for this public constructor too, or make it private since I think it is only called here if I am not mistaken.

src/main/java/Duke.java
```diff
@@ -1,10 +1,58 @@
...
+                c.getResponse(tasks, ui, storage);
+                isExit = c.isExit();
+            } catch (DukeException e) {
+                ui.showError(e.getMessage());
+            }
+        }
+        ui.closeUserInput();
+    }
+
     public static void main(String[] args) {
```
jayasting98, line 47: Maybe you should add a Javadoc header comment for this public method too, though perhaps it is not that necessary for the main method.

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,79 @@
...
+                    LocalDateTime.parse(s.substring(by + 5), DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm"))
+            );
+            tasks.addTask(curr);
+            return ui.showAddTask(curr, tasks.numOfTasks());
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeException("The description and time of a deadline cannot be empty.");
+        } catch (DateTimeParseException e) {
+            throw new DukeException("The time must be in this format: yyyy-MM-dd HH:mm");
+        }
+    }
```
jayasting98: Perhaps you should consider adding Javadoc header comments for this public class and public members as well. I noticed you were missing Javadoc header comments for other public things as well.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    Path currentRelativePath = Paths.get("");
+    String s = currentRelativePath.toAbsolutePath().toString();
```
jayasting98: Is there any reason these are non-private? Does `Storage` have subclasses that require access to these? On the other hand, even if there are, perhaps you could consider using getter methods even for only subclasses to use.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,46 @@
...
+        this.tasks.add(task);
+    }
+
+    public Task getTask(int i) {
+        return tasks.get(i);
+    }
+
+    public Task removeTask(int i) {
+        return this.tasks.remove(i);
+    }
```
jayasting98: Since there are no other similarly-named local variables as `tasks` in some of these, perhaps you could consider removing `this` from `this.tasks`.

src/main/java/duke/Storage.java
jayasting98: Since `s` is a variable with a large scope (a class instance variable), maybe consider using a longer, more descriptive name for it, or perhaps consider moving it to a smaller scope.

jayasting98 (commented):
Overall, I found your code had minimal coding style violations, except maybe missing Javadoc header comments in some places. Still, it is pretty clean overall. Keep it up!

src/main/java/duke/TaskList.java
ErnestCuong: Noted. Thank you for this suggestion.

src/main/java/duke/TaskList.java
ErnestCuong: Yes, thank you for pointing this out. I will update them as soon as possible. I was in a rush to submit so I forgot to put javadocs for some classes and methods.

src/main/java/duke/Ui.java
ErnestCuong: Yes, I initially put the methods as `void` but after changing them to return `String` I didn't change the name. Thank you for the suggestion.

src/main/java/duke/FindCommand.java
ErnestCuong: Yes, you are right. Thank you for pointing this out.

src/main/java/MainWindow.java
ErnestCuong: You are right. Your suggestion seems to be better.

src/main/java/duke/Parser.java
ErnestCuong: I was not aware that there is no need to include an empty constructor. Thank you for pointing this out.

src/main/java/duke/Storage.java
ErnestCuong: Yes, I forgot to put the access modifier `private`. And you suggestion for the naming is correct. However, these fields are not really used anywhere, so I probably should delete them. Thank you for the suggestions.

src/main/java/duke/Storage.java
ErnestCuong: Yes, you are right.

src/main/java/duke/Storage.java
ErnestCuong: You are right, I should throw an exception here.

src/main/java/duke/Task.java
ErnestCuong: Yes, thank you for pointing out.

src/main/java/Duke.java
ErnestCuong: Yes, thank you for pointing this out. I will update them as soon as possible. I was in a rush to submit so I forgot to put javadocs for some classes and methods.

PR #314, opened 2021-08-23
[Ngu Yi Yang] iP

nguyiyang:
# Lania
> Your **personal** chatbot to handle your tasks

Forget about remembering your own tasks, _Lania_ does it for you. It's
* easy to use
* fast
* friendly :blush:

Start using _Lania_ by doing the following,
1. download it from https://github.com/nguyiyang/ip/releases/tag/v0.1.
2. double-click it.
3. add and manage your tasks.
4. **enjoy**!

Features include:
- [x] Managing tasks
- [x] Finding tasks
- [ ] Graphical User Interface(coming soon)

Here's a short snippet of the Java program if you're interested! The `main` method:
```javascript
        Lania lania = new Lania("data/lania.txt");
        lania.run();
    }
}
```


KelvinSoo (commented):
The code looks impressive! I can't find any coding violations.

src/main/java/lania/Parser.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public class Parser {
+
+    /**
+     * Decides which command to execute based on the given input.
+     *
+     * @param fullCommand The input given by user.
+     * @return The type of command to be executed.
+     */
+    public Command parse(String fullCommand) throws LaniaException {
```
DonnyS57, line 30: This code is fine on its own but perhaps a switch statement can help to increase the readability of the code?

DonnyS57 (commented):
Aside from the single method that I commented, everything else looks good! All the methods are very clear on their purpose and very readable! Keep up the amazing work 👍

src/main/java/lania/Parser.java
```diff
@@ -0,0 +1,128 @@
...
+        } else if (firstCommand.equals("list")) {
+            return new ListCommand();
+        } else if (firstCommand.equals("find")) {
+            return new FindCommand(parseTaskDescription(fullCommand));
+        } else if (firstCommand.equals("done")) {
+            return new CompleteCommand(getIndex(fullCommand));
+        } else if (firstCommand.equals("delete")) {
+            return new DeleteCommand(getIndex(fullCommand));
+        } else {
+            if (firstCommand.equals("todo")) {
```
dannylty: I understand that these lines below refer to commands that create new tasks but would it be better to either keep it at the same indentation or abstract this out into another method to create tasks?

src/main/java/lania/Parser.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Checks the description of the given task,
+     * provided that the type of task is specified.
+     *
+     * @param command The command given to the Lania.
+     * @return The description of the given task.
+     * @throws LaniaEmptyDescriptionException If the task description is not given.
+     */
+    public String parseTaskDescription(String command) throws LaniaEmptyDescriptionException {
```
dannylty, line 81: Would it be good to standardize the word "Description" in the method names given that the below two methods for Deadline and Event does not contain it?

src/main/java/lania/Storage.java
```diff
@@ -0,0 +1,97 @@
...
+        if (f.createNewFile()) {
+            return new TaskList();
+        } else {
+            TaskList tasks = new TaskList();
+            Scanner s = new Scanner(f);
+            while (s.hasNext()) {
+                String next = s.nextLine();
+                String[] split = next.split("\\|", 4);
+                Task t;
+                if (split[0].equals("T")) {
```
dannylty: I didn't like the raw logic being in the 4th indentation level here. Consider packaging it into a separate method?

src/main/java/lania/task/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+     * Converts the deadline object to a string representation.
+     *
+     * @return The string representation of the deadline object.
+     */
+    public String getStringFormat() {
+        return "D|" + this.getStatusIcon() + "|" + this.description + "|" + this.by + "\n";
+    }
+
+    /**
+     * Another string representation of the deadline object.
```
dannylty: Maybe include the differences between the two convert-to-String methods. Applies for the below classes too.

dannylty (commented):
I liked how every method and class has a description/comments. Good job!

src/main/java/lania/Parser.java
nguyiyang: Thank you for pointing that out, I agree with you and will do so to improve the readability.

src/main/java/lania/Parser.java
nguyiyang: Thank you for pointing that out, I think keeping them at the same indentation would be great.

src/main/java/lania/Parser.java
nguyiyang: Thank you for pointing that out, I think that would make the methods more consistent and easily understood.

src/main/java/lania/Storage.java
nguyiyang: Thank you for the advice, abstracting the code into a method would indeed make it cleaner. Will do so!

PR #317, opened 2021-08-23
[Yan Xiaozhi] iP

david-eom:
This is a **bullet list**:
* `public`
* `protected`
* `private`

This is a **numbered list**:
1. _one_
2. _two_
3. _three_

This is a **task list**:

- [ ] CS2103T iP
- [ ] CS2106 lab

Link to [google](https://www.google.com.sg/)

```python
# this is a code block
print("hello world")
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,184 @@
...
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.TaskList;
+import duke.task.Todo;
+
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
```
kheekheekhee: Should java imports come before duke imports?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,184 @@
...
+    private Ui ui;
+
+
+    /**
+     * Constructor for Duke.
+     *
+     * @param filePath the path to an existing txt database file or the path to save the new txt database file
+     */
+    public Duke(String filePath) {
+        this.storage = new Storage(filePath);
```
kheekheekhee: Perhaps you can consider excluding unnecessary uses of `this` as a more advanced coding standard? I noticed this in other parts of your code as well.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,184 @@
...
+                case FIND:
+                    this.ui.showSearchResult(this.taskList, Parser.parseArgs(userInput));
+                    break;
+                }
+            } catch (DukeException e) {
+                this.ui.showDukeException(e);
+            }
+        }
+        this.ui.showFarewell();
+    }
```
kheekheekhee: The code is very clean and neat and everything is very well abstracted! Well done! 👍

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class TaskList {
+    private ArrayList<Task> list = new ArrayList<>();
+
+    /**
+     * Constructor for TaskList.
+     *
+     * @param tasks varargs of tasks to be added to the list
+     */
+    public TaskList(Task... tasks) {
```
kheekheekhee, line 19: I really like the use of varargs here! 😊

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,175 @@
...
+    public void closeInput() {
+        this.input.close();
+    }
+
+
+    /**
+     * Show greeting message to the user.
+     */
+    public void showGreet() {
+        String logo = " ____        _\n"
```
kheekheekhee: Should `logo` be a constant in the class' attributes instead?

kheekheekhee (commented):
LGTM! Extremely clean code! I had a hard time finding any errors at all and most of them were just nitpicking! Excellent code, keep up the good work! 👍

src/main/java/duke/exception/DukeInvalidArgumentException.java
```diff
@@ -0,0 +1,17 @@
+package duke.exception;
+
+/**
+ * Represents a duke exception when an invalid argument is provided for the command.
+ */
+public class DukeInvalidArgumentException extends DukeException {
```
sherrpass, line 6: I like that you created subtypes for each common Exception type for greater readability!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,184 @@
...
+                this.ui.showInvalidDateRange();
+                return;
+            }
+
+            Task event = new Event(splits[0], startTime, endTime);
+            this.taskList.add(event);
+            this.ui.showAdd(event, this.taskList.getLength());
+        } catch (IndexOutOfBoundsException e) {
+            throw new DukeMissingArgumentException();
+        } catch (DateTimeParseException e) {
```
sherrpass, line 127: Since it seems that you have created a subtype of DukeException for each possible common errors, perhaps we could keep this consistent by creating one for this too? Since it is used twice in this file.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,184 @@
...
+            this.ui.showInvalidDateFormat();
+        }
+    }
+
+
+    private void markAsDone(String args)
+            throws DukeNoTaskFoundException, DukeMissingArgumentException, DukeInvalidArgumentException {
+        try {
+            int taskNum = Integer.parseInt(args);
+            if (taskNum > this.taskList.getLength()) {
```
sherrpass, line 153: Should we check if taskNum is a negative integer as well?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+        switch (fields[0]) {
+        case "T":
+            taskToAdd = new Todo(fields[2]);
+            break;
+        case "E":
+            taskToAdd = new Event(fields[2],
+                    LocalDateTime.parse(fields[3], formatter),
+                    LocalDateTime.parse(fields[4], formatter));
+            break;
+        default: // (case "D")
```
sherrpass: Hmm, can we assume that the todo type in the data file is always valid? Perhaps we could explicitly still have a case for D and use default for potential erroneous data!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+        ArrayList<Task> taskList = new ArrayList<>();
+        try {
+            Scanner fileReader = new Scanner(db);
+            while (fileReader.hasNextLine()) {
+                taskList.add(readEntry(fileReader.nextLine()));
+            }
+            fileReader.close();
+        } catch (FileNotFoundException e) {
+            this.createDatabase();
+        }
```
sherrpass, line 53: Hmm, is using try-catch for logic the best way to go about this? We could use File.exists() for this instead!

sherrpass (commented):
Awesome work with your iP! I can tell you put tons of work into this project and it clearly paid off. Here's a couple of nits that you could consider to improve your product!

src/main/java/duke/Duke.java
david-eom: Yep i think so... Will be making the changes!

src/main/java/duke/Ui.java
david-eom: Yep that's absolutely correct.. Should've made it static 🤦‍♂️

src/main/java/duke/Duke.java
david-eom: Yep I completely forgot about this! Thanks for the catch!

src/main/java/duke/Storage.java
david-eom: Yep I agree with the suggestion! `File::exists` is really a more elegant way to handle missing file here.

src/main/java/duke/Storage.java
david-eom: Yep I should've throw a `DukeDatabaseException` for the default case here instead!

src/main/java/duke/task/TaskList.java
david-eom: 截圖 2022-09-26 下午3 39 36

PR #320, opened 2021-08-23
[Siddharth Srivastava] iP

Siddharth-Sid:

- Text based
- VERY HARD to use
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. Download it from [here](https://github.com/Siddharth-Sid/ip)
2. Run Gradle to build the project
3. Run Duke.main()

#### Features:

- [x] Adding Tasks
- [x] Adding Deadlines
- [x] Adding Events
- [ ] Managing Deadline (Coming Soon)
- [ ] Reminders (Coming Soon)

If you Java programmer, you can use it to practice Java too. Here's the ```main``` method:

```java
public class Main {
    }
}
```


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+    /**
+     * Method to check for the List Command and render the UI
+     *
+     * @param taskDetails The String describing the various attributes for the command
+     * @param ui The Object to render the UI
+     * @param tasks The Object to contain the List of the Tasks
+     * @throws DukeException An Exception class to be thrown if the command taskDetails is not valid
+     */
+    private void displayTaskList(String taskDetails, UI ui, TaskList tasks) throws DukeException {
```
TaoJun99: taskDetails is not a collection of items so maybe it can be renamed to taskDetail instead?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,168 @@
...
+                        ui.showLine();
+                        ui.showError("Error Occurred While Storing Data");
+                        ui.showLine();
+                    }
+                }
+                storeData.close();
+            }
+        } catch (IOException e) {
+            File deletedFile = new File("data/duke.txt");
+            boolean isDeleted = deletedFile.delete(), shouldShow = false;
```
TaoJun99: Maybe canShow is more appropriate than shouldShow for a boolean?

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * Method to check for the ToDO Command and render the UI
+     *
+     * @param taskDetails The String describing the various attributes for the command
+     * @param ui The Object to render the UI
+     * @param tasks The Object to contain the List of the Tasks
+     * @throws DukeException An Exception class to be thrown if the command taskDetails is not valid
+     */
+    private void addTodo(String taskDetails, UI ui, TaskList tasks) throws DukeException {
```
TaoJun99: Same as the previous comment, taskDetail instead of taskDetails?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import duke.task.Task;
+
+/**
+ * Class to abstract the List of the Tasks
+ */
+public class TaskList {
+
+    private final List<Task> taskList;
```
TaoJun99: Maybe tasks is more appropriate than taskList since it is a collection?

TaoJun99 (commented):
Overall, there are very few coding standard violations except for some minor ones 👍

src/main/java/duke/Command.java
Siddharth-Sid: Absolutely

src/main/java/duke/Storage.java
Siddharth-Sid: Both Seem similar so will discuss with tutors and accordingly change. Thank you for pointing it out.

src/main/java/duke/Command.java
Siddharth-Sid: Understood.

src/main/java/duke/TaskList.java
Siddharth-Sid: Sure

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,51 @@
...
+            }
+        } catch (IllegalArgumentException e) {
+            return Command.Commands.UNKNOWN;
+        }
+    }
+
+    /**
+     * Method to return the specific Command processed from the given from the Command String
+     *
+     * @param taskDescription The Command inputted as a String
```
luodan01, line 56: Minor error, 'inputted' could maybe be changed to 'input'. In addition, perhaps a better variable name would be something like ```userInput``` as I feel that ```taskDescription``` is not the most intuitive, and could be construed as the task details.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,298 @@
...
+import duke.task.ToDoTask;
+
+/**
+ * Class to abstract the type and execution of a command
+ */
+public class Command {
+    /**
+     * Enum to encapsulate the type of Commands
+     */
+    public enum Commands {
```
luodan01: Perhaps this could be renamed to ```CommandType```? Plural form doesn't seem very intuitive

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * Method to check for the Bye Command and render the UI
+     *
+     * @param taskDetails The String describing the various attributes for the command
+     * @param ui The Object to render the UI
+     * @throws DukeException An Exception class to be thrown if the command taskDetails is not valid
+     */
+    private void goodbye(String taskDetails, UI ui) throws DukeException {
```
luodan01: Perhaps this could be renamed to sound like an action?

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    private final Commands typeOfCommand;
+    private final String taskDetails;
+
+    /**
+     * Constructor for the Class Command
+     *
+     * @param type The type of Command
+     * @param taskDetails The String describing the various attributes for the command
```
luodan01, line 36: Perhaps it is better to add a full stop behind each parameter description?

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,298 @@
...
+    public void execute(UI ui, TaskList tasks, Storage storage) throws DukeException {
+        switch (typeOfCommand) {
+        case BYE:
+            goodbye(taskDetails, ui);
+            storage.saveTaskList(tasks);
+            break;
+        case DEADLINE:
+            addDeadline(taskDetails, ui, tasks);
+            break;
+
```
luodan01: Perhaps it's good to remove this line break?

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,298 @@
...
+     * @param storage The Object to save the List of the Tasks
+     * @throws DukeException An Exception class to be thrown if the command taskDetails is not valid
+     */
+    public void execute(UI ui, TaskList tasks, Storage storage) throws DukeException {
+        switch (typeOfCommand) {
+        case BYE:
+            goodbye(taskDetails, ui);
+            storage.saveTaskList(tasks);
+            break;
+        case DEADLINE:
```
luodan01, line 310: Perhaps it would be good to name this ```ADD_DEADLINE```? I feel that ```DEADLINE``` is a bit confusing (same for ```EVENT``` and ```TODO```) since it doesn't sound like a command

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,88 @@
+package duke;
+import java.util.Scanner;
+
+/**
+ * Class to abstract the rendering of the UI of the Duke
+ */
+public class UI {
```
luodan01, line 6: Perhaps it would be better to use pascal case and name the class ```Ui```?

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final String ANSI_RESET = "\u001B[0m";
+    public static final String ANSI_CYAN = "\u001B[36m";
+    public static final String ANSI_RED = "\u001B[31m";
+    private final Scanner input;
```
luodan01: Perhaps this could be renamed to something like ```scanner``` or ```sc```?

src/main/java/duke/task/EventTask.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public String writeToFile() {
+        return String.format("EVENT | %s | %s | %s\n", getIsCompleted(), getDescription(), getDetails());
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + details.format(DateTimeFormatter.ofPattern("MMM dd yyyy, HH:mm"))
+                + ")";
```
luodan01: Perhaps it would be better if ```String.format``` was used here?

luodan01 (commented):
LGTM! I like your comprehensive tests and how your code is organised. 🙂

src/main/java/duke/Command.java
Siddharth-Sid: Sure

src/main/java/duke/Command.java
Siddharth-Sid: Ah ok

PR #336, opened 2021-08-24
[Ang Yang Cheng] iP

djiangel:
✨ **DukePro** ✨

- Text-based
- Easy to learn
- Fast to use

1. Add tasks
2. Delete tasks
3. Update tasks as done

Upcoming

- [ ] Managing deadlines
- [ ] Reminders

> _Free_ and _amazing_ product, download [here](https://github.com/djiangel/ip)

Some sample JAVA code for beginners, `main` method:
```
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/duke/Parser.java [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+public class Parser {
+
+    /**
+     * Makes sense of user input command
+     *
+     * @param command
+     * @return String list to make sense of user input
+     * @throws DukeException
+     */
+    public static String[] parse(String command) throws DukeException {
```
chingh20, line 12: This method is quite long. Maybe you can consider extracting some lines out as different methods and call the methods, or declare the common used variable, such as fullCommand at the beginning of the method and return it at the end of the method :)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Loads task list from file
+     *
+     * @return List of task in file
+     * @throws IOException
+     */
+    public ArrayList<Task> load() throws IOException {
```
chingh20, line 27: This method is also a bit long. Maybe you can extract some lines out into different methods, such as a method for checking the type of task to add or whether a task is done or not.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        ArrayList<Task> tasks = new ArrayList<Task>(100);
+        try {
+            File data = new File(filePath);
+            Scanner dataReader = new Scanner(data);
+            while (dataReader.hasNextLine()) {
+                String task = dataReader.nextLine();
+                System.out.println(task);
+                String[] split = task.split(" , ");
+                String type = split[0];
+                boolean done;
```
chingh20: Maybe naming this boolean as isDone or hasCompletedTask might be better? (since done may be read as a verb)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,92 @@
...
+        } catch (IOException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Method to run entire programme flow
+     */
+    public void run() {
```
chingh20: This method is quite long. Maybe you can extract actions for each command as separate methods :)

chingh20 (commented):
I think your coding quality is quite good:) There's not a lot to comment about variable or methods names. Each line of your code is also quite short and clear too. Perhaps you can make several methods shorter to make it even better. Overall well done!!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,104 @@
...
+    public TaskList() {
+        this.tasks = new ArrayList<Task>(100);
+    }
+
+    /**
+     * Gives the number of tasks in the list
+     *
+     * @return the number of tasks in the list
+     */
+    public int size() {
```
sharmainec: Maybe we can use getSize() instead of size(), it will be easier to read this way.

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,24 @@
...
+public class Todo extends Task {
+
+    /**
+     * Todo constructor
+     *
+     * @param description
+     * @param done
+     */
+    public Todo(String description, boolean done) {
+        super(description, done);
```
sharmainec: I think maybe renaming the boolean to isDone, instead of done, will make it more obvious that done is a boolean.

sharmainec (commented):
LGTM. Codes are easy to read! Maybe just some minor changes to make it even more readable?

PR #339, opened 2021-08-24
[Leeroy Liu] iP

leeroy999:
# AilurusPro

>Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

AilurusPro frees your mind of having to remember things you need to do. It's,

* text-based
* easy to learn
* ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from here.
2. double-click it. 
3. add your tasks. 

And it is **FREE**!

_Features:_

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```java
public class Main {
    }
}
```


src/main/java/ailurus/Ailurus.java
```diff
@@ -0,0 +1,118 @@
...
+                    break;
+                default:
+                    this.ui.sayInvalidCommand();
+                }
+            } catch (AilurusException e) {
+                ui.showError(e.getMessage());
+            }
+        }
+    }
+
```
qinguorui2001, line 72: Hi, i think you can put the switch case  to the parser class (parse function) instead. Because that way can divide the functionality of classes, and makes run() shorter.

src/main/java/ailurus/AilurusException.java
```diff
@@ -0,0 +1,59 @@
...
+    private Error error;
+
+    /**
+     * List of errors for Ailurus.Ailurus
+     */
+    public enum Error {
+        EMPTYLIST, EMPTYTODO, EMPTYEVENT, EMPTYDEADLINE,
+        EMPTYBY, EMPTYAT, BY, AT, NUMBER, NAN, LOAD, FIND,
+    }
+
```
qinguorui2001, line 13: Define error enum outside of this class could be better？

src/main/java/ailurus/Parser.java
```diff
@@ -0,0 +1,95 @@
...
+     * Parse data stored in hard disk by line
+     *
+     * @param data line of data to be parsed
+     * @return task that is parsed from line
+     */
+    public static Task parseData(String data) {
+        String[] dataArr = data.split("[|]");
+        switch (dataArr[0]) {
+            case "T":
+                if (dataArr.length < 3) {
```
qinguorui2001: perhaps use a variable "lens" to store the dataArr's length to avoid recalculating?

src/main/java/ailurus/Storage.java
```diff
@@ -0,0 +1,93 @@
...
+                    list.add(task);
+                }
+            }
+            this.writer = new FileWriter(this.directoryName.concat(this.filename));
+        } catch (FileNotFoundException e) {
+            throw new AilurusException(AilurusException.Error.LOAD);
+        } catch (IOException e) {
+            throw new AilurusException(AilurusException.Error.LOAD);
+        }
+        return list;
```
qinguorui2001, line 69: use (FileNotFoundException e |  IOException e) could be better?

src/main/java/ailurus/task/TaskList.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Get a particular task from the task list
+     *
+     * @param i index of the task
+     * @return the particular task
+     */
+    public Task getTask(int i) {
+        return list.get(i);
+    }
+
```
qinguorui2001, line 55: I think it violets tell don't ask principle, may be you can find what should be done to tasks, like compare task times, compare two tasks content, they can be done using a method defined in tasklist class.

src/test/java/AilurusTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        String task1String = "[T][ ]Go grocery shopping";
+        String task2String = "[D][ ]CS2103T ip week 2 (by: Aug 19 2021)";
+        String task3String = "[E][ ]CS2103T lecture (at: Aug 27 2021)";
+
+        assertEquals(task1.toString(), task1String);
+        assertEquals(task2.toString(), task2String);
+        assertEquals(task3.toString(), task3String);
+    }
+
```
qinguorui2001, line 45: i like your tests, it could be better if you divide tests methods into different test packages.

qinguorui2001 (commented):
Generally, well done! just need to be more careful with the coding standards and functionality of each class.

src/main/java/ailurus/Ailurus.java
greyguy21: Hi, perhaps you can just use one try-catch block to catch the possible errors since they are all the same kind of exception.

greyguy21: Hi, perhaps you can just make use of one try-catch block to deal with the possible errors thrown since they are all the same type of exception.

src/main/java/ailurus/AilurusException.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Overriden method to get error message
+     *
+     * @return error message string
+     */
+    @Override
+    public String getMessage() {
+        switch (this.error) {
+            case EMPTYLIST:
```
greyguy21, line 31: For this, I think there should not be indentation for the case clauses.

src/main/java/ailurus/Parser.java
```diff
@@ -0,0 +1,95 @@
+package ailurus;
+
+import ailurus.task.*;
```
greyguy21, line 3: I think it would be better to import the classes in the task package explicitly for better documentation.

src/main/java/ailurus/Ui.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Welcomes user to use the chatbot
+     */
+    public void showWelcome() {
+        this.say(String.format("Hello! I'm %s. What can I do for you?", this.CHATBOT));
+    }
+
+    /**
+     * Read user input
```
greyguy21, line 21: It would be better to use singular form for the verbs here.

src/main/java/ailurus/Ailurus.java
greyguy21: Hi, perhaps you can just make use of one try-catch block to deal with the possible errors thrown since they are all the same type of exception.

src/main/java/ailurus/AilurusException.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Overriden method to get error message
+     *
+     * @return error message string
+     */
+    @Override
+    public String getMessage() {
+        switch (this.error) {
```
greyguy21, line 30: For switch statements, I think there should not be indentation for the case clauses.

src/main/java/ailurus/Parser.java
greyguy21: I think it would be better to import the classes in the task package explicitly for better documentation.

src/main/java/ailurus/Ui.java
greyguy21: It would be better to use singular form for the verbs here.

greyguy21 (commented):
Overall, I think except for some minor coding violations, your code is pretty clean and readable. I like the name you gave the chatbot too 😄

src/main/java/ailurus/Ailurus.java
leeroy999: Thank you for your suggestions! The reason why I placed them in the run function, was because the the ui, storage and tasks are all in the main Ailurus class, and it may not make sense to pass them into a parse function.

Yes thank you! Will remove the try catch blocks.

src/main/java/ailurus/Storage.java
leeroy999: Tried that and this is the prompt: `Types in multi-catch must be disjoint: 'java.io.FileNotFoundException' is a subclass of 'java.io.IOException.` So I believe I can just catch for IOException. Thank you!

PR #340, opened 2021-08-24
[Dasco Gabriel] iP

DystoriaX:
# Duke
Your personal task manager. :exploding_head:

- Text-based
- _Intuitive_
- Suitable for people who ~is lazy~ **type fast**

Get started with 4 steps:
1. Download the program [here](https://github.com/DystoriaX/ip/releases/download/A-Jar/main.jar)
2. Open your CLI and navigate to the folder containing the `.jar` file
3. Run `java -jar main.jar [path_to_file]`, where `path_to_file` is the path of the file that you want to store your task into.

That's all!

Features:
- [x] Manage task
- [x] Find task
- [ ] Reminder

> To be or not to be
> ~ you-know-who

```java
public class Main{
        System.out.println("Hello world\n");
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
...
+import duke.command.Command;
+import duke.exception.DukeException;
+import duke.parser.Parser;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+public class Duke {
+    private Ui ui;
+    private TaskList taskList;
```
benedictchuajj, line 21: Small nit but perhaps use a plural form name such as *tasks* to represent a collection of objects?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.format.DateTimeParseException;
+
+import duke.command.Command;
+import duke.exception.DukeException;
+import duke.parser.Parser;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+public class Duke {
```
benedictchuajj, line 19: Perhaps add a descriptive header comment for Duke and its methods since they are public?
> Write descriptive header comments for all public classes/methods. - [CS2103T Java coding standard](https://se-education.org/guides/conventions/java/intermediate.html#comments)

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,31 @@
...
+    public AddCommand(Task task) {
+        this.task = task;
+    }
+
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws IOException {
+        taskList.addTask(this.task);
+        ui.sayAddTask(this.task, taskList.size());
+        storage.store(taskList.serialize());
+    }
```
benedictchuajj, line 28: Since there's no Javadoc from the inherited class for execute, perhaps you can add a Javadoc here explaining what execute will do. I noticed the same issue in a few other classes that inherit from Command.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+        } else if (input.startsWith("find ")) {
+            String keyword = input.substring(5);
+            return new FindCommand(keyword);
+        } else if (input.startsWith("done ")) {
+            int index = Integer.parseInt(input.substring(5));
+            return new MarkCommand(index);
+        } else if (input.startsWith("delete ")) {
+            int index = Integer.parseInt(input.substring(7));
+            return new DeleteCommand(index);
+        } else if (input.startsWith("todo ")) {
```
benedictchuajj: Perhaps you can use switch statements here to improve readability for this method. 😃

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+
+import duke.exception.DukeException;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/** A class that represents a deadline task */
+public class Deadline extends Task {
+    private LocalDate by;
```
benedictchuajj, line 11: Small nit here, but perhaps rename *by* to *byDate* for greater clarity that it is a LocalDate object?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Serializes the task as a string representation.
+     *
+     * @return Serialized string representation.
+     */
+    public String serialize() {
+        return this.description;
+    }
```
benedictchuajj: Any reason why this is a separate method from getDescription instead of just reusing getDescription? If they serve the same intended purpose, perhaps you can remove one of them?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,107 @@
...
+                // Deadline duke.task
+                String by = params[3];
+
+                return new Deadline(description, by, done);
+            } else if (taskType.equals("E")) {
+                // Event duke.task
+                String at = params[3];
+
+                return new Event(description, at, done);
+            }
```
benedictchuajj: Perhaps you can consider using switch statements here too.

benedictchuajj (commented):
Looks good with very few coding standard violations. 👍
Just a few nits to fix to improve readability.

Keep up the good work for iP 😃

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
...
+            try {
+                response = ui.getUserCommand();
+                Command c = Parser.parse(response);
+                c.execute(taskList, ui, storage);
+                isExit = c.isExit();
+            } catch (IOException e) {
+                ui.sayError(e.getMessage());
+            } catch (DukeException e) {
+                ui.sayError(e.getMessage());
+            }
```
ailing35: Perhaps these two catch statements can be merged into one using ` ||` since they have the same body?

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,31 @@
+package duke.command;
+
+import duke.storage.Storage;
+
+import duke.task.Task;
+import duke.task.TaskList;
+
+import duke.ui.Ui;
```
ailing35: Perhaps it's better to group these together for better consistency across different files?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Stores data to the file.
+     *
+     * @param data The data to be stored.
+     * @throws IOException If an I/O error occurs.
+     */
+    public void store(String data) throws IOException {
+        FileWriter fout = new FileWriter(filePath.toString());
+        BufferedWriter bout = new BufferedWriter(fout);
```
ailing35, line 72: Could the namings be better? One suggestion is using `bufferedOutput` instead of `bout`, or since the scope of use for the variable is small, it could be `b`.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+            + "  %1$s\n"
+            + "You have %2$d task(s) in the list";
+    public static final String MESSAGE_LIST_TASKS = "Your current task(s):\n"
+            + "%1$s";
+    public static final String MESSAGE_FOUND_TASKS = "Found the following task(s):\n"
+            + "%1$s";
+    public static final String MESSAGE_MARK_TASK = "Marked task %1$d as done";
+    public static final String MESSAGE_LOADING_SUCCESS = "Data loaded";
+    public static final String MESSAGE_LOADING_FAIL = "Failed to load data";
+    public static final String MESSAGE_GOODBYE = "Good bye!";
```
ailing35: Any reason why these are all `public` instead of `private`?

ailing35 (commented):
Overall great use of OOP, perhaps can add more javadoc comments for consistency.

src/main/java/duke/Duke.java
DystoriaX: Ah yes. Thanks for the catch!

src/main/java/duke/task/Task.java
DystoriaX: `getDescription()` is intended to access the `description` property from outside. I think implementation wise, I'd still prefer to directly call `this.description` as there is no abstraction barrier that we break. I guess this also helps improve the program performance by a bit.

src/main/java/duke/Duke.java
DystoriaX: That's a good idea. Thanks!

src/main/java/duke/ui/Ui.java
DystoriaX: Since it's constant, I guess it may be a good idea if other classes can access it. Outside user won't be able to modify it either, so this is to improve the flexibility. Will consider to change it to `private` if it is necessary. Thanks!

PR #341, opened 2021-08-24
[Dion Neo] iP

noobmaster19:
hi


lib/src/main/java/duke/utils/Storage.java
```diff
@@ -0,0 +1,130 @@
+package duke.utils;
+
+import java.util.*;
```
tsm1820: Overall everything looks good, but perhaps can use specific import statements instead of using wildcard imports?

lib/src/main/java/duke/utils/Storage.java
```diff
@@ -0,0 +1,130 @@
...
+
+import duke.tasks.*;
+
+import java.io.*;
```
tsm1820: Same point regarding wildcard imports as above :)

tsm1820 (commented):
LGTM. Just the bits regarding the import statements.

lib/src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,62 @@
+package duke;
+
+import java.util.*;
```
Neha-5678: Maybe you could consider explicitly stating the imported classes? :)

lib/src/main/java/duke/utils/Storage.java
```diff
@@ -0,0 +1,130 @@
...
+     * @throws DukeException If database file is not found.
+     */
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> savedTasks = new ArrayList<>();
+        try {
+            Scanner s = new Scanner(data);
+            while (s.hasNext()) {
+                String[] parsed = s.nextLine().split(",");
+                String type = parsed[0];
+                boolean done = (parsed[1].equals("1"));
```
Neha-5678: Maybe it would be clearer if you changed the name of the boolean variable to isDone? Just a suggestion!

lib/src/main/java/duke/utils/Storage.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return An ArrayList containing the saved tasks
+     * @throws DukeException If database file is not found.
+     */
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> savedTasks = new ArrayList<>();
+        try {
+            Scanner s = new Scanner(data);
+            while (s.hasNext()) {
+                String[] parsed = s.nextLine().split(",");
+                String type = parsed[0];
```
Neha-5678: This is a very minute detail but maybe you could consider renaming the variable to make it clearer? for instance, typeOfTask?

lib/src/main/java/duke/utils/Storage.java
```diff
@@ -0,0 +1,130 @@
...
+                boolean done = (parsed[1].equals("1"));
+                if (type.equals("todo")) {
+                    savedTasks.add(new Todo(parsed[2], done));
```
Neha-5678: I was thinking that maybe it would be better if you had separate variables declared for parsed[2] and parsed[3] so that it is clearer what you are passing into the constructor methods? Again, this is just a suggestion! :)

lib/src/main/java/duke/utils/TaskList.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param input  the index of task of interest
+     */
+    public void deleteTask(String input) {
+        try {
+            System.out.println("Noted. I've removed this task: ");
+            int taskIndex = Integer.parseInt(input);
+            Task task = userInputs.get(taskIndex - 1);
+            userInputs.remove(taskIndex - 1);
+            System.out.println(" " + task);
+            task = null;
```
Neha-5678: Very trivial but I was just thinking if there is a need to set the task variable back to null since you won't be using it elsewhere?

Neha-5678 (commented):
good job! Overall, a well-written code and I could only spot the minute details.

PR #347, opened 2021-08-24
[Alyssa Savier] iP

alyssa-savier:
Joker

> “What doesn't kill you simply makes you STRANGER.” – Joker ([source](https://www.youtube.com/watch?v=o5o-w0F7kfQ))

Joker frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/nus-cs2103-AY2122S1/ip/pull/368).
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,27 @@
...
+        TaskManager taskManager = new TaskManager();
+
+        Scanner scan = new Scanner(System.in);
+        while (!Objects.equals(inData, "bye")) {
+            inData = scan.nextLine();
+            taskManager.executeCommand(inData);
+        }
+    }
+
+    public static void introMessage() {
```
wpinrui: Would it be clearer if this was named `printIntroMessage`?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
+import java.io.*;
```
wpinrui: Would it be possible to list all imports?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+import java.time.LocalDate;
+import java.util.Locale;
+import java.util.Objects;
+
+public class TaskManager {
+    Task[] taskList = new Task[100];
+    int i = 0;
```
wpinrui: Wonder if a more descriptive name would be helpful here?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+    File txtFile;
```
wpinrui: Would it be better to avoid abbreviations and name this as `textFile`?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+                txtFile.createNewFile();
+                this.txtFile = txtFile;
+            } catch (IOException e) {
+                System.out.println("An error occurred.");
+                e.printStackTrace();
+            }
+        } else {
+            try {
+                BufferedReader br = new BufferedReader(new FileReader(txtFile));
+                String txtLine;
```
wpinrui: Similarly, should this be named as `textLine`?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+                        taskList[i] = new Event(description, date);
+                        eventAddedMessage();
+                        i++;
+                    } catch (Exception e) {
+                        System.out.println("Please enter date in this format: yyyy-mm-dd (e.g., 2019-10-15)");
+                    }
+                }
+            } else if (inData.contains("done ")) {
+                if (isNumeric(inData.substring(5, inDataLength))) {
+                    int taskNo = Integer.parseInt(inData.substring(5, inDataLength));
```
wpinrui: Would it be better to call this `taskNumber` to prevent confusion with the negative reply 'No'?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+            } else {
+                System.out.println(j + 1 + ". [" + taskList[j].getTask() + "]" +
+                        "[" + taskList[j].getStatusIcon() + "] " + taskList[j].getDescription()
+                        + taskList[j].getDate());
+            }
+        }
+        System.out.println("____________________________________________________________");
+    }
+
+    public void toDoAddedMessage() {
```
wpinrui: Would it be better to use an action to name this, like `printToDoAddedMessage` (similar to the other printing methods too)?

wpinrui (commented):
I think the naming is generally pretty good, save for a few tips that we have just learnt about naming methods with actions. Coding standard looks good too.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
```
Anonymxtrix, line 4: Should there be header comments following the coding standards layed out [here](https://se-education.org/guides/conventions/java/intermediate.html)? I noticed the same issue in several other places too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,27 @@
...
+ */
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String inData = "yo";
```
Anonymxtrix: Could you perhaps explain to me what this is supposed to be? I don't quite understand this part. 

EDIT: Ahh... I get it now... But would it be more accurate to change it to a final variable `START_COMMAND` instead? Since it is passed into a method `executeCommand`.

src/main/java/Event.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public Event(String description, String date) {
+        super(description);
+        this.dateNum = date;
+        this.date = LocalDate.parse(date);
+    }
+
+    @Override
+    public String getTask() {
+        return "E";
```
Anonymxtrix, line 29: Any reason why you did it this way instead of using a constant or final variable? Same for a few other places...

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public String getDescription() {
+        return this.description;
+    }
+
+    public String getTask() {
```
Anonymxtrix, line 46: Would it be more accurate to have named this `getTaskTypeIcon` following the style from `getStatusIcon` in line 12?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+                }
+            } catch (IOException e) {
+                System.out.println("ERROR OCCURRED");
+            }
+        }
+    }
+
+    public void executeCommand(String inData) {
+        int inDataLength = inData.length();
+        if (!Objects.equals(inData.toLowerCase(), "bye")) {
```
Anonymxtrix: Perhaps `!inData.toLowerCase().equals("bye")` would read more easily here?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+            if (Objects.equals(inData, "list")) {
+                listAll();
+            } else if (inData.contains("todo")) {
```
Anonymxtrix: Nice! I like the way `inData.contains("todo")` reads as "In data contains todo" in english! Very comfortable to read 👍

Anonymxtrix (commented):
Nicely done! I really am able to see the amount of effort you put in to follow the coding standards [here](https://se-education.org/guides/guidelines/PRs-reviewing.html).

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+                    }
+                    txtLine = br.readLine();
+                }
+            } catch (IOException e) {
+                System.out.println("ERROR OCCURRED");
+            }
+        }
+    }
+
+    public void executeCommand(String inData) {
```
itsraveen: I think a better way to implement this would be to extract all the methods for the different if else blocks instead of trying to handle them all in this executeCommand method. That'll help make the code easier to read and understand, while keeping each method simple and concise.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
...
+                        + "Got it. I've added this task: \n"
+                        + "[" + taskList[i].getTask() + "]"
+                        + "[" + taskList[i].getStatusIcon() + "] "
+                        + taskList[i].getDescription()
+                        + "\n"
+                        + "Now you have " + (i + 1) + " tasks in the list.\n"
+                        + "____________________________________________________________");
+    }
+
+    public void deadlineAddedMessage() {
```
itsraveen: I like how each of these are stored as methods, makes it easier to read in the code above. 👍

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,289 @@
+import java.io.*;
+import java.time.LocalDate;
+import java.util.Locale;
+import java.util.Objects;
+
+public class TaskManager {
+    Task[] taskList = new Task[100];
```
itsraveen: Having an array size of 100 is fine here, but maybe can consider an ArrayList instead.

itsraveen (commented):
Code looks good with no major issues. Hope my suggestions are useful.

PR #348, opened 2021-08-24
[Lim Wan Ning] iP

wanninglim:
# Ponyo

Ponyo frees your mind of having to remember things you need to do. It's,

* text-based
* easy to learn

All you need to do is,

1. download it from [here](https://github.com/wanninglim/ip/releases/tag/v0.1).
1. double-click it.
1. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```
public class Main {
    }
}
```


src/main/java/ponyo/commands/AddCommand.java
```diff
@@ -0,0 +1,99 @@
...
+import ponyo.data.task.Task;
+import ponyo.data.task.TaskList;
+import ponyo.data.task.Todo;
+import ponyo.storage.Storage;
+import ponyo.ui.Ui;
+
+/**
+ * Adds a new task to the task list.
+ */
+public class AddCommand extends Command {
```
charltonator, line 20: Would it be better if the class was named CommandAdder? I think AddCommand might sound like a method name :)

src/main/java/ponyo/commands/AddCommand.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Formats the date from yyyy-mm-dd to MM d yyyy
+     *
+     * @param inputDate The original date in the form of yyyy-mm-dd
+     * @param slashIndex The index to split the command to get the date
+     * @return The string representation of the date in MMM d yyyy
+     * @throws DateTimeParseException Thrown when the inputDate is of wrong format.
+     */
+    public String dateFormatter(String inputDate, int slashIndex) throws DateTimeParseException {
```
charltonator: Would it be better to name this method formatDate so it follows coding standards better? 💯

src/main/java/ponyo/commands/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+
+import ponyo.data.task.Task;
+import ponyo.data.task.TaskList;
+import ponyo.storage.Storage;
+import ponyo.ui.Ui;
+
+/**
+ * Deletes a task identified using its index in the task list.
+ */
+public class DeleteCommand extends Command {
```
charltonator, line 12: Same as AddCommand maybe the class name could be CommandDeleter

src/main/java/ponyo/commands/ListCommand.java
```diff
@@ -0,0 +1,22 @@
+package ponyo.commands;
+
+import ponyo.data.task.TaskList;
+import ponyo.storage.Storage;
+import ponyo.ui.Ui;
+
+/**
+ * Displays the list of stored tasks.
+ */
+public class ListCommand extends Command {
```
charltonator, line 11: Same as the other classes that inherit from command maybe you could change the naming?

src/main/java/ponyo/data/task/TaskList.java
```diff
@@ -0,0 +1,56 @@
...
+    public TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * Adds a task to the task list.
+     *
+     * @param task the task to be added to the list
+     */
+    public void add(Task task) {
```
charltonator, line 25: Maybe better if the name was clearer such as addTask?

src/main/java/ponyo/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    /**
+     * Checks if the file and folder exists
+     *
+     * @param filePath path to check
+     * @throws IOException if file creation fails
+     */
+    public void fileFolderCheck(String filePath) throws IOException {
```
charltonator: maybe better if the method was named checkFileFolder?

src/main/java/ponyo/commands/AddCommand.java
wanninglim: That's true! Thanks for the feedback! :-)

src/main/java/ponyo/storage/Storage.java
wanninglim: Thanks for spotting that!

src/main/java/ponyo/Main.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class Main extends Application {
+
+    private Ponyo ponyo = new Ponyo();
+
+    @Override
+    public void start(Stage stage) {
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
+            AnchorPane ap = fxmlLoader.load();
```
Anonymxtrix, line 23: Perhaps we can consider using `anchorPane` here instead of `ap`. I get that since the number of lines in this function is short, the meaning of `ap` is pretty obvious, however once the number of lines here get too long or the `Scene scene = new Scene(ap);` line moves too far away, the meaning of `ap` might not be as obvious anymore.

src/main/java/ponyo/Ponyo.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Generates response to user input.
+     *
+     * @param input Input provided by user.
+     * @return Response message by system.
+     */
+    public String[] handleInput(String input) {
+        boolean isExit = false;
```
Anonymxtrix: Maybe `isExiting` or `isExited` since I can't really understand what this `isExit` is used for O.o

src/main/java/ponyo/commands/AddCommand.java
Anonymxtrix: No actually I think that `CommandAdder` will distort the meaning of the class. This class is meant to represent the command to add a task into the task list. As such, either `Add` or `AddCommand` should be fine. Any thoughts?

src/main/java/ponyo/commands/AddCommand.java
```diff
@@ -0,0 +1,95 @@
...
+import ponyo.data.task.TaskList;
+import ponyo.data.task.Todo;
+import ponyo.storage.Storage;
+import ponyo.ui.Ui;
+
+/**
+ * Adds a new task to the task list.
+ */
+public class AddCommand extends Command {
+    private final String[] taskToAdd;
```
Anonymxtrix, line 21: Since this variable is of type `String[]`, I think it would be nicer to have a plural name like `tasksToAdd` instead. There are a few other similar instances in your pull request (eg. Line 21).

src/main/java/ponyo/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        read.close();
+    }
+
+    /**
+     * Formats the text to be written in the file
+     *
+     * @param t task to be written into file
+     */
+    public void fileLineToWrite(Task t) {
```
Anonymxtrix: Since method names should make use of verbs, I suggest using `writeTaskToFile` or `appendTaskToFile` or `addTaskToFile` instead. This is because of two reasons:
- Method names should make use of verbs. 
- Method name tells me you are dealing with "line" when the parameter is actually a `Task`. 

Alternatively, `addToFile` or `writeToFile` etc. may also be good enough since the parameter already implies that we are only working with a `Task`.

src/main/java/ponyo/commands/AddCommand.java
```diff
@@ -0,0 +1,95 @@
...
+                Task t = new Todo(taskToAdd[1]);
+                tasks.add(t);
+                storage.fileLineToWrite(t);
+                return showTask(t, tasks.size());
+            } catch (ArrayIndexOutOfBoundsException e) {
+                return Ui.show("☹ OOPS!!! The description of a todo cannot be empty.");
+            }
+        case "deadline":
+            try {
+                int slashIndex = taskToAdd[1].indexOf("/");
```
Anonymxtrix: How about making `"/"` into a final constant since its used in a few places in this method? What are your thoughts?

src/main/java/ponyo/commands/Command.java
```diff
@@ -0,0 +1,23 @@
...
+     * @param storage Load tasks and save tasks to file
+     * @return a String array of responses to communicate with user
+     */
+    public abstract String[] execute(TaskList tasks, Storage storage);
+
+    /**
+     * @return true if the command is meant to exit the program,
+     *         false otherwise.
+     */
+    public abstract boolean isExit();
```
Anonymxtrix: I think I know what your `isExit` is trying to do now. Perhaps `isExitingCommand` might be more descriptive about the use of this method?

src/main/java/ponyo/commands/DeleteCommand.java
Anonymxtrix: Refer to the original comment on AddCommand for my reply

src/main/java/ponyo/commands/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+    private final int taskToDelete;
```
Anonymxtrix: Perhaps `taskIdToDelete` would be better here since we are working with an `int` instead of a `Task` type. Similarly for other `Command` type class fields which are `int`s like the one in `DoneCommand` etc...

src/main/java/ponyo/commands/ExitCommand.java
```diff
@@ -0,0 +1,21 @@
+package ponyo.commands;
+
+import ponyo.data.task.TaskList;
+import ponyo.storage.Storage;
+
+/**
+ * Exits the program.
+ */
+public class ExitCommand extends Command {
+    private static final String MESSAGE_BYE = "\tBye. Hope to see you again soon!";
```
Anonymxtrix: This is probably just me being picky but maybe we can try `BYE_MESSAGE` since it reads better than `MESSAGE_BYE`.

src/main/java/ponyo/commands/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    public String[] execute(TaskList tasks, Storage storage) {
+        ArrayList<Task> taskClone = new ArrayList<>();
+
+        for (int i = 0; i < tasks.size(); i++) {
+            if (tasks.retrieveTask(i).containsKeyword(keywordToFind)) {
+                taskClone.add(tasks.retrieveTask(i));
+            }
+        }
+
+        if (taskClone.size() != 0) {
```
Anonymxtrix: In cases like this, I love to use `taskClone.isEmpty()` instead since it reads better than `taskClone.size() != 0` although both works nicely.

src/main/java/ponyo/commands/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+public class FindCommand extends Command {
+    private final String keywordToFind;
+
+    public FindCommand(String keywordToFind) {
+        this.keywordToFind = keywordToFind;
+    }
+
+    @Override
+    public String[] execute(TaskList tasks, Storage storage) {
+        ArrayList<Task> taskClone = new ArrayList<>();
```
Anonymxtrix: Perhaps `taskListClone` or `tasksClone`? Although I think something like `matchingTasks` would do better in terms of explaining what this represents.

src/main/java/ponyo/commands/ListCommand.java
Anonymxtrix: Same reply as the above similar comments :)

src/main/java/ponyo/data/task/Deadline.java
```diff
@@ -0,0 +1,30 @@
...
+
+    // File operations
+    @Override
+    public String toStringInFile() {
+        return "D - " + super.toStringInFile() + " - " + by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
Anonymxtrix: Nice! Although perhaps we can consider putting `"D"` here in a final constant or variable since its used both in your `toStringInFile` and `toString` methods. Similarly for the rest of the `Task` related classes.

src/main/java/ponyo/data/task/Task.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Converts the Task into a string.
+     *
+     * @return A string representation of a Task object.
+     */
+    public String toString() {
+        return "[" + this.getStatusIcon() + "] " + this.description;
+    }
+
+    public boolean containsKeyword(String kw) {
```
Anonymxtrix: I think not really necessary but I would suggest using `keyword` instead for the same reason as `ap` above.

src/main/java/ponyo/data/task/TaskList.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+/**
+ * Stores the list of tasks
+ */
+public class TaskList {
+    protected List<Task> tasks;
+
+    public TaskList() {
+        this.tasks = new ArrayList<>();
```
Anonymxtrix, line 13: Not naming related but you could probably make use of `super(new ArrayList<>());` here instead to link this constructor to the one defined in line 16.

src/main/java/ponyo/data/task/TaskList.java
Anonymxtrix: I think its fine since the type of the parameter when mousing over the add method would have already implied that this is dealing with adding a `Task`. What are your thoughts?

src/main/java/ponyo/data/task/TaskList.java
```diff
@@ -0,0 +1,56 @@
...
+    public void add(Task task) {
+        tasks.add(task);
+    }
+
+    /**
+     * Removes a task from the task list.
+     *
+     * @param task the task to be removed from the list.
+     */
+    public void remove(int task) {
```
Anonymxtrix: Perhaps using `taskId` here might be more indicative of the parameter type instead. Same for the `retrieveTask` method.

src/main/java/ponyo/storage/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    /**
+     * Read the tasks from the file separated with delimiter (-)
+     *
+     * @param tasks stores the list of tasks read from the file
+     * @param filePath path of file
+     * @throws IOException if error reading file
+     */
+    public void readTasksFromFile(ArrayList tasks, String filePath) throws IOException {
```
Anonymxtrix: Also not naming related but I am curious why you decided to take in the `ArrayList tasks` as a parameter instead of either returning a newly created `ArrayList` or returning the raw text from `Stoage` class and parsing it in your `Tasks` or `TaskList` class.

src/main/java/ponyo/ui/Ui.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Shows a message to user.
+     *
+     * @param msg the message to be printed
+     * @return an array of a string to be printed
+     */
+    public static String[] show(String msg) {
+        return new String[] { msg };
+    }
```
Anonymxtrix: Arent these 2 methods the same O.o? Whats the use of the `show` method in line 67?

Anonymxtrix (commented):
Nicely done! I like the way you followed all the instructions on the IP project task descriptions to abstract your program using `Command`, `Task`, `Ui`, etc... If you need to clarify any of my comments feel free to tag me (although I tend to not look at my notifs) XD

src/main/java/ponyo/Ponyo.java
wanninglim: Ah, I used `isExit` to symbolise whether a type of command is meant to exit the application! And you're right haha after looking at it now it does seem a little vague! How does `isExitCommand` sound?

src/main/java/ponyo/commands/AddCommand.java
wanninglim: Oops! Must have missed this out when I was refactoring my code haha T_T

src/main/java/ponyo/storage/Storage.java
wanninglim: Thanks for spotting this :-)

src/main/java/ponyo/commands/Command.java
wanninglim: Just saw your comment here! That sounds good too 👍

src/main/java/ponyo/commands/DeleteCommand.java
wanninglim: Sounds good to me. 👍

src/main/java/ponyo/commands/ExitCommand.java
wanninglim: I named the constants such that all messages were prefixed with `MESSAGE` :) but your suggestion definitely sounds better!

src/main/java/ponyo/commands/FindCommand.java
wanninglim: It completely slipped my mind that there's such a method in ArrayList! Thanks for your input :)

src/main/java/ponyo/ui/Ui.java
wanninglim: LOL omg you're right TT thanks!!

src/main/java/ponyo/storage/Storage.java
wanninglim: Ah you're right! Made a rookie mistake of mutating my arguments in the process. 👎  Thanks! :)

PR #351, opened 2021-08-24
[Quan Teng Foong] iP

kaldius:
### Duke

> "Random quote here" - Quoted Person (Source here)

- text-based
- easy to learn
- ~FAST~ _SUPER FAST_ to use

All you need to do is,

1. download it from [here](https://www.google.com).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 💩

And it is **FREE**!

Features:

- [ ] Managing tasks

___
```java
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,70 @@
...
+            file.createNewFile();
+        }
+        return file;
+    }
+
+    /**
+     * Makes a parent directory for the current filePath. If parent does not have a parent, this method
+     * recursively calls itself.
+     */
+    private void mkParentDir() {
```
qreoct, line 40: Name for this method could use a verb instead, such as `makeParentDir()`

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        switch (command) {
+        case "todo":
+            return new ToDo(input);
+        case "event":
+            String[] message_and_eventDate = input.split("/at ");
+            return new Event(message_and_eventDate[0], message_and_eventDate[1]);
+        case "deadline":
+            String[] message_and_endTime = input.split("/by ");
+            return new Deadline(message_and_endTime[0], message_and_endTime[1]);
```
qreoct: According to the style guide _('Variable names must be in camelCase')_, it might be clearer to name these variables as `messageAndEventDate` instead.

README.md
```diff
@@ -1,4 +1,4 @@
-# Duke project template
+# duke.Duke project template
```
qreoct, line 1: Did you do an auto-refactor here using IntelliJ? It might have changed `Duke` to `duke.Duke` when you introduced packages, and it also affects non-code files such as in this README.

qreoct (commented):
I enjoyed reading your code! I've learned a thing or two for my own program, such as having a Command pattern to handle the various commands, as well as using `toStorage()` in your Task objects, which is really clever!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        return storage;
+    }
+
+    /**
+     * Returns the number of items in the TaskList.
+     *
+     * @return size of TaskList
+     */
+    public int size() {
```
NaokiShum, line 114: Maybe the method name could be something like getSize() so that it will be more clear

src/main/java/command/DeleteTaskCommand.java
```diff
@@ -0,0 +1,23 @@
+package command;
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+import exceptions.DukeException;
+
```
NaokiShum, line 8: Missing javadocs for this file

NaokiShum (commented):
Code looks good overall with no real violation of the coding standard. You might have missed out on adding the javadocs for one of the file. Best of luck in the remaining part of the project!

PR #353, opened 2021-08-24
[Nguyen Quang Vinh] iP

IceWizard4902:

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is

1.  download it from [here](https://nus-cs2103-ay2122s1.github.io/website/schedule/week4/project.html)
2.  double-click it.
3.  add your tasks.
4.  let it manage your tasks for you 😉 

And it is **FREE**!
 
Features:
- [x] Managing tasks
- [ ] Reminders (coming soon) 

If you Java programmer, you can use it to practice Java too. Here's the `main` method
```java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,176 @@
+package duke;
+
+import tasks.ToDo;
+import tasks.Deadline;
+import tasks.Event;
+import tasks.TaskList;
```
vimuthm: Alphabetically ordered imports might increase readability!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.util.Scanner;
+
+/**
+ * Input parsing and displaying output to the user.
+ */
+public class Parser {
+    private final Scanner sc;
```
vimuthm: It might make sense to handle `Scanner` object creation and handling within the `Ui` class. Thoughts?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+public class Storage {
+    private static final String DIRECTORY = "src/data";
+    private static final String FILEPATH = "src/data/duke.txt";
+
+    /**
+     * Save the task list to a location on the hard disk
+     *
+     * @param tasks The task list to be saved.
+     */
+
```
vimuthm: There shouldn't be a line between the ending javadoc and function declaration.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Load the saved task list, if the task list does not exist, then raise an exception
+     *
+     * @return Saved task list.
+     * @throws IOException Task list does not exist.
+     */
+    public static ArrayList<Task> loadTaskListFromHardDisk() throws IOException {
+        File dir = new File(DIRECTORY);
+        File logs = new File(FILEPATH);
+        ArrayList<Task> result = new ArrayList<>();
```
vimuthm: Collections variables should be named with plural words

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,50 @@
+package duke;
+
+/**
+ * Utility to help deals with interactions with the user
+ */
+public class Ui {
+
+    public static void greetings() {
```
vimuthm: This public method would need javadocs

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,61 @@
...
+     * Constructor for the tasks.Task class
+     *
+     * @param description The description of the task.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getDescription() {
```
vimuthm: This would require javadoc

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,133 @@
...
+            Ui.displayContent("Can't read the save file.");
+        }
+        this.tasks = temp;
+    }
+
+    /**
+     * Add a task to the list
+     *
+     * @param task The added task.
+     */
```
vimuthm: This comment would require the return type.

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,133 @@
...
+            return res.toString();
+        }
+    }
+
+    /**
+     * Return the size of the task list
+     *
+     * @return Size of the task list.
+     */
+    public int size() {
```
vimuthm: This method might be better named `getSize`

src/main/java/tasks/TaskList.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public String toString() {
+        if (this.size() == 0) {
+            return "There is no task in the list";
+        } else {
+            StringBuilder res = new StringBuilder("Here are the tasks in your list:\n");
+            for (int i = 0; i < tasks.size(); i++) {
+                String temp = "\t" + " " + (i + 1) + "." + tasks.get(i).toString();
+                if (i < tasks.size() - 1) { // remove the last \n char, ugly but get the job done
```
vimuthm: Nice :) I handled this similarly!

src/test/java/duke/StorageTest.java
```diff
@@ -0,0 +1,30 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+import tasks.*;
```
vimuthm: Wildcard imports should not be used

vimuthm (commented):
Overall, it looks quite good and nice use of OOP!

One thing I would point out is that the `Parser` class is also handling the user input from the scanner which might be better suited for the `Ui` class. 

A `Command` class would also be a nice addition if time permits.

Other than that, it looks good to me!

src/main/java/Duke.java
```diff
@@ -1,10 +1,13 @@
...
+ */
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Ui.greetings();
```
Yttruire: Method names should be verbs instead of nouns, according to the coding standard that we are following. Instead, consider something like `Ui.sendGreetings()` or `Ui.greetUser()`

src/main/java/duke/Parser.java
Yttruire: I think I agree with Vimuth. We might naturally expect a `Parser` class to only parse input and sort them out/respond to input that cannot be parsed, but not necessarily scanning for the input itself. The `Ui` class on the other hand, could definitely handle both input (receiving input) and output (printing messages to user)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    private void handleFind(String[] parsedInput) throws DukeInvalidCommandException {
+        if (parsedInput.length < 2) {
+            throw new DukeInvalidCommandException("OOPS!!! Type in the keyword you want to search");
+        }
+        Ui.displayContentBetweenLines(taskList.findTask(parsedInput[1]));
+    }
+
+    private void dukeCommandController(String[] parsedInput) throws DukeInvalidCommandException {
```
Yttruire: Notice that this method is a noun, and actually, sounds like an object. Perhaps this would be better as an individual class for the sake of OOP if you have time. Else, the actions to take in response to a command you retrieve from parsing, may fit better elsewhere (Maybe in the main class if you're not encapsulating this into a different class) rather than leave this in the `Parser`

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,50 @@
...
+    public static void displayContent(String content) {
+        System.out.println("\t" + " " + content);
+    }
+
+    /**
+     * Print out the formatted version of any string content between two horizontal lines
+     *
+     * @param content Content to display.
+     */
+    public static void displayContentBetweenLines(String content) {
```
Yttruire: This method name sounds a little verbose, although there is nothing wrong with it. Perhaps try names that describe what  it does less, and more of what it is for? E.g. If this is the function to be used to display every form of output, this could simply be named `display` or `displayMessage`.

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * tasks.Task that need to be done before a specific date/time
+ */
+public class Deadline extends Task {
+
+    protected LocalDate by;
+
+    /**
+     * Constructor for the class.
```
Yttruire: The first sentence as a summary for the method should be a verb to be as descriptive as possible. According to the coding standards, it could start with `Returns...`, `Sends...`, or `Adds...` for example. Thus, I think that `Constructs` might be more appropriate here.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    /**
+     * Return string representation of the task to write to hard disk.
+     *
+     * @return The string representation.
+     */
+    public String toSaveInHardDisk() {
+        return "";
```
Yttruire: If this method is not intended to be useful as it is, perhaps consider making this and the class, an abstract method and abstract class responspectvely

src/test/java/duke/StorageTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.ArrayList;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class StorageTest {
+    @Test
+    public void checkIfSavedFileExist() {
```
Yttruire: Since this is a test rather than a simple boolean check (if a certain file exists), and is instead a test for whether a save file can be saved and reloaded, I think it can be named more appropriately. E.g. `saveFileTest` perhaps?

Yttruire:
Everything looks great, and I think that it is an overall job well done! There are no major issues, and the code base was easy to understand. There are just some small pointers that we have both brought up.

IceWizard4902:
Thanks for the reviews! Really appreciate you guys going through my code!

PR #354, opened 2021-08-24
[Ee Kar Hee, Nicholas] iP

kheekheekhee:
# Duke

> "The inner machinations of my mind are an enigma." - Patrick Star ([source](https://youtu.be/1z-pVZiRjac))

Duke helps you keep track of tasks you have to do. It's
* text-based
* easy to learn
* ~~maybe slow :(~~ _EXTREMELY FAST_ 😫👌
* **FREE** 💦💦💦

All you have to do is:

1. download it
2. double-click it
3. add whatever you want
4. relax and let it manage your tasks for you!

Features:
- [x] Managing tasks (todos)
- [x] Managing deadlines
- [x] Managing events

If you're a Java programmer, you can use it to practice Java too! Here's the `main` method:
```
public class Main {
    }
}
```


src/main/java/duke/exception/IncorrectFormatException.java
```diff
@@ -0,0 +1,18 @@
+package duke.exception;
+
+/**
+ * Exception of Duke that occurs when an incorrect format is used for event or deadline.
+ */
+public class IncorrectFormatException extends DukeException {
+
+    /**
+     * Constructor of IncorrectFormatException.
+     * @param command event or deadline depending on what user called.
```
kevinchua6, line 11: Should we have an empty line between the description and parameter section?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,96 @@
...
+            ui.printMessage(e.getMessage());
+        } finally {
+            parser = new Parser(tasks, ui, storage);
+        }
+    }
+
+    /**
+     * Runs the command line version of the Duke chat bot.
+     */
+
```
kevinchua6: Shouldn't we not have an empty line between the documentation block and the method?

src/main/java/duke/exception/InvalidIndexException.java
```diff
@@ -0,0 +1,18 @@
+package duke.exception;
+
+/**
+ * Exception of Duke that occurs when attempting to access an index out of bounds of the task list.
+ */
+public class InvalidIndexException extends DukeException {
+
+    /**
+     * Constructor of InvalidIndexException.
+     * @param start starting index of task list.
```
kevinchua6, line 11: Should we have an empty line between the description and parameter section?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+import duke.storage.Storage;
+import duke.tasks.Task;
+import duke.tasks.TaskList;
+import duke.ui.Ui;
+
+
+/**
+ * Handles and interprets user commands for Duke.
+ */
+
```
kevinchua6: Maybe we should not have an empty line between the documentation block and the method to comply with the coding standard?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,196 @@
...
+        this.filePath = filePath;
+    }
+
+    /**
+     * Saves and writes the current tasks to the data file.
+     * If the file does not exist, it will create a new file.
+     *
+     * @param taskList contains all current tasks.
+     * @throws IOException if the named file exists but is a directory
+     * rather than a regular file, does not exist but cannot be created,
```
kevinchua6, line 58: Should we indent continuation lines four (or more) spaces from the position of the `@`? I noticed the same issue in several other places too.

src/main/java/duke/tasks/TaskList.java
```diff
@@ -0,0 +1,247 @@
...
+        this.taskList = taskList;
+    }
+
+    public ArrayList<Task> getTaskList() {
+        return taskList;
+    }
+
+    /**
+     * Adds a task to the list of Tasks with a confirmation message printed out after.
+     * @param task The duke.tasks.Task to be added to the list of Tasks
```
kevinchua6: Should there be punctuation after each parameter description?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,96 @@
...
+    private void run() {
+
+        String input;
+
+        ui.printMessage(Ui.greet());
```
bryanwee023: Maybe consider `ui.greet()` internally print the greeting message?

Rerouting ui's return value back to ui seems to break OOP's principle of encapsulation

src/main/java/duke/Launcher.java
```diff
@@ -0,0 +1,12 @@
...
+import javafx.application.Application;
+
+/**
+ * A launcher class to workaround classpath issues.
+ */
+public class Launcher {
+    public static void main(String[] args) {
+        Application.launch(Main.class, args);
+    }
+}
```
bryanwee023, line 12: I'm not sure what's the purpose of Launcher. Is it a test class?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+                ArrayList<Task> matchedTasks = findCommand(words);
+                return Ui.printList(matchedTasks);
+            case "": // empty user input
+                throw new EmptyCommandException();
+            default: // all other inputs that are not supported
+                throw new InvalidCommandException();
+            }
+        } catch (DukeException | IOException e) {
+            return e.getMessage();
+        }
```
bryanwee023: I think a switch-case flow might get messy if we have to continue adding more commands.
Maybe consider using a `Hashtable<String, Function>`? 

Alternatively, maybe handle error-checking within the methods invoked (eg. `tasks.deleteTask()`) instead of outside? I think it makes the code more readable because it trims the switch-case block and it complies to OOP's principle of encapsulation as well :)

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Extracts a deadline from the data file in the proper format for duke to read.
+     *
+     * @param text the deadline in the data file.
+     * @return a new deadline that represents the deadline from the data file.
+     * @throws DataFileChangedException if the data file was changed and any entry contains a wrong format.
+     */
+    private Deadline extractDeadline(String text) throws DataFileChangedException {
```
bryanwee023: Just curious to know. What's the reason for placing this function in the Storage class (rather than Deadline)?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,196 @@
...
+        String at = text.substring(lastOccurrenceOfAt + 6, text.length() - 1);
+
+        // prepare variables
+        String atWithoutWhiteSpace = at.replaceAll("\\s", "");
+        int lengthOfAtNoSpace = atWithoutWhiteSpace.length();
+
+        // throws error if it doesn't even contain sufficient number of characters for correct format
+        if (lengthOfAtNoSpace < 22 || lengthOfAtNoSpace > 25) { // MMM d yyyy, HH:mm - HH:mm
+            throw new DataFileChangedException();
+        }
```
bryanwee023: It seems like the parsing of your input string takes up a chunk of your code. I think you can consider abstracting out all these details into a new StringParser class (or push it into the Task classes). 

Maybe doing so might make the code more reader-friendly?

bryanwee023 (commented):
Overall nice job! I like how your Event class has both a start and end time.

I think one thing that can be done is to abstract away details and push them into new methods or into the Task classes (which seems a little too empty)?

But other than that, the code is nicely commented and errors seem to be handled appropriately 👍

PR #377, opened 2021-08-24
[Nigel Tan] iP

nniiggeell:
# Lebron Chat Bot
> “You have to be able to accept failure to get better." -Lebron James

Lebron Chat Bot frees your mind of having to remember things you need to do. It's,
* Text-based
* Easy to learn
* Super fast to use!
All you need to do is:,
1. download it from https://github.com/nniiggeell/ip
2. double-click it.
3. add your tasks.

And it is **FREE**!
Features:
- [x]  Managing tasks
- [x] Managing deadlines 
- [ ] Reminders (coming soon)

```
    public static void main(String[] args) throws Exception {
          new Lebron().run();
      }
```


src/main/java/lebron/task/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+public class TaskList {
+
+    private ArrayList<Task> tasks;
+    private int position;
+    private Ui ui;
+
+    /**
+     * Constructor.
+     *
+     * @param lst The list of tasks.
```
pranav-ganesh, line 22: Shouldn't parameter lebron be included in the javadoc?

src/main/java/lebron/task/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+    public String add(Task task) {
+        tasks.add(task);
+        String reply = ui.replyAdd(this.tasks, task);
+        return reply;
+    }
+
+    /**
+     * This method handles the markDone event when the user says done, marking the Task as done.
+     *
+     * @param pos the position of the Task in the ArrayList lst
```
pranav-ganesh: Shouldn't there be an @return statement in your javadoc since your method returns a value? I noticed the same issue in several other places too...

src/main/java/lebron/task/Task.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public String getName() {
+        return this.name;
+    }
+
+    public String getDoneValue() {
+        return isDone ? "1" : "0";
+    }
+
+    public String toFile() {
```
pranav-ganesh: Shouldn't method names be verbs or contain action words? Not a significant issue though! It would be more ideal if your method name started with a verb, like get, set.....

src/main/java/lebron/Ui.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Represents a response the bot makes to the user.
+ *
+ * @author Nigel Tan
+ */
+public class Ui {
+    private final String HORIZONTAL_LINE = "    ____________________________________________________________\n";
```
pranav-ganesh: Line looks longer than 110 characters. Don't you think It would be ideal to split it into two lines to avoid very long lines?

src/main/java/lebron/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+                String at = splitWords[3];
+                task = new Events(splitWords[2], at);
+                if (isDone.equals("1")) {
+                    task.mark();
+                }
+                taskList.add(task);
+                break;
+            default:
+                System.out.println(HORIZONTAL_LINE
+                        + "    :( OOPS! I'm sorry, but I don't know what that means.\n"
```
pranav-ganesh: I don't see how splitting this print statement into 3 lines improves readability significantly when compared to 2 lines, do you? Wouldn't it be better to just split it into 2 lines since that won't violate the coding standard in any possible way?

src/main/java/lebron/controller/MainWindow.java
```diff
@@ -0,0 +1,70 @@
+package lebron.controller;
+
+import javafx.scene.control.Label;
```
pranav-ganesh: Isn't this inconsistent ordering of import statements? In other files you have added spaces between different types of import statements, here you haven't. Again not a significant issue, but remember that a consistent ordering of import statements makes it easier to browse the list and determine the dependencies when there are many imports.

src/main/java/lebron/task/Deadline.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param description the name of the task
+     * @param by the deadline
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = LocalDate.parse(by);
+    }
+
+    @Override
+    public String toString() {
```
pranav-ganesh, line 42: Shouldn't there be a javadoc for this toString method? Not sure if you have left it out purposely or not!

src/main/java/lebron/task/Task.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public Task(String name) {
+        this.name = name;
+        this.isDone = false;
+    }
+
+    /**
+     * This method marks the task as true.
+     */
+    public void mark() {
```
pranav-ganesh: Isn't the method name 'mark' slightly more ambiguous than 'markAsDone' for example? It is ambiguous in this case whether mark denotes mark as true or mark as false (unless I read the javadoc or look at the method implementation).

src/test/java/UiStub.java
```diff
@@ -0,0 +1,22 @@
+import lebron.Lebron;
```
pranav-ganesh: Shouldn't every class be part of some package? Refer to the Java coding standard (Package and Import Statements section) regarding this.

src/main/java/lebron/Lebron.java
```diff
@@ -0,0 +1,173 @@
...
+    public String run(String text) throws IOException {
+        int position = 0;
+        Parser parser = new Parser();
+        String reply = "";
+
+        if (!text.equals("bye")) {
+            String commandWord = parser.parseText(text);
+            String[] splitWords = parser.split(text);
+            Command command = Command.fromString(commandWord);
+            switch (command) {
```
pranav-ganesh, line 88: I like the fact that you used a switch statement rather than if else! 😄

pranav-ganesh (commented):
Overall I think you have done a good job and your code is well written! Just a few nits to fix here and there!

src/main/java/lebron/Ui.java
nniiggeell: Thanks for spotting this!

PR #385, opened 2021-08-24
[Sreenivasa Kalpana Surya] iP

Tetrerox:
# UhtredRagnarson
> Courage is finding the will to overcome your fear, nothing more.
>  -Uhtred Ragnarson

#### UhtredRagnarson remembers things for you, so *you* don't have to. It's 
* text-based
* easy to use
* has ~~some~~ **NO** learning curve

#### All you need to do is, 
1. download it from [here](https://github.com/Tetrerox/ip/releases/tag/A-Jar)
2. open it without *any* installation
3. Add your tasks
4. And watch the **magic** happen! 👍 

And the *best* part is that you can use it even if you are **BROKE!**

#### Features:
- [X] Add tasks
- [X] Delete tasks
- [X] Mark tasks as completed
- [X] Find matching tasks based on a keyword
- [X] GUI
- [ ] Custom replies to *each* command
- [ ] Better UI design

If you **LOVE** java programming, then you can learn from this project too! Here's the main method:

```java
    }
```


src/main/java/lawbringer/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+                        line.substring(byIndex + 6, line.length() - 1));
+                if (isDone) {
+                    deadline.markAsDone();
+                }
+                tasks.add(deadline);
+                break;
+            case 'E':
+                int atIndex = line.indexOf(" (at: ");
+                Event event = new Event(line.substring(7, atIndex),
+                        line.substring(atIndex + 6, line.length() - 1));
```
elroygohjy: Should the constants be extracted out as final constant variables?

src/main/java/lawbringer/TaskList.java
```diff
@@ -0,0 +1,129 @@
...
+        ui.showTodoMessage(todo, tasks);
+    }
+
+    /**
+     * Add a deadline task to the list and prints a message to the user.
+     *
+     * @param userInput The input of the user.
+     * @param ui        Ui class to print messages to the user.
+     */
+    protected void addDeadlineTask(String userInput, Ui ui) throws DateTimeParseException {
```
elroygohjy: Should you include @ throws as you are throwing DateTimeParseException?

src/main/java/lawbringer/ToDo.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class ToDo extends Task {
+
+    /**
+     * Constructor for this class.
+     *
+     * @param title Description of the to-do.
+     */
+    public ToDo(String title) {
+        super(title);
```
elroygohjy: Do you think title as a name for description is abit ambiguous?

elroygohjy (commented):
I think you have done a great job! The testing part is very clean.

src/main/java/uhtredragnarson/Deadline.java
```diff
@@ -0,0 +1,26 @@
+package uhtredragnarson;
+
+/**
+ * The Deadline class extends the Task class and represents a deadline that has a specific
+ * deadline timing.
+ */
+public class Deadline extends Task {
+
+    protected String by;
```
zhenxuantan: While short, I didn't like the naming for the variable "by". Shouldn't both Deadline and Event have the same variable name (and also they do not consist of only timing) even though the tags "/at" and "/by" are different?

src/main/java/uhtredragnarson/UhtredRagnarson.java
```diff
@@ -0,0 +1,47 @@
+package uhtredragnarson;
+
+import java.io.IOException;
+
+/**
+ * UhtredRagnarson is a chat bot that help users to manage their to-dos, deadlines and events.
+ */
+public class UhtredRagnarson {
+
+    private final Storage storage;
```
zhenxuantan, line 16: Shouldn't these variables be in uppercase? I think the IntelliJ detects it as a warning but we probably have to follow the convention! 😄

zhenxuantan (commented):
Overall the code is very readable and very clean to read! 😄

PR #386, opened 2021-08-24
[Sherman Ng Wei Sheng] iP

shermannws:
# Duke

> The ~Fake~ **Superior** Duke :sunglasses:
> More emojis can be found [here](https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md)

Duke is a CLI task manager application that allows you to unleash your productivity potentials within you. It is,
* Text-based
* User Friendly
* Super fast to use

## Commands that Duke understands

1. **list** *(list out all the outstanding task)*
2. **todo [description]** *(add a todo task)*
3. **deadline [description] /by [yyyy-mm-dd]** *(add a deadline task)*
4. **event [description] /at [yyyy-mm-dd]** *(add an event task)*
5. **done [index]** *(mark task at given index as done)*
6. **delete [index]** *(removes task at given index)*
7. **find [search string]** *(find all corressponding tasks that contains that string)*

## Running Duke

```java
    Duke duke = new Duke("data/duke.txt");
    duke.start();
}
```
You can also use the JAR file directly!
1. Copy the jar file into an empty folder
2. Open a command window in that folder
3. Run the command `java -jar Duke.jar`

## iP Progress
- [x] Basic Functionalities of Duke
- [x] Javadoc
- [x] Unit & Regression Testing
- [x] Gradle
- [x] GUI


src/main/java/duke/MarkDoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+     * Executes the command to mark a task as done.
+     * 
+     * @param list TaskList before execution of the command.
+     * @param ui Ui object to log the execution of the command.
+     * @param storage Storage object that references the path to store the updated list of tasks.
+     * @throws DukeException If index provided is invalid.
+     */
+    @Override
+    public void execute(TaskList list, Ui ui, Storage storage) throws DukeException {
+        boolean markedDone = list.markDoneAtIndex(index);
```
ZhaoPeiduo: Maybe isMarekedDone is a better boolean variable name in this case :)

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,47 @@
...
+        this.isExit = false;
+    }
+
+    /**
+     * Returns true if the command is a programme terminating command.
+     * 
+     * @return True if it is a terminating command and false otherwise.
+     */
+    @Override
+    public boolean isExit() {
```
ZhaoPeiduo: Maybe getIsExit() is a better choice for this method name, as isExit sounds like a boolean name to me.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,26 @@
...
+ * @author Sherman Ng Wei Sheng
+ */
+public abstract class Command {
+
+    /**
+     * Returns true if the command is a programme terminating command.
+     *
+     * @return True if it is a terminating command and false otherwise.
+     */
+    public abstract boolean isExit();
```
ZhaoPeiduo: Similar issue as in AddCommand.class, maybe you need to make the change for Command and its subclasses.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
+package duke;
+
+/**
+ * Represents a Personal Assistant Chatbot that helps a person keeps track of various things.
+ *
+ * @author Sherman Ng Wei Sheng
+ */
+public class Duke {
+    private final Storage storage;
+    private final Ui ui;
```
ZhaoPeiduo, line 17: I like the use of final keyword here to indicate that storage and ui wouldnt be further modified.

src/main/java/duke/FindCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    /**
+     * Returns true if the command is a programme terminating command.
+     *
+     * @return True if it is a terminating command and false otherwise.
+     */
+    @Override
+    public boolean isExit() {
+        return false;
```
ZhaoPeiduo: Maybe replace false with isExit (even though they are always the same)

ZhaoPeiduo (commented):
There are only few violations for java coding standard, as I have pointed out in the comments.
Well done!
(PS: The structure of different components of Duke is really clear and demonstrates strong OOP skills, I have personally learnt some tips from your code :))

src/main/java/duke/Command.java
ZhaoPeiduo: EDIT: My bad your original method name follows the correct naming convention for boolean methods.

src/main/java/duke/AddCommand.java
ZhaoPeiduo: EDIT: My bad your original method name follows the correct naming convention for boolean methods.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+     * Returns the string stored in the storage.
+     * 
+     * @return The string in the storage if it is existent or an empty string otherwise.
+     * @throws StorageLoadingException If error encountered during loading.
+     */
+    public String load() throws StorageLoadingException {
+        try {
+            File storageDir = new File(this.fileDir);
+            File storageFile = new File(this.filePath);
+            boolean isDirExistent = storageDir.exists();
```
ZhaoPeiduo: isDirExistent sounds a bit akward for me, maybe isDirPresent is a better alternative?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Prints the log for task object being added
+     *
+     * @param task The task object to be printed.
+     * @param list The task list object that the task is being added to.
+     */
+    public void printAddedTaskMessage(Task task, TaskList list) {
+        String message =
+                "Got it. I've added this task:\n" +
+                "  " + task + "\n" +
```
ZhaoPeiduo: According to the coding standard, the operator should be at the beginning of the new line. 
屏幕截图 2021-08-31 220809

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,56 @@
...
+     * Starts the Chatbot, listens for user input and executes the command accordingly.
+     */
+    public void start() {
+        this.ui.printGreetings();
+
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String input = this.ui.getInput();
+                Command c = Parser.parse(input);
```
yunpeng1234: Command C can be renamed to reflect it better. Something like ParsedCommand would be more helpful to the readers

src/main/java/duke/MarkDoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+     * Executes the command to mark a task as done.
+     *
+     * @param list TaskList before execution of the command.
+     * @param ui Ui object to log the execution of the command.
+     * @param storage Storage object that references the path to store the updated list of tasks.
+     * @throws DukeException If index provided is invalid.
+     */
+    @Override
+    public void execute(TaskList list, Ui ui, Storage storage) throws DukeException {
+        boolean markedDone = list.markDoneAtIndex(index);
```
yunpeng1234: MarkedDown does not sound like a boolean. Maybe isMarked/isMarkedDone is more appropriate

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+     * Returns the string stored in the storage.
+     *
+     * @return The string in the storage if it is existent or an empty string otherwise.
+     * @throws StorageLoadingException If error encountered during loading.
+     */
+    public String load() throws StorageLoadingException {
+        try {
+            File storageDir = new File(this.fileDir);
+            File storageFile = new File(this.filePath);
+            boolean isDirExistent = storageDir.exists();
```
yunpeng1234: Maybe doesDirExist would be a better name for the boolean

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+            boolean isFileExistent = storageFile.exists();
```
yunpeng1234: Likewise, doesFileExist would fit the naming convention of this class

src/test/java/duke/DeadlineTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.junit.jupiter.api.Test;
+
+public class DeadlineTest {
+    @Test
+    public void encodeTaskForStorage_notDoneDeadlineTask_encodedString() {
+        assertEquals("D | 0 | testingDeadlineTaskOne | 2021-08-26",
+                new Deadline("testingDeadlineTaskOne",
+                        LocalDate.of(2021, 8, 26),
+                false)
```
yunpeng1234: Indentation of false here should be consistent with the rest of the test cases. +4 more spaces to the right

yunpeng1234 (commented):
In general the coding standards are strong adhered to minus some minor changes.

For code quality wise, Duke exceptions can be abstracted out better so you don't have to input the full line to print per case of IllegalFormatting etc.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    private void loadFromStorage(String data) {
+        String[] lines = data.split("\n");
+        if (lines[0].equals("")) {
+            return;
+        }
+        for (String line : lines) {
+            String[] split = line.split(" \\| ");
+            String taskType = split[0];
+            boolean isDone = !split[1].equals("0");
```
neojxuan: Would it be better to catch an array out of bounds exception here? Just in case the file in storage is not formatted properly and split contains less than three elements .

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Prints the formatted list of content in Task List.
+     */
+    public void printList(TaskList list) {
+        String message = list.generateMessage();
+        printMessage(message);
+    }
+
+    /**
+     * Prints the log for task object being added
```
neojxuan: Missing full stop

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,161 @@
...
+     * Marks the task at the index as done.
+     * If given a valid index, returns true and mark the Task as done.
+     * If given an invalid index, returns false.
+     *
+     * @param index The given index to mark the Task located there as done.
+     * @return A boolean value, true if successfully marked as done and false if invalid index given.
+     */
+    public boolean markDoneAtIndex(int index) {
+        if (index > this.size() - 1 || index < 0) {
+            return false;
```
neojxuan: Is it possible to just throw the InvalidIndexException here instead of returning false?

src/main/java/duke/FindCommand.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param storage Storage object that references the path to store the updated list of tasks.
+     * @throws DukeException If problem encountered during execution of command.
+     */
+    @Override
+    public void execute(TaskList list, Ui ui, Storage storage) throws DukeException {
+        int size = list.size();
+        TaskList newList = new TaskList();
+        for (int i = 0; i < size; i++) {
+            Task current = list.get(i);
+            String currentDescription = current.getDescription();
```
neojxuan: Could this be refactored so that a getter is not needed? Perhaps the Task class can have a containsKeyword function that checks if the given keyword is in the description.

src/main/java/duke/MarkDoneCommand.java
```diff
@@ -0,0 +1,54 @@
...
+        if (markedDone) {
```
neojxuan: Similar to the other comment about throwing the exception directly, I believe this boolean value may not be necessary if the exception can be thrown by markDoneAtIndex.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param input The input given by user in the console.
+     * @return A command corresponding to the interpreted intention of the user input.
+     * @throws DukeException If an error is encountered.
+     */
+    public static Command parse(String input) throws DukeException {
+        String[] splitWord = input.split(" ", 2);
+        String firstWord = splitWord[0];
+
+        if (input.equals("bye")) {
```
neojxuan: Do correct me if I'm wrong, but you should be able to use firstWord here as well rather than input. Then you can convert the if-else blocks to case statements which are slightly shorter/may be more readable. However, this is up to your preference haha

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+        } else if (firstWord.equals("todo")) {
+            String[] splitCommand = input.split(" ", 2);
+            if (splitCommand.length == 1) {
+                throw new IllegalFormatException("☹ OOPS!!! The description of a todo cannot be empty.");
+            }
+
+            String toDoDescription = input.split(" ", 2)[1];
+            if (toDoDescription.length() == 0) {
+                throw new IllegalFormatException("☹ OOPS!!! The description of a todo cannot be empty.");
+            }
```
neojxuan: Would it be possible to abstract these two checks into helper functions? For example, the code here could then be condensed into something like ```checkNonEmptyDescription(input, "todo");```. This can be reused in the blocks for other tasks like events and deadlines as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+            if (splitTask.length == 1) {
+                throw new IllegalFormatException("☹ OOPS!!! Specify the deadline date in the format yyyy-mm-dd.");
+            }
+
+            LocalDate deadlineDateTime;
+            try {
+                deadlineDateTime = LocalDate.parse(splitTask[1]);
+            } catch (DateTimeParseException e) {
+                throw new IllegalFormatException("☹ OOPS!!! Specify the deadline date in the format yyyy-mm-dd.");
+            }
```
neojxuan: Could this be into another method like parseDateTime that can be reused for both deadlines and events?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+
+            if (splitCommand.length != 2) {
+                throw new IllegalFormatException("☹ OOPS!!! Specify task to be removed in the correct format.");
+            }
+
+            String secondWord = splitCommand[1];
+            int secondWordLength = secondWord.length();
+            for (int i = 0; i < secondWordLength; i++) {
+                if (!Character.isDigit(secondWord.charAt(i))) {
+                    throw new IllegalFormatException("☹ OOPS!!! Specify the task to be mark done as a number.");
```
neojxuan: Error message should read task to  be deleted rather than to be marked done

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,132 @@
...
+            String[] splitCommand = input.split(" ");
+
+            if (splitCommand.length != 2) {
+                throw new IllegalFormatException("☹ OOPS!!! Specify task to be removed in the correct format.");
+            }
+
+            String secondWord = splitCommand[1];
+            int secondWordLength = secondWord.length();
+            for (int i = 0; i < secondWordLength; i++) {
+                if (!Character.isDigit(secondWord.charAt(i))) {
```
neojxuan: Since something similar is being done in both the done and delete commands, would it be better to abstract some parts into methods?

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,26 @@
...
```
neojxuan: Perhaps the default value returned by isExit() can be false? Assuming there will be fewer terminating commands than non-terminating ones. Then in the terminating commands, this function will be overridden, otherwise there is no need to override.

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,46 @@
+package duke;
+
+/**
+ * Represents a command class that adds a task.
+ *
+ * @author Sherman Ng Wei Sheng
+ */
+public class AddCommand extends Command {
+    private boolean isExit;
```
neojxuan: What is the use of the isExit attribute here? Would it be possible to remove the attribute and just return false in isExit?

neojxuan (commented):
I think your code is generally very clean, clear and easy to understand. 😀 It also follows the Coding Standard very closely. I gave some minor suggestions for refactoring.

PR #389, opened 2021-08-24
[Donny Suriyanto] iP

DonnyS57:
## Alexa

> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes) 

Alexa frees your mind of having to remember things you need to do. It's, 
 * text-based
 * easy to learn
 * ~~FAST~~ *SUPER FAST* to use

All you need to do is,
1. Download it from [here](https://github.com/DonnyS57/ip).
1. Double-click it.
1. Add your tasks.
1. Let it manage your tasks for you :wink:

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

## 

```Java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Scanner;
+
+
+public class Duke {
+
+    private duke.Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+    private static Parser parser = new Parser();
+
```
solarrabbit99, line 13: I believe it was mentioned in one of the later tasks that at least half of the public classes and methods are supposed to have a header comment/javadoc.

src/main/java/duke/Time.java
```diff
@@ -0,0 +1,28 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+
+public class Time {
+    public static boolean validateJavaDate(String strDate) {
+        SimpleDateFormat sdfrmt = new SimpleDateFormat("dd/MM/yyyy");
+        sdfrmt.setLenient(false);
+
+        try
```
solarrabbit99: Is this considered Egyptian-styled brackets? I think it safer as per following:
```java
try {
} catch (Exception e) { // Narrow the catch statement
}
```

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+public class Deadline extends duke.Task {
+
+    private String by;
```
solarrabbit99: I would have preferred names like `deadline` or `dateString` that are more descriptive. See if you agree with that.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                        int indexOfOpenBracketE = data.indexOf("(");
+                        int indexOfCloseBracketE = data.indexOf(")");
+                        String eventDate = data.substring(indexOfOpenBracketE + 4, indexOfCloseBracketE);
+                        String eventTitle = data.substring(9, indexOfOpenBracketE);
+                        Event newEvent = new Event(eventTitle, eventDate);
+                        newStorage.add(newEvent);
+                        break;
+                    default:
+                        break;
+                }
```
solarrabbit99: I believe there are supposed to have no indentations for case clauses.

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Todo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
+    }
+}
```
solarrabbit99: I like the simplicity of the class! Good practice of inheritance!

solarrabbit99 (commented):
Good work!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+        storage = new duke.Storage(filePath);
+        try {
+            Ui.print("Hello! My name is Alexa \nHow can I help you today?");
+            tasks = new TaskList(duke.Storage.load());
+        } catch (DukeException e) {
+            tasks = new TaskList();
+        }
+    }
+
+    public static void main(String[] args) {
```
marcusc55: Should these methods have JavaDoc comments?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,18 @@
+package duke;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends duke.Task {
```
marcusc55: Would it be better to change `extends duke.Task` to `extends Task`, since both classes are in the same package?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,18 @@
...
+    private DateTimeFormatter test;
+
+    public Event(String description, String date) {
+        super(description);
+        this.date = date;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at:" + date + ")";
```
marcusc55, line 40: Would it be better to change the `[E]` string to a named constant? I noticed this for the other classes that extend from Task as well.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,37 @@
+package duke;
+public class Ui {
+
+    public Ui() {
+
+    }
+
+    public static void print(String text) {
+        System.out.println("=======================================");
```
marcusc55: Would it be better to extract the strings in this class (e.g. `==========`, `Congratulations on finishing this task!`) into named constants?

marcusc55 (commented):
Your code is clear and understandable, other than some small comments. LGTM!

PR #393, opened 2021-08-24
[Erin Ong] iP

greyguy21:
# <h1>Duke
> "What can I do for you?"

Duke is a personal assistant that does your remembering for you! It's 
* text - based
* easy to use 
* ~~fast~~ _super fast_! 

All you need to do is, 
1. download it from [here](https://github.com/greyguy21/ip)
2. double-click it
3. add your tasks 
4. let it manage your tasks for you

and the best thing is, **IT'S FREE** 🤩✨!!

Features: 
- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [ ] Reminders (coming soon)
#
If you are a Java programmer, you can use it to practice Java too! Here's the `main` method: 
```java
public class Duke {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,204 @@
...
+                   String name = getName(s);
+                   if (!name.equals("")) {
+                       ToDo toDo = new ToDo(name);
+                       tasks.addTask(toDo);
+                       ui.showTaskAdded(toDo, tasks.getSize());
+                       storage.updateData(tasks);
+                   }
+               } else if (s.contains("deadline")) {
+                   String description = getName(s);
+                   if (!description.equals("")) {
```
Th-429B: Perhaps it would be better if instead of using if-else which leads to deeply nested code (arrowhead style code) to use switch and abstract out some components.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        return -1;
+    }
+
+    private String getName(String s) {
+        try {
+            if (s.contains(" ")) {
+                String[] parts = s.split(" ", 2);
+                if (parts[0].equals("todo") || parts[0].equals("deadline")
+                        || parts[0].equals("event") || parts[0].equals("find")) {
```
Th-429B: Is this comparison necessary? Since getName is only used in parse where there was already a previous comparison to check if the string contains "todo", "deadline", "event", "find".

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        return null;
+    }
+
+    private boolean descriptionIsValid(String description, String pre, String task) {
+       try {
+           if (description.contains(pre)) {
+               if (!description.equals(pre)
+                       && !(description.startsWith(pre) && description.endsWith(" "))
+                       && !(description.startsWith(" ") && description.endsWith(pre))) {
```
Th-429B: Not sure I like how complicated this statement is.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Shows user task has been deleted
+     *
+     * @param task task to be deleted
+     * @param size current number of tasks
+     */
+    public void showTaskDeleted(Task task, int size) {
+        String t = size == 1 ? "task" : "tasks";
+        System.out.printf("Noted.I've removed this task:\n%s\nNow you have %d %s in the list\n", task, size, t);
```
Th-429B: I like the use of format specifiers.

src/main/java/duke/exception/EmptyDescriptionException.java
```diff
@@ -0,0 +1,27 @@
+package duke.exception;
+
+/**
+ * Thrown when no descriptions are provided for tasks
+ */
+public class EmptyDescriptionException extends DukeException {
```
Th-429B, line 6: I like the use of specific custom errors

Th-429B (commented):
Overall, the use of multiple complicated expression makes the code hard to read. I noted some code quality guidelines violation too. However, the use of different specific errors logs are a nice touch!

PR #398, opened 2021-08-24
[Elvis Teo] iP

TLChicken:
# CS2103TaskDashboard

> "There is some _science_ but there is **art** as well." - Damith C. [(source)](https://www.youtube.com/watch?v=5jvOBbP6u2o)

CS2103TaskDashboard helps you overwhelm your mind by reminding you that you still have Weekly Tasks to be done whenever you look at it!

* Does not forget what you need to do.
* Tracks your weekly tasks.
* Convenient to use!

To set it up:
1. Download it from [here](https://github.com/TLChicken/ip/releases/tag/A-Jar).
1. Set it up.
1. Add your weekly tasks.
1. Now look at it every now and then to remind yourself of the tasks you need to do! 

Features:
- [ ] Track down your CS2103T Deadlines!
- [ ] Keeps track of your CS2103T Tasks!
- [ ] Look at your task list to remember what tasks you still have yet to finish! :stuck_out_tongue_winking_eye:

The best part is that CS2103TaskManager is **FREE**!!! (No school fees required!)

Also this app can help you with your Software Engineering skills just like CS2103T! I used Object-Oriented Programming, using concepts like inheritance and polymorphism in our `TodosTask`, `DeadlineTask` and `EventTask` which inherits from `BaseTask`!

Just take a look at this code:
```Java
public class CS2103T {
    private IndividualProject ip;
    private TeamProject tp;
    private TopicsList topicsList;
    
        System.out.println("You still have " + ip.getTasksLeft + ", " + tp.getTasksLeft + ", " + topicsList.getUnreadTopics + " left to do for this week!!!!! :D ");
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+
+    }
+
+    /**
+     * Use to get the current instance of Duke that is running.
+     * Creates a new instance of Duke if there is none running.
+     * @return the Duke object that is running.
+     */
```
James-Kua: Should there be an empty line before @return?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,258 @@
...
+    public static Duke getCurrDuke() {
+        if (dukeInstance == null) {
+            dukeInstance = new Duke();
+        }
+        return dukeInstance;
+    }
+
+    /**
+     * Getter for Duke's Command Parser.
```
James-Kua: Should the first word be a verb like 'gets'? I noticed the same issue in several other places too

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,258 @@
...
+        Duke currDuke = Duke.getCurrDuke();
+
+        currDuke.currUiCtrl.addToDukeBuffer(printLater);
+    }
+
+    private void unknownCommandEntered() throws DukeExceptionBase {
+        throw new DukeExceptionBase("Please enter something valid!");
+    }
+
+    private void addToTDL(String str, BaseTask.TaskType currTaskType) throws DukeExceptionBase {
```
James-Kua: Should it be addToTdl instead? Reference: Abbreviations and acronyms should not be uppercase when used as a part of a name.

src/main/java/duke/DukeListMgr.java
```diff
@@ -0,0 +1,234 @@
...
+
+
+    /**
+     * Used to mark the corresponding task as done.
+     *
+     * @param taskNo The task number in the list.
+     *
+     * @return Duke's output from this command
+     *
+     * @throws DukeExceptionBase when an invalid task is specified or if task is already done.
```
James-Kua: Should these lines have no space between them?

src/main/java/duke/Tasks/DeadlineTask.java
```diff
@@ -0,0 +1,154 @@
...
+            formatterToUse = DateTimeFormatter.ofPattern("d/M/yyyy");
+        }
+
+        // Tries parsing the date string.
+        try {
+            LocalDate byDate = LocalDate.parse(byWhen, formatterToUse);
+            this.byWhenDate = byDate;
+
+        } catch (DateTimeParseException e) {
+            throw new DukeExceptionBase("Deadline Date does not follow the format required. " +
```
James-Kua: Should the operator '+' be included in the next line instead? Reference: Break before an operator.

James-Kua (commented):
I think the there's some inconsistencies in the JavaDocs and it is recurring in different parts. I think it just needs some cleaning up and it should be good to go! Also spotted some minor issues as well.

src/main/java/duke/Tasks/DeadlineTask.java
TLChicken: Thanks, I have fixed all 5 of them.

PR #402, opened 2021-08-24
[Jian Wei] iP

jianoway:
# InferiorDuke

Welcome, ***stalker***, to my *personal* rendition of ~~Duke~~ aka InferiorDuke, where I try to follow requirements for my iP in hopes of passing this module!

> "If it ain't broke, don't fix it" - Thomas Bertram Lance ([Source](https://digital.hagley.org/Nationbiz_197705#page/30/mode/2up))

# Features

As this is a work in progress, most of the features currently available are still being worked upon with many more planned for the future!

## Available Features ✔️

* Able to add 3 different types of tasks (Event, Deadline and ToDo)
* View your Task list with the list command, with your tasks stored locally on your computer
* Mark tasks as done or undone with the done and undo commands
* Delete tasks from the Task list

## Upcoming Key Features 📝

1. Implement Graphical User Interface (GUI) 
2. Implement Gradle
3. Add JavaDocs to all public methods availble

## Additional Optionals

- [X] Add JavaDocs to all public methods
- [x] All code 100% CheckStyle
- [x] Make cute logo for InferiorDuke

# Code Quality

Hopefully, my code is in accordance to the quality acceptable by our tutors and professors. An sample would be as follows:

```java
}
```

## Errors

One thing you might notice about my code is that I chose to create my errors as DukeErrors as seen below:

```java
    public class DukeException extends Exception
```

I used to use DukeErrors as an enumeration but I decided that making it extend the Exception class would make the flow of logic cleaner.


src/main/java/duke/errors/DukeException.java
```diff
@@ -0,0 +1,125 @@
...
+        SEPARATOR_DETECTED(11, "Please do not use '_~_' in your input as it breaks me! ):"),
+        INVALID_DATE_FORMAT(12, "Invalid date. Please use the dd-mm-yyyy convention!"),
+        POLLUTED_EXIT_COMMAND(13, "Please only type 'bye' to exit"),
+        EMPTY_SEARCH_STRING(14, "Please key in something for me to find!");
+
+        /** The error code. */
+        private final int code;
+
+        /** The error message. */
+        private final String description;
```
limdanqi: Perhaps code can be named as errorCode for better clarity, the same for description, getCode() and getDescription() methods below?

src/main/java/duke/errors/DukeException.java
```diff
@@ -0,0 +1,125 @@
...
+            this.code = code;
+            this.description = description;
+        }
+        /**
+         * This method gets the respective DukeError from the code.
+         *
+         * @param code The error code
+         * @return The respective DukeError
+         */
+        public static DukeExceptionType getError(int code) {
```
limdanqi: This method is fine but perhaps the method can be shorter for better readability? Maybe instead of having all the exceptions under DukeException, there could be multiple sub-classes?

src/main/java/duke/parser/CustomDateFormatter.java
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @param text The string to be formatted
+     * @return
+     */
+    public LocalDate formatWithoutTime(String text) {
+        for (String pattern: patternsWithoutTime) {
+            try {
+                return LocalDate.parse(text, DateTimeFormatter.ofPattern(pattern));
+            } catch (DateTimeParseException e) {
+                // Don't want to print anything here as it would disrupt the UI
```
limdanqi: Perhaps there could still be errors printed in the case whereby the date provided is invalid?

src/main/java/duke/parser/Cleaner.java
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * This method cleans the user input for the list command and flags errors if found.
+     *
+     * @param fullCommand User input
+     * @return The cleaned list command
+     */
+    private String listCleaner(String fullCommand) {
+        if (fullCommand.strip().length() >= 5) {
+            return "error 1";
```
limdanqi: Maybe a named constant for all the errors can be used instead as this is a case of magic string? Also, it can help those that are reading your code understand what the error is.

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * This method edits the file stored locally to edit mark as done,
+     * undo or delete the Task specified by the lineNumber.
+     *
+     * @param lineNumber The line on which the Task is to be edited
+     * @param commandType The type of edit to be made
+     */
+    public void editStorage(int lineNumber, CommandType commandType) {
+        int actualLineNumber = lineNumber + 1;
+        try {
```
limdanqi, line 42: The try block could be shorter by implementing more helper methods?

limdanqi (commented):
In general, I find your code to be pretty cool. However, the code can be more succinct so that it can appear to be cleaner and easier to read. Good job! :))

src/main/java/duke/errors/DukeException.java
jianoway: Thanks for the suggestion, will implement in my next commit! :)

src/main/java/duke/errors/DukeException.java
jianoway: Was thinking of that as well, but was afraid that having 13 and more subclasses would seem excessive. Will try to implement as subclasses!

src/main/java/duke/parser/CustomDateFormatter.java
jianoway: Missed this out, was thinking of the same thing after adapting the More-OOP structure but didn't get around to implementing in the UI. Thanks~

src/main/java/duke/parser/Cleaner.java
jianoway: I didn't know what magic string was until you mentioned it, thanks for that haha learnt something new!

src/main/java/duke/storage/Storage.java
jianoway: Ah yes, it does seem pretty long now that you've mentioned it, will fix it!

src/main/java/duke/errors/DukeException.java
```diff
@@ -0,0 +1,33 @@
...
+     * Gets the error message of the DukeError object.
+     *
+     * @return The error code
+     */
+    public int getErrorCode() {
+        return this.errorCode;
+    }
+
+    @Override
+    public String toString() {
```
Siddharth-Sid: I think in general we don't do use toString to get the exception details instead we use a method called getMessage which is a method of the Exception class. That said I think this is fine as well. Excellent work on the multiple types of exceptions you have created from this abstract class.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public String toString() {
+        String str = super.toString();
+        str += String.format(" (at: %s)", this.date.format(DateTimeFormatter.ofPattern("dd MMM yyyy")));
+        return str;
+    }
+
+    @Override
```
Siddharth-Sid, line 26: May I recommend that you call super.getDataString() instead

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public String toString() {
+        String str = super.toString();
+        str += String.format(" (by: %s)", this.date.format(DateTimeFormatter.ofPattern("dd MMM yyyy")));
+        return str;
+    }
+
+    @Override
+    public String getDataString() {
```
Siddharth-Sid, line 28: May I recommend that you call super.getDataString() instead

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,59 @@
+package duke.task;
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected TaskType taskType;
+    enum TaskType {
```
Siddharth-Sid: Is this Enum needed? 
Also if needed it is better to declare it outside this class i.e. in a separate file

src/main/java/duke/task/ToDo.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * This is a constructor for the ToDo object.
+     *
+     * @param description The description of the ToDo
+     */
+    public ToDo(String description) {
+        super(description);
+        this.taskType = TaskType.T;
+    }
+
```
Siddharth-Sid, line 14: I recommend you Override the getDataString Method here and simply call the super.getDataString.

I know it may seem redundant but I think it is a better practice.

text-ui-test/ACTUAL.TXT
```diff
@@ -0,0 +1,41 @@
...
+Here are the tasks in your list:
+1. [T][ ] Water Plants
+2. [T][X] Feed Cows
+3. [E][ ] Kay's Birthday (at: 32 May)
+4. [D][X] Submit proposal (by: 33 May)
+Oh no! I've marked this task as undone:
+  [D][ ] Submit proposal (by: 33 May)
+Noted. I've removed this task:
+  [T][X] Feed Cows
+Now you have 3 tasks in the list.
```
Siddharth-Sid, line 41: Not that this is needed now but if you are testing the Input you should also test all the possible error logs since it will help track if anything changes for the both the Happy Flow and the Error flow

Siddharth-Sid (commented):
I think you have done an excellent job and there are little or not errors. I have made a couple of suggestions that may help you so request you to take a look.

src/main/java/duke/errors/DukeException.java
jianoway: Thanks for the suggestion! Wasn't aware of using getMessage haha have implemented the change~ :)

src/main/java/duke/task/Event.java
jianoway: Thanks for the suggestion! Made my code much cleaner!! (:

src/main/java/duke/task/ToDo.java
jianoway: Cool! Will proceed to implement (:

text-ui-test/ACTUAL.TXT
jianoway: To be honest, these test files were from like week 2/3 and are really outdated. Will be revamping it soon! Thanks for noticing!!

src/main/java/duke/task/Task.java
jianoway: Only implemented the enum to fulfil the 'A-Enum' optional so I think I shall leave it in for now~ Will follow your suggestion to declare it outside the class~!

PR #403, opened 2021-08-24
[Clive Ong] iP

cliveong:
This is a task manager with the following features
* Adding ToDo
* Adding Deadlines
* Adding Events

All you need to do is
1. Download the [Duke.jar file](https://github.com/cliveong/ip/releases)
2. Run it!

You can have **ALL OF IT** for ~a smol price of *$13.37*~ free! :grinning:

Stuff I still need to do here:
- [x] Finish week 3 requirements
- [x] Add gradle support
- [x] Add gui
- [ ] Actually make it nice

If you face any issues while using it `<☹ OOPS!!! I'm sorry>` will try to fix asap.

___
If you have read till here,

```java
public class Main {
         System.out.println("Thank you");
    }
}
```


src/main/java/duke/InvalidArguement.java
```diff
@@ -0,0 +1,7 @@
+package duke;
+
+public class InvalidArguement extends DukeException{
```
trash-bin99: Just a misspelling, InvalidArguement should be InvalidArgument.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+            } else if (checkInvalidDescription(input)) {
+                throw new InvalidDescription(input);
+            } else if (input.length() > 0) {
+                throw new InvalidArguement();
+            }
+
+        } catch (DukeException e) {
+            System.out.println(e.getMessage());
+        }
+    }
```
trash-bin99: I think it would be useful to include some comments in between cases, to help guide the thought process for each condition.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Check if description is missing
+     *
+     * @param input
+     * @return validity
+     */
+    private boolean checkInvalidDescription(String input) {
+        if (input.trim().equals("todo") || input.trim().equals("deadline") || input.trim().equals("event")) {
```
trash-bin99: This is perfectly valid, but I think it would be better to trim the input once beforehand then compare to the three cases:

```

String trimmedInput = input.trim();
if (trimmedInput.equals("todo") || trimmedInput.equals("deadline") || trimmedInput.equals("event")) {
    //rest of the code

```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+
+/**
+ * Manages the writing and creation of files duke.txt
+ */
+public class Storage {
+    String filepath = "src/main/data/duke.txt";
```
trash-bin99: If this is a constant, might want to make it final and change the name to FILE_PATH.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Mark task as done
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    abstract String logo();
```
trash-bin99, line 31: Perhaps renaming it as getLogo() would sound better.

trash-bin99 (commented):
Hello!! Your code is relatively simple to understand and browse through, which is good! There are only a few issues I noticed, which I feel are relatively minor and easy to fix.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,8 @@
+package duke;
+
+public class DukeException extends Exception{ //cs
```
ntwbruce: I think you might need a white space separating Exception and {

src/main/java/duke/InvalidArguement.java
ntwbruce: Also needs a white space between DukeException and {.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Returns logo corresponding to type
+     *
+     * @return the logo [E]
+     */
+    public String logo() {
+        return logo;
+    }
+
+    private String dateFormat() {
```
ntwbruce: Methods should sound like verbs, so maybe something like formatDate().

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,67 @@
...
+    public void commands(String input) {
+        try {
+            if (input.equals("list")) {
+                taskList.printList();
+            } else if (input.length() > 5 && input.substring(0, 5).equals("done ")) {
+                String[] split = input.split("\\s+");
+                int number = Integer.parseInt(split[1]);
+                taskList.changeStatus(number);
+            } else if (input.length() > 9 && input.substring(0, 9).equals("deadline ")) { //date then time
+                String[] split = input.substring(9).split(" /by ");
```
ntwbruce: Maybe not entirely necessary but it might be good to use more descriptive array names here, to complement comments like the one directly below explaining what the splitting does.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * The parent of deadline, event, todo
+ */
+abstract class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
ntwbruce, line 12: I think you don't need the 'this' keyword for isDone unless it's being shadowed by a constructor parameter (eg. if constructor had a param boolean isDone then you would do this.isDone = isDone).

ntwbruce (commented):
Looks good, just highlighted a few Java code standard issues. Also in general, JavaDoc descriptions should be full sentences, so I believe you need to add full stops. Remember to add descriptions for the parameters in the JavaDocs as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Take in input from user and checks what commands to run
+     *
+     * @param input from user
+     */
+    public String commands(String input) {
+        try {
+            if (input.equals("list")) { //checks for list
+                return taskList.printList();
```
LiuZhi1an: Perhaps can change the method name to something like readCommands or getCommands?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Takes in arraylist of tasks and write the contents to duke.txt
+     *
+     * @param listOfText2
+     */
+    public void save(List<Task> listOfText2) {
+        String temp = "";
+        for (int i = 0; i < listOfText2.size(); i++) {
+            temp = temp + listOfText2.get(i).saveWithFormat() + "\n";
+        }
```
LiuZhi1an, line 47: Changing to more descriptive names like saveToFile might be clearer.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        if(counter == 0) {
+            return "There are no matching tasks in your list";
+        } else {
+            return stringToReturn;
+        }
+    }
+
+    public void load() {
+        File existing = new File("src/main/data/duke.txt");
```
LiuZhi1an: Maybe can change to names like currentFile instead of a boolean sounding name?

LiuZhi1an:
Good job! I think your code is pretty clear and well structured. I like how you credited the functions from the JavaFX tutorials.

src/main/java/duke/InvalidArguement.java
cliveong: Thanks, fixed

src/main/java/duke/Parser.java
cliveong: Good idea

src/main/java/duke/Parser.java
cliveong: Good point!

src/main/java/duke/Storage.java
cliveong: Ty!

PR #405, opened 2021-08-24
[Tay Jun Yang] IP

dannytayjy:


src/main/java/Command.java
```diff
@@ -0,0 +1,17 @@
+public class Command {
+    private String type;
+    private String details;
```
SkyBlaise99: Maybe you should use `detail` instead as
> Plural form should be used on names representing a collection of objects.

src/main/java/Duke.java
```diff
@@ -1,10 +1,333 @@
+import java.io.File;
+import java.io.IOException;
+
 public class Duke {
+    private Parser parser;
+    private Storage storage;
+    private TaskList tasks;
```
SkyBlaise99: Perhaps `private TaskList taskList;` will be better?

src/main/java/Duke.java
```diff
@@ -1,10 +1,333 @@
...
+        if (tasks.size() != 0) {
+            String response = ui.tasksInYourList(tasks);
+            ui.displayResponse(response);
+        } else {
+            throw new DukeException("There are no tasks in your list!");
+        }
+    }
+
+    public void markDone(Command command, TaskList tasks) throws DukeException {
+        String commandDetails = command.getDetails();
```
SkyBlaise99: Similar idea, I think `commandDetail` may be a better choice

src/main/java/Duke.java
```diff
@@ -1,10 +1,333 @@
...
+
+        if (commandDetails.trim().length() > 0) {
+            try {
+                int taskNum = Integer.parseInt(commandDetails);
+
+                if (taskNum >= 1 && taskNum <= tasks.size()) {
+                    tasks.get(taskNum - 1).setDone();
+                    Task taskDone = tasks.get(taskNum - 1);
```
SkyBlaise99: Not really related to coding standard but I think it might be better when you swap them around. Consider this:
``` java
Task taskDone = tasks.get(taskNum - 1);
taskDone.setDone();
```

src/main/java/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    public void loadTasksFromFile(File dataFile, TaskList tasks) throws IOException {
+        FileReader fileReader = new FileReader(dataFile);
+        BufferedReader bufferedReader = new BufferedReader(fileReader);
+        String line;
+
+        while ((line = bufferedReader.readLine()) != null) {
+            String[] task = line.trim().split("\\|");
```
SkyBlaise99: Plural should be used in this case, perhaps consider changing to `tasks`?
I noticed the same issue in several other places too, I will not repeat this further.

SkyBlaise99 (commented):
Keep up with the good work! Just some minor suggestions wrt to the coding standard 😄

Some other side notes unrelated to coding conventions:
1. I can't see the Javadocs from Github, not sure if you don't have them yet or Github does not display it.
1. Avoid arrowhead code style (e.g. you can extract out the first `if` in `deleteTask()` as `if (... <= 0) { throw ... }`)
1. You can try using `checkstyle` as it will point you all the places that violates any rules according to a set of rules defined by `checkstyle.xml`

damithc:
@dannytayjy I'm closing this as you already have another PR sent to this repo ... But do take note of the review comment already given to this PR.

PR #409, opened 2021-08-25
[Yirui] iP

Zhouyirui:
### Duke

> The one for all Task Manager for everyone

You no longer need to write down or memorize your tasks, with Duke, you can:

- Learn to add, delete tasks with ease
- Access your task manger anywhere
- Super fast to  #learn!!

All you need to do is:

1. Download from[ here](www.youtube.com)
2. Launch the application
3. Start adding tasks!

😉Good news! it is **FREE!**

Features coming soon

- [x] Reminders
- [x] Creating accounts

If you like programming, you can. Here's the `main` method:
```ruby
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,35 @@
...
+        String[] detailD = str.split(" /by ", 2);
+        if (detailD.length == 1) {
+            throw new TaskException("When is the deadline?");
+        }
+        DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm");
+        LocalDateTime byTime = LocalDateTime.parse(detailD[1], formatter1);
+        Deadline newD = new Deadline(detailD[0], byTime, false);
+        return newD;
+
+    }
```
sunjc826, line 30:
```suggestion
    }

```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.io.IOException;
+import java.time.format.DateTimeParseException;
+import java.util.Scanner;
+
+
+public class Duke {
+    private TaskList tasks;
+//    private String filePath = "data/tasks.txt";
+//    private String directoryName = "data";
```
sunjc826: nit: remove commented out code
I do however think it is quite normal to leave commented code in an in-development app.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+                if (command.equals("list")) {
+                    tasks.printList();
+                    continue;
+                }
+                String[] parsedCommand = Parser.parseCommand(command);
+
+                switch (parsedCommand[0]) {
+                case "done":
+                    //Marks tasks as done
```
sunjc826:
```suggestion
// Marks tasks as done
```
While this may not be specifically stated in the coding standards, I believe comments look nicer with a space after the `//`.
Likewise for the other inline comments.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,7 @@
+package duke;
+
+public class DukeException extends Exception {
+    public DukeException() {
+        super("Sorry, duke.Duke cannot understand your command");
```
sunjc826: I think more varied exceptions can be considered. Also, a suggestion: `duke.Duke` could be replaced with just `Duke`.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,36 @@
...
+        LocalDateTime atTime = LocalDateTime.parse(detailE[1], formatter1);
+        Event newE = new Event(detailE[0], atTime, false);
+        return newE;
+    }
+
+    @Override
+    public String getSavedAs() {
+        return (super.getSavedAs() + "|" + this.at);
+    }
+}
```
sunjc826, line 56: In general, files are recommended to have a newline character at the end.

```suggestion
}

```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public int getNumTask() {
+        return numTask;
+    }
+//    private void writeToFile(String text) throws IOException {
+//
+//        fw.write(text);
+//        fw.close();
+//    }
```
sunjc826:
```suggestion
```
I believe this responsibility is handled by your `Storage` class, hence this may no longer be necessary.

src/test/java/duke/TodoTest.java
```diff
@@ -0,0 +1,12 @@
+package duke;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+//import org.junit.jupiter.api.Test;
```
sunjc826, line 5:
```suggestion
```

sunjc826 (commented):
On the whole, I like the organization of the code and separation of responsibilities. It is quite clean. My only nitpicks are certain commented out pieces of code, some of which do seem unnecessary and can be deleted safely.

For the future, I also suggest putting a bit more documentation as from what I observe, documentation is only restricted to the class-level and not the methods.

PR #413, opened 2021-08-25
[Liao Keng I] ip

KengXIII:
# Duke

> "2 🐝 or 🚫 2 🐝" - William Shakespeare

Welcome to my Duke repository! It is simple, but definitely has all you need in a task manager.

```
Fairly Simplistic
```

### Introduction

* text-based
* easy to learn
* ~~FAST~~ SUPER FAST to use

### How to use?
1. All you need to do is,

* download it from here.
* double-click it.
* add your tasks.
* let it manage your tasks for you 😉

2. That's it!

All for absolutely free!

### Features

- [x] Managing tasks
- [ ] Managing deadlines (<em>to be implemented</em>)
- [ ] Reminders (<em>to be implemented</em>)
- [ ] Help function using `/help`

[Click here to find out more!](https://github.com/KengXIII/ip)
Stay tuned for more updates in the future!


src/main/java/command/Command.java
```diff
@@ -0,0 +1,82 @@
...
+     * Invalid commands will throw the respective DukeExceptions
+     *
+     * @param input string input by users
+     * @return The specific type of command that can be executed
+     * @throws InvalidCommand If command cannot be understood
+     * @throws NullDescription If AddCommand is intended but without any description
+     * @throws InvalidDateFormat If AddCommand is intended but the date format is not followed
+     */
+    public static Command parse(String input) throws InvalidCommand, NullDescription, InvalidDateFormat {
+        String[] output = input.split(" ");
```
TohWeiJie01: Should the name of the array be outputs?

TohWeiJie01 (commented):
LGTM. The naming conventions for your code generally seems good to me. Good job 😄!

src/main/java/command/AddCommand.java
```diff
@@ -0,0 +1,108 @@
...
+    private void checkDescription(String description, String type) throws NullDescription {
+        int numOfCharacters =  type.length();
+        if (description.length() <= numOfCharacters || description.split(" ").length < 2) {
+            throw new NullDescription(type);
+        }
+    }
+
+    private void verifyDate(String dateInfo, String type) throws InvalidDateFormat {
+        switch(type) {
+            case "deadline":
```
xiongjya: In order to better abide by the Java coding standard, that states that there is no indentation for case clauses, would it be better to remove the indentation here? (and similarly to other areas in your code if you have used switch, case)

src/main/java/command/Command.java
```diff
@@ -0,0 +1,82 @@
...
+        if (output.length == 0 || output[0].isEmpty() || output[0].equals(" ")) {
```
xiongjya: It seems that the three conditions are rather similar, do you think that you can abstract at least one of them away (so it would be easier for others to read/be more straight to the point)?

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public DeleteCommand(int i) {
+        this.taskNumber = i;
+    }
+
+    public void execute(TaskList tasks, Ui ui, Storage storage)  throws InvalidValue {
+        try {
+            Task deletedTask = tasks.deleteTask(taskNumber);
+            storage.write(tasks.getTaskList());
+            System.out.printf("\tNoted. I've removed this task:\n" +
```
xiongjya: The coding standard suggests to break before an operator, in this case '+', so would it be better if this minor detail is changed to follow the standard?

xiongjya (commented):
LGTM 👍 The use of abstract classes for commands and tasks is good, and it is great that you have considered and implemented multiple types of errors/exceptions, but do keep in mind that you have a manageable number of classes/files to work on!

The rest that are suggested would simply be cosmetic changes, as your code looks pretty short (not over 30 lines for function declarations other than switch statements) and avoids using deep nesting.

src/main/java/command/AddCommand.java
Timothyoung97: In general, I think the `addCommand.java` is pretty well done abiding by the coding standard mostly.  However, I believe for `switch` cases in this module are supposed to be aligned with the `switch`? (Same as Jingya's comment)

src/main/java/command/DoneCommand.java
```diff
@@ -0,0 +1,33 @@
...
+import task.TaskList;
+import ui.Ui;
+
+/**
+ * DoneCommand will mark as task as completed when executed.
+ */
+public class DoneCommand extends Command {
+
+    private int taskNumber;
+    private final boolean EXIT = false;
```
Timothyoung97: I like how you make `EXIT` as `private final` and always set to `false` for any command that is not `exit`

src/main/java/exception/NullDescription.java
```diff
@@ -0,0 +1,14 @@
+package exception;
+
+public class NullDescription extends DukeException {
+
+    private String type;
+
+    public NullDescription(String type) {
+        this.type = type;
+    }
+    @Override
```
Timothyoung97: I think there should be a space between lines 9 and 10?

Timothyoung97 (commented):
LGTM 😀
I like how you extract out commands as abstract classes, it keeps your code relatively easy to read!
Overall, I think it is just the switch cases that require some formatting changes 👍

src/main/java/command/DeleteCommand.java
```diff
@@ -0,0 +1,41 @@
...
+                    "\t%s\n" +
+                    "\tNow you have %d tasks in the list.\n", deletedTask, tasks.size());
```
Timothyoung97: I feel that maybe you can try using `String.format()` to help tidy up the coding appearance?

JasonC01 (commented):
The code looks great! But one thing that caught my eyes were the boolean EXIT which kept on appearing at different parts of the code. Maybe its usage can be unified in some way so that EXIT does not have to be repeated. Overall the code really is great!

src/main/java/command/Command.java
KengXIII: I originally named it output as it processed the string into individual words. On second thought, I think it might be ambiguous with the new Command that it returns. Thanks for the suggestion!

src/main/java/command/AddCommand.java
KengXIII: Thanks for the catch, I'll amend them in the future commits!

src/main/java/command/Command.java
KengXIII: I have removed the 2nd case as I cannot think of an example that turns the statement true 🤔 I will keep that in mind and figure out why I had that before, thanks

src/main/java/exception/NullDescription.java
KengXIII: Thanks for the spot!

PR #417, opened 2021-08-25
[Rachel Cheah] ip

RachelCheah:
DukePro

* text-based
* easy to learn
* ~~FAST~~ *SUPER* FAST to use

All you need to do is,

1. download it from [here](https://github.com/RachelCheah/ip/releases/tag/A-Packages).
2. double-click it.
3. add your tasks.
5. And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,50 @@
+package duke;
+
+import duke.commands.Command;
+import duke.exceptions.DukeException;
+import duke.exceptions.NoSuchCommandException;
+import duke.exceptions.NoSuchTaskException;
+
+import java.io.IOException;
+import java.util.ArrayList;
+
```
qinguorui2001, line 17: i recommend to list the imports from java... before those from duke..., since it follows check style.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    public void run() {
+        ui.showWelcomeMessage();
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                //#ui.showLine(); // show the divider line ("_______")
```
qinguorui2001: It's free to delete lines which are commented out to make the code clear, since you can still review the old version through source tree or git if u think it's better to change back.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+            case "event":
+                try {
+                    EventsTask.isLegitInput(input);
+                    return new AddCommand(new EventsTask(EventsTask.getNameInput(input), false, Task.parseDateTime(EventsTask.getDeadlineInput(input))));
+                } catch (NoSuchCommandException ex) {
+                    throw new NoSuchCommandException("Invalid Command");
+                }
+            default:
+                throw new NoSuchCommandException("Invalid Command");
+        }
```
qinguorui2001, line 72: I think you can also use a try catch for the whole switch cases, instead of using try catch for each case for the purpose of making code short.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,57 @@
...
+                StringBuffer sb = new StringBuffer();
+                String line;
+
+                while((line=br.readLine())!=null){
+                    Task temp = Tasklist.parseInput(line);
+                    tasklist.addTask(temp);
+                }
+                return tasklist;
+            }
+
```
qinguorui2001, line 76: it could be better to be careful with the indentations.

src/main/java/duke/Tasklist.java
```diff
@@ -0,0 +1,69 @@
...
+    public Integer getTasklistSize(){
+        return tasks.size();
+    }
+
+    public Task getTask(int index){
+        return tasks.get(index);
+    }
+
+
+
```
qinguorui2001, line 58: Similar problems as above, indentation between methods and javadoc need to be improved.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+import duke.tasks.ToDosTask;
+
+
+public class Parser {
+    Parser(){
+
+    }
+
+    public static Command parse(String input) throws NoSuchCommandException {
+        switch(input.split(" ")[0]) {
```
qinguorui2001:
```suggestions
switch(...) {
case "list":
    ...
...
}
```
Be careful about this, for this module the requirement for writing switch cases is little different from what we encountered before.

qinguorui2001 (commented):
Generally, i think you did a good task, just be careful about some java coding standards.

src/main/java/duke/exceptions/NoSuchCommandException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exceptions;
+
+public class NoSuchCommandException extends Exception{
```
nhjryan: Great use of creating your own exception to deal with errors in the code. Perhaps instead of using `super(message);`, you could print "Invalid command" here and append any additional information after it, since I noticed that most of the messages specified were "Invalid command".

nhjryan (commented):
LGTM! The code adheres to the coding standard specified and is neat and readable. :)

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.File;
+
+
+public class Task {
+
+    private final String name;
+    private boolean done;
```
srj31: Perhaps naming it isDone would be better, since it helps to understand later that the variable is boolean.

src/main/java/duke/tasks/ToDosTask.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public String toStringOutput() {
+        return String.format("%s | %s | %s", "T", (getDone())? "X" : "O", getName());
+    }
+
+    public static String getNameInput(String input) {
+        return input.split("todo")[1].strip();
+    }
+
+    public static void isLegitInput(String input) throws NoSuchCommandException {
```
srj31, line 25: Naming it checkInput would have been better, since isLegitInput would be better for a boolean function.

srj31 (commented):
LGTM. Adding JavaDoc comments would have made the code more readable. The coding standard seems to have been followed pretty well otherwise.

PR #418, opened 2021-08-25
[Elizabeth Chow Ting San] iP

lizchow:
DukePro frees your mind of having to remember things yo need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

All you need to do is,
1. download it from [here](https://github.com/lizchow/ip/releases/download/A-Jar/ip.jar).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :wink:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

***
If you are a Java Programmer, you can use it to practice Java too. Here's the `main`method:
```java
public class Duke {
        new Duke().run();
    }
}
```


src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return String message to be displayed.
+     */
+    public abstract String execute(TaskList tasks, Ui ui, Storage storage);
+
+    /**
+     * Indicates whether the program should exit after this {@code Command}.
+     *
+     * @return {@code true} if the program should exit, {@code false} otherwise.
+     */
+    public abstract boolean isExit();
```
yzhilong:
```suggestion
    public boolean isExit() {
        return false;
    }
```

Would it be better if the logic for isExit is handled within the command class? Since every other command class (excluding ExitCommand) returns false.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,129 @@
...
+            throw new DukeException(String.format("Line %d is missing input values!", lineNo));
+        }
+
+    }
+
+    /**
+     * Writes to file at the specified file path with the given tasks.
+     *
+     * @param tasks Tasks to be written to be file.
+     *
```
yzhilong:
```suggestion
```
I don't think that you need to add a new line between `@param` and `@throws` here

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    private static Command parseDoneInput(String input) {
+        try {
+            validateDescription(input, "done");
+            int taskNo = Integer.parseInt(input) - 1;
+            return new DoneCommand(String.valueOf(taskNo));
+        } catch (NumberFormatException e) {
+            throw new DukeException("Enter a number for a done action!");
```
yzhilong, line 87: Would it make the error clearer to the user if you had different exception classes extending DukeException?

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,114 @@
...
+        TaskList taskList = new TaskList(tasks);
+        Assertions.assertEquals(taskList.markTaskDone(0), new Pair<Boolean, Task>(true, task));
+    }
+
+    @Test
+    public void testInvalidMarkTaskDone() {
+        TaskList taskList = new TaskList(new ArrayList<Task>());
+        Assertions.assertThrows(DukeException.class, () -> taskList.markTaskDone(2));
+    }
+}
```
yzhilong, line 114: Very comprehensive testing!! 👍

yzhilong (commented):
LGTM! I like how comprehensive your javadocs are!

src/main/java/duke/command/Command.java
lizchow: Yep, I think it makes more sense to be handled in the command class!

src/main/java/duke/Storage.java
lizchow: Hmm ok

src/main/java/duke/Parser.java
lizchow: haha yes

PR #424, opened 2021-08-25
[Uyen Ngo] iP

uyencfi:
What Duke can do:
- [x] save your tasks
- [x] mark them as done

What Duke ~~will not do~~ cannot do yet:
- order tasks by importance
- send reminders

Life of a Duke user:
1. Start Duke
2. Type away!
    * hit enter on a super long command but made a typo? Just be patient and type again :relieved:
3. Stop Duke.

__iP may have broken your heart :broken_heart: (it certainly broke mine)__
### *but Duke can heal it :gift_heart:*

Have a [sneak peek](https://www.youtube.com/watch?v=dQw4w9WgXcQ) at the code behind:
```java
public class Duke {
        new Duke().run();
    }
    // ...
}
```


src/main/java/duke/Main.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+
+import duke.logic.Duke;
+import duke.ui.MainWindow;
+import javafx.application.Application;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
```
GabrielWLM, line 11: Perhaps you could split this group of import statements into smaller groups? I noticed similar sizable groups of import statements in several other classes.

src/main/java/duke/TextDuke.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    private void exit() {
+        try {
+            Storage.save(taskList.toSaveFormat());
+        } catch (IOException e) {
+            textUi.showSavingError(e.getMessage());
+        }
+        textUi.showGoodbye();
+        textUi.close();
+        System.exit(0);
```
GabrielWLM: Any reason why you chose not to use a `finally` block?

src/main/java/duke/logic/Duke.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param userCommandText The user input.
+     * @return A string describing the program's response to the user input.
+     */
+    public String getResponse(String userCommandText) {
+        Command command = parser.parse(userCommandText);
+        command.setTaskList(taskList);
+        CommandResult result = command.execute();
+        return result.getMessage();
+    }
```
GabrielWLM, line 48: Perhaps you could mention what this function returns in the summary of the method description?

I noticed a similar trend in some of the method descriptions. Is there a reason you prefer describing what the method is intended to be used for in another part of the code instead of what the method does and returns?

src/main/java/duke/TextDuke.java
```diff
@@ -31,15 +33,15 @@ private void run() {
...
         parser = new Parser();
         try {
             taskList = new TaskList(Storage.load());
         } catch (IOException e) {
-            ui.showLoadingError(e.getMessage());
+            textUi.showLoadingError(e.getMessage());
             taskList = new TaskList();
         }
-        ui.showWelcome();
+        textUi.showWelcome();
```
Dr-Octavius: Minor point, but maybe just leave this as "welcome" instead of "showWelcome"?

src/main/java/Event.java
```diff
@@ -1,11 +1,20 @@
...
+        super(description, isDone);
+        this.at = at;
+    }
+
     public Event(String description, String at) {
         super(description);
         this.at = at;
     }
 
+    public String formatForSave() {
```
Dr-Octavius: Maybe "formatSave" would be the same?

Dr-Octavius (commented):
Overall, great use of OOP, apart from minor nomenclature (naming) issues. Otherwise, LGTM! Keep up the good work 😊

src/main/java/duke/logic/parser/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+        case "delete":
+            return handleDelete(tokens);
+        case "todo":
+            return handleToDo(tokens);
+        case "deadline":
+            return handleDeadline(tokens);
+        case "event":
+            return handleEvent(tokens);
+        case "find":
+            return handleFind(tokens);
```
GabrielWLM: Should a `//Fallthrough` comment be added in these cases as there is no break statement?

GabrielWLM (commented):
Nice job! I like your organization and your code is very readable.

Besides a few comments, I think your code has minimal coding violations. Good Job!

src/main/java/Event.java
uyencfi: Thanks for the suggestion! I couldn't think of a more concise name for this method so I had to put `formatForSave` 😅

src/main/java/duke/TextDuke.java
uyencfi: I wanted to have `show` in the method name to indicate an action, will change to `showWelcomeMessage` to be more specific.

src/main/java/duke/Main.java
uyencfi: Thanks for the review! I configured my IntelliJ import statement organization according to the [AB3 convention](https://se-education.org/guides/tutorials/intellijCodeStyle.html#tweak-import-order), but I'll look into splitting them if it makes them more readable.

src/main/java/duke/TextDuke.java
uyencfi: Oops never noticed this, I guess it's because the app doesn't exit after catching the error so the subsequent code will be executed regardless. But I'll add a `finally` block for clarity!

src/main/java/duke/logic/Duke.java
uyencfi: I don't think the module's Java coding standard really specify what the header comment should contain (do correct me if I'm wrong!) so I followed how the Java API writes comment, which is just a short description of what the method does. And personally I think it would be verbose to include the return value, since there is a `@return` clause dedicated to that purpose already. But I suppose I can check with prof on this issue! Thanks for the input.

src/main/java/duke/logic/parser/Parser.java
uyencfi: I'm not sure if it's needed here, because the function immediately returns in each case so there is no possibility of falling through to a different case.

src/main/java/duke/TextDuke.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Reads the user command, executes it and displays the result until exit signal.
+     */
+    private void runCommandLoopUntilExitCommand() {
+        Command command;
+        do {
+            String userCommandText = textUi.getUserCommand();
```
rgbpokka, line 53: I like how the methods are clearly named and easy to understand!

src/main/java/duke/logic/parser/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Categorizes the input into respective command types,
+     * to prepare for further processing.
+     *
+     * @param tokens The tokenized input array.
+     * @return The command based on the user input.
+     */
+    private Command parseTokens(String[] tokens) {
+        switch (tokens[0]) {
```
rgbpokka: Good use of switch statement here!

src/main/java/duke/logic/parser/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    /**
+     * Parses the tokens for an add event task command.
+     *
+     * @param tokens The tokenized input array.
+     * @return The prepared command.
+     */
+    private Command handleEvent(String[] tokens) {
```
rgbpokka, line 249: Thorough handling of the different cases, nice!

rgbpokka (commented):
Overall the code adheres to the code quality guidelines and I've personally learned from reading your code, great job!

src/main/java/duke/TextDuke.java
uyencfi: Thanks for the review! All the best for your iP too 💪

PR #427, opened 2021-08-25
[Peh Kai Min, Ryan] iP

ryanpeh:
# Duck 🦆

> “The perils of duck hunting are great - especially for the duck." – Walter Cronkite

Duck🦆 frees your mind of having to remember things you need to do. It's,
* text-based
* easy to learn
* ~~FAST~~ _SUPER_ FAST to use ⚡️⚡️⚡️

All you need to do is,
1. download it from [here](https://github.com/ryanpeh/ip/releases).
1. double-click it.
1. add your tasks.
1. let it manage your tasks for you :wink:

And it is **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming eventually)

---

If you're a Java programmer(why?), you can use it to practice Java too (but why would you). Here's the `main` method:
```java{.no-line-numbers}
public class Duke {
    }
}

### CODE REVIEW

File: src/main/java/duke/Parser.java [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+     * @return The corresponding Command.
+     * @throws DukeException If invalid command entered by the user.
+     */
+    public Command parse(String fullCommand) throws DukeException {
+        Matcher matcher = COMMAND_FORMAT.matcher(fullCommand.trim());
+        if (!matcher.matches()) {
+            throw new DukeException("Invalid Command Entered");
+        }
+        String command = matcher.group("command");
+        String arguments = matcher.group("arguments");
```
aryagiat, line 40: Should this variable be in plural form?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+import duke.tasks.Event;
+import duke.tasks.Task;
+import duke.tasks.Todo;
+
+/**
+ * Represents a storage which allows for reading and writing of tasks to a file.
+ */
+public class Storage {
+    private static final Pattern STORE_FORMAT = Pattern.compile("\\[(?<type>\\S)] \\[(?<done> |X)] (?<arguments>.*)");
+    private final String fileName;
```
aryagiat, line 23: should a constant variable be in cammelcase?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+                break;
+            case "E":
+                parts = arguments.split(" \\(at: |\\)");
+                task = new Event(parts[0], parts[1]);
+                if (done.equals("X")) {
+                    task.markAsDone();
+                }
+                taskList.addTask(task);
+                break;
+            default:
```
beefham, line 75: Minor nitpick: maybe you could use the default case to throw an exception again if needed.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+
```
yeppog: perhaps the spaces here could be removed unless they serve some purpose?

src/main/java/duke/commands/EventCommand.java
```diff
@@ -0,0 +1,51 @@
...
+     * Constructs an EventCommand object with given arguments.
+     * @param args Arguments entered for the event.
+     * @throws DukeException Invalid arguments entered.
+     */
+    public EventCommand(String args) throws DukeException {
+        Matcher matcher = EVENT_FORMAT.matcher(args.trim());
+        if (!matcher.matches()) {
+            throw new DukeException("Events must be in the format 'event <description> /at <dd MMM yyyy HH:mm>");
+        }
+        description = matcher.group("description").trim();
```
yeppog, line 33: might be nitpicking but would using the `this` keyword help to convey that the variable that is being assigned is an instance variable?

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Constructor for the Deadline.
+     *
+     * @param description Description of the deadline.
+     * @param by The date the deadline is due.
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = LocalDate.parse(by, formatter);
```
yeppog, line 21: for some constructors, `this` is utilised but not in all, perhaps we could standardise this?

src/main/java/duke/tasks/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a deadline entered by the user.
+ */
+public class Deadline extends Task {
+    protected final LocalDate by;
+    private DateTimeFormatter formatter = DateTimeFormatter.ofPattern("dd MMM yyyy");
```
yeppog: could possibly be declared `final` since we do not mutate its value.

yeppog (commented):
Good job!

src/main/java/duke/Storage.java
ryanpeh: Constant names must be all uppercase using underscore to separate words, as mentioned in the java coding standard.

src/main/java/duke/Storage.java
ryanpeh: Fixed, thanks!

src/main/java/duke/Storage.java
ryanpeh: Fixed, thanks!

src/main/java/duke/commands/EventCommand.java
ryanpeh: Actually in the advanced coding standard, it recommends only using the `this` keyword only when a field is shadowed by a method or constructor parameter.

src/main/java/duke/tasks/Deadline.java
ryanpeh: In this case, the `this` is necessary, therefore I didn't remove it.

src/main/java/duke/tasks/Deadline.java
ryanpeh: Fixed, thanks!

PR #428, opened 2021-08-25
[Sharmaine Cheong] iP

sharmainec:
### Tokio

> 
Tokio remembers things for Rio, so Rio don't have to! It is 
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

Features:
1. Managing tasks
2. Managing deadlines
3. Managing events

---

If you Java programmer, you can use it to practice Java too. Here's the ``main`` method:
```java
public static void main(String[] args) throws IOException {
        Duke duke = new Duke("data/tasks.txt");
        duke.run();
    }
```


src/main/java/viper/Instruction.java
```diff
@@ -0,0 +1,29 @@
...
+    TODO("todo"),
+    EVENT("event"),
+    DEADLINE("deadline"),
+    DELETE("delete"),
+    BYE("bye"),
+    INVALID("invalid"),
+    EMPTY("empty"),
+    ;
+
+    private final String s;
```
James-Kua: Should this variable have a more descriptive name to let the user know what this variable is storing?

src/main/java/viper/Storage.java
```diff
@@ -0,0 +1,149 @@
...
+            if (currLine.equals(lineDone)) {
+                task.setDone();
+                fw.write(task + System.lineSeparator());
+            } else {
+                fw.write(currLine + System.lineSeparator());
+            }
+        }
+        
+        fw.close();
+        boolean success = tempFile.renameTo(file);
```
James-Kua: Should this boolean have a more intuitive name like isRenameSuccess?

src/main/java/viper/Duke.java
```diff
@@ -0,0 +1,47 @@
...
+        } catch (FileNotFoundException | DukeException | ParseException e) {
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
```
James-Kua: Would renaming boolean isRunning = true, followed by while (isRunning) be more intuitive?

James-Kua (commented):
Overall, I thought that the code was very neat and tidy and the naming of variables were also very intuitive which made it easy for me to understand what was happening as I was going through your code. Each function also has a clear sole purpose. Nice!

src/main/java/tokio/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+                ui.printBye();
+                PauseTransition delay = new PauseTransition(Duration.seconds(2));
+                delay.setOnFinished(event -> Platform.exit());
+                delay.play();
+            }
+            return command.execute(tasks, ui, storage);
+        } catch (DukeException | IOException e) {
+            return e.getMessage();
+        }
+    }
```
JosephN37: A friendly reminder to update the javadoc for this method :)

src/main/java/tokio/commands/AddCommand.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class AddCommand extends Command {
+    protected String cmdLine;
+
+    public AddCommand(String cmdLine) {
+        this.cmdLine = cmdLine;
+    }
+
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
```
JosephN37: Maybe you can make this function easier to be understood by providing javadocs, though the name might be self-explanatory to some extent

PR #438, opened 2021-08-25
[Tan Joe Wel] iP

tanjoewel:
# Duke
## The task manager you never knew you needed

> "Your mind is for having ideas, not holding them" - David Allen ([source](https://dansilvestre.com/productivity-quotes/))

#### Duke frees your mind of having to remember things you need to do. It's,

* text-based
* easy to learn
* very fast to use

and all you need to do is `while(True)` ~~all this very tedious things~~,

download it from here.
~~Insert download link here~~
1. run the file
1. add your tasks.

Features:

- [ ] Creating tasks
- [X] Marking tasks as done
- [ ] GUi (Coming soon!)

Now with 
```Java
System.out.println("200% more types of tasks!")
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * Executes when Duke receives an event command
+     * Adds an event task to the list of tasks
+     */
+    public void event_execute() {
+        try {
+            int i = userInput.indexOf("/");
+            String description = userInput.substring(6,i-1);
+            //String time = userInput.substring(i+1);
```
Jeromesyl: Should this line be left here? Is it still needed?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+            String message = "    ----------------------------\n"
+                    + "Got it, I've added this task: \n"
+                    + A.toString() + "\n"
+                    + "Now you have " + Duke.index + " tasks in the list\n"
+                    + "    ----------------------------";
+            System.out.println(message);
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("Oops! The description of a deadline cannot be empty and must contain a time!");
+        }
+    }
```
Jeromesyl: Since Todo, Deadline, and Event all inherit from Task, perhaps it would be better to make this an abstract method in the Task class and implement them in the individual sub classes instead?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param description
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getTaskIcon() {
+        return "a";
+    }
```
Jeromesyl: You are missing a javadoc comment here.

src/main/java/duke/Parser.java
Jeromesyl: Or you could just have the method in the Task class and override the method in each subclass.

Jeromesyl (commented):
LGTM. Hardly any coding standard violations. I like how organized and easy to read your code is.

tanjoewel:
Hi Jerome thanks for reviewing my code! Appreciate the input!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
+package duke;
+
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
```
charliemoweng, line 5: Maybe can separate this from the two above?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Duke is an application to help store and track tasks.
+ * Users are able to add and delete various types of tasks such as deadlines, events, and things to do.
+ * Duke can also list the existing tasks and mark tasks as done.
+ */
+
+@SuppressWarnings("checkstyle:Regexp")
+public class Duke {
+
+    private int index = 0; //will figure out a way to make this private :<
```
charliemoweng: what is this comment for? isn't it already private?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+
+    public static void main(String[] args) {
+
+        Duke bot = new Duke("data/Duke.txt");
+        Scanner scanObj = new Scanner(System.in);
+
+        bot.ui.printGreeting();
+
+        while (true) {
```
charliemoweng, line 59: Maybe can extract this out into a run() method? like suggested in the weekly task page.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,215 @@
...
+        this.userInput = userInput;
+        this.index = index;
+        this.duke = duke;
+    }
+
+    /**
+     * Executes when Duke receives a bye command
+     * Sends a goodbye message to the user
+     */
+    public void bye_execute() {
```
charliemoweng: i think for methods need to use camelCase.

charliemoweng (commented):
Overall good job! Keep up the good work.

PR #440, opened 2021-08-25
[Ryan Chung Yi Sheng] iP

NekrozQliphort:
Duke is a day-to-day helper, which attempts to help with:
- [X] Tracking Tasks 
- [X] Backing Up Tasks
- [X] Accompany you because you need **friends**

All you need to know about Duke is:
- Text-based
- Easy to Learn
- Fast to Use

To set up just:
1. Download it from [here](https://github.com/NekrozQliphort/ip)!
2. Click and enter.
3. Literally just play around with it.

Did you know that you can use \``` to fence your code?
Allow me to demonstrate with `main`:
```java
public class Main {
        System.out.println("I want my reward!");
    }
}
```

Here's a reward for reaching the [end](https://youtu.be/dQw4w9WgXcQ)!


src/main/java/duke/CommandParser.java
```diff
@@ -0,0 +1,74 @@
+package duke;
+
+/**
+ * A class representing a parser for commands from the CLI.
+ */
+public class CommandParser extends Parser<String[]> {
+	private final static String[] COMMANDS_WITH_ARGS = new String[] {"todo", "deadline", 
+			"event", "delete", "done", "find"};
+	private final static String[] COMMANDS_WITHOUT_ARGS = new String[] {"list", "bye"};
+	private final static String UNKNOWN_COMMAND_MSG = "☹ OOPS!!! " +
```
simonjulianl: Should we remove the plus to a new line, it's part of the coding standard

src/main/java/duke/CommandParser.java
```diff
@@ -0,0 +1,74 @@
...
+		for (String cmdCheck : COMMANDS_WITHOUT_ARGS) {
+			if (cmd.equals(cmdCheck)) {
+				return new String[]{cmd};
+			} else if (cmdSplit[0].equals(cmdCheck)) {
+				throw new DukeException("☹ OOPS!!! Unknown Argument for " + cmdCheck);
+			}
+		}
+
+		for (String cmdCheck : COMMANDS_WITH_ARGS) {
+			isInvalidCommand = isInvalidCommand && !(cmdSplit[0].equals(cmdCheck));
```
simonjulianl: Should we assign a variable to define a meaning of `!(cmdSplit[0].equals(cmdCheck))` ?

src/main/java/duke/CommandParser.java
```diff
@@ -0,0 +1,74 @@
...
+			throw new DukeException(UNKNOWN_COMMAND_MSG);
+		} else if (cmdSplit.length != 2) {
+			throw new DukeException(String.format(MISSING_ARGUMENT_TEMPLATE, cmdSplit[0]));
+		} else if ((cmdSplit[0].equals("delete") || cmdSplit[0].equals("done"))
+				&& !cmdSplit[1].matches("[0-9]+")) {
+			throw new DukeException(String.format(NON_NUMBER_ARGUMENT_TEMPLATE, cmdSplit[0]));
+		}
+
+		switch (cmdSplit[0]) {
+		case "todo":
```
simonjulianl: Shall we explicitly mention `Fallthrough` here, another suggestion is that you can combine this into a list and check if command in list

src/main/java/duke/DateTimeParser.java
```diff
@@ -0,0 +1,41 @@
...
+	 * 
+	 * @param s The string to be parsed.
+	 * @return A date if the string can be parsed as a date, null otherwise.
+	 */
+	public LocalDate parse(String s) {
+		for (DateTimeFormatter formatter : dateTimeFormatters) {
+			try {
+				return LocalDate.parse(s, formatter);
+			} catch (DateTimeParseException e) {
+				// Don't need to do anything here
```
simonjulianl: Shall we rename this to /* no-op */ ?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A class representing a deadline task.
+ */
+public class Deadline extends Task {
+    private static final DateTimeFormatter DATE_TIME_FORMAT = DateTimeFormatter.ofPattern("MMM dd yyyy");
+    protected String by;
```
simonjulianl: Shall we make it private to keep it simple for now as no classes currently extends `Deadline`

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * A constructor for the deadline task.
+     * 
+     * @param description The description of the task.
+     * @param by The note represented by a string.
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        this.byDate = null;
```
simonjulianl, line 23: `this` is not required as its not shadowed

src/main/java/duke/DateTimeParser.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+
+/**
+ * A class representing a parser that returns a date.
+ */
+public class DateTimeParser extends Parser<LocalDate> {
```
simonjulianl, line 11: I like how you have your own `Parser` with generic, I think it makes the code neat

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,18 @@
+package duke;
+
+/**
+ * An abstract class representing a parser.
+ * 
+ * @param <T> The type of Object for the string to be parsed into.
+ */
+public abstract class Parser<T> {
```
simonjulianl, line 8: This is elegant, but shall we use `interface` instead of `abstract class` here as you are only defining behaviours here?

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,42 @@
...
+     * 
+     * @param isDone The status of the task.
+     * @param description The description of the task.
+     */
+    public Todo(boolean isDone, String description) {
+        super(isDone, description);
+    }
+
+    /**
+     * {@inheritDoc}
```
simonjulianl, line 28: Wah I really just know you can do this!

simonjulianl (changes requested):
Overall really clean and nice code! clear separation of concerns between classes!

src/main/java/duke/CommandParser.java
```diff
@@ -0,0 +1,74 @@
...
+public class CommandParser extends Parser<String[]> {
+	private final static String[] COMMANDS_WITH_ARGS = new String[] {"todo", "deadline", 
+			"event", "delete", "done", "find"};
+	private final static String[] COMMANDS_WITHOUT_ARGS = new String[] {"list", "bye"};
+	private final static String UNKNOWN_COMMAND_MSG = "☹ OOPS!!! " +
+			"I'm sorry, but I don't know what that means :-(";
+	private static final String MISSING_ARGUMENT_TEMPLATE = "☹ OOPS!!! " +
+			"Seems like there are missing argument(s) for %s";
+	private static final String NON_NUMBER_ARGUMENT_TEMPLATE = "☹ OOPS!!! " +
+			"Should have entered a number for %s, BOI";
```
okyntary: Should the "+"s in lines 10, 12 and 14 be moved to be part of the following line, as part of the coding standard?

src/main/java/duke/CommandParser.java
```diff
@@ -0,0 +1,74 @@
...
+		} else if ((cmdSplit[0].equals("delete") || cmdSplit[0].equals("done"))
+				&& !cmdSplit[1].matches("[0-9]+")) {
+			throw new DukeException(String.format(NON_NUMBER_ARGUMENT_TEMPLATE, cmdSplit[0]));
+		}
+
+		switch (cmdSplit[0]) {
+		case "todo":
+		case "done":
+		case "delete":
+		case "find":
```
okyntary: As Simon also said, the coding standard indicates that ```case``` without a ```break``` should have ```// Fallthough``` commented

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,68 @@
+package duke;
+
+/**
+ * A class representing a task.
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    /**a
```
okyntary: Small typo here.

okyntary (commented):
Looks good to me. I don't have much to add about any coding standard violations.

src/main/java/duke/CommandParser.java
NekrozQliphort: > Should we remove the plus to a new line, it's part of the coding standard

Yes it should. Do allow me to change it.

src/main/java/duke/CommandParser.java
NekrozQliphort: Yes, I believe so. Please allow me to change it.

src/main/java/duke/CommandParser.java
NekrozQliphort: I do somewhat agree with this assessment. I have assigned `cmdSplit[0]` to a variable, but as for `cmdSplitFront.equals(cmdCheck)` as a whole, I believe it is only used in 2 places, so it might seem redundant. However, do let me know if you disagree.

src/main/java/duke/CommandParser.java
NekrozQliphort: Yes, do allow me to change that.

src/main/java/duke/CommandParser.java
NekrozQliphort: Agreed, I shall change that.

src/main/java/duke/DateTimeParser.java
NekrozQliphort: Good idea!

src/main/java/duke/Deadline.java
NekrozQliphort: Agreed. I shall change it.

src/main/java/duke/Deadline.java
NekrozQliphort: I added `this` for clarity as there is another constructor that has `byDate` as a parameter, and in my opinion, better indicates this is a object variable.

src/main/java/duke/Parser.java
NekrozQliphort: May I ask for the reason? In my opinion, in the case I want to extend it further to include some variables, which might be the case, it seems an abstract class would work better, but do let me know if you disagree.

src/main/java/duke/Task.java
NekrozQliphort: Thanks! Would not have caught that myself.

src/main/java/duke/CommandParser.java
simonjulianl: I see, in my opinion its more like the same level of abstraction, if the left side is `isInvalidCommand`, the right side usually is not supposed to do some lower-level operations, it just makes it easier to read, not really about redundancy

src/main/java/duke/Parser.java
simonjulianl: hem its fine if you want to extend it in the future, I think its just my preference to not add something that I wont be certain I will add it or not (like just focus on the present first), but it really depends on the scope of the project

src/main/java/duke/Deadline.java
simonjulianl: I see, but I think its just part of our coding guideline that if it's not shadowed, then we should not use null (or use null if and only if its shadowed) unfortunately

simonjulianl: I see, I might have missed it

PR #446, opened 2021-08-25
[Bryan Tee Pak Hong] iP

SpdPnd98:
Duke

* text-based
* easy to learn
* FAST SUPER FAST to use

All you need to do is,

1. download it from here.
2. double-click it.
3. add your tasks.
And it is FREE!

Features:

 - [X] Managing tasks
 - [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/duke/Chatbot.java
```diff
@@ -0,0 +1,126 @@
...
+    private ChatContinue builtInCommands(ChatCommands command) {
+        switch (command) {
+            case BYE:
+                return this.farewell();
+            case LIST:
+                return this.taskList.list(this.ui);
+            default:
+                this.ui.showNotSupported();
+                return ChatContinue.END;
+        }
```
clyveycui: I think the indentation for switch case here might be off.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+            throw new DukeArgumentException("Unknown command provided to duke.Deadline! did you use '/by'?");
+        } else if (commandAndDate.length == 1) {
+            throw new DukeArgumentException("No date specified!");
+        }
+
+        try {
+            this.date = new SimpleDateFormat("dd/mm/yyyy").parse(commandAndDate[1]);
+        } catch (ParseException e) {
+            throw new DukeDateParseException(e);
+        }
```
clyveycui: Interesting design here, putting the parsing of user input in each individual command class.

src/main/java/duke/FileDB.java
```diff
@@ -0,0 +1,52 @@
+package duke;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class FileDB {
+    public static String DEFAULT_SAVE = "ip/src/main/resources/storage.txt";
```
clyveycui: I think this should be lowercase?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,39 @@
+package duke;
+
+public abstract class Task {
+    protected static final String COMPLETED = "[X]";
+    protected static final String NOT_COMPLETED = "[ ]";
+
+    protected final String description;
```
clyveycui, line 10: This might be better if it is in uppercase.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,94 @@
...
+            case TODO:
+                return this.addTodo(input, fileDB, ui);
+            case DEADLINE:
+                return this.addDeadline(input, fileDB, ui);
+            case EVENT:
+                return this.addEvent(input, fileDB, ui);
+            default:
+                ui.showNotSupported();
+                return Chatbot.ChatContinue.CONTINUE;
+        }
```
clyveycui: Same as mentioned above, the indentation for switch case here might be off.

clyveycui (commented):
Good job! Looks good to me, just a few minor fixes.

src/main/java/duke/Task.java
nujiak: I think lowercase is fine since description is not a constant

src/main/java/duke/FileDB.java
SpdPnd98: This should be all caps and separated with underscores because the default save location should be constant

PR #452, opened 2021-08-25
[Wilfred Wong] iP

Wilfredwongkc:
# DukePro #
> “Your mind is for having ideas, not holding them.” – David Allen ([Source](https://dansilvestre.com/productivity-quotes/))

* text-based
* easy to learn
*  ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/Wilfredwongkc/ip/releases/tag/A-Jar).
2. double-click it.
3. add your task.
4. let it manage your task for you :kissing_heart:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

## ##

```java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,83 @@
...
+ * Duke class.
+ * Uses storage class to load existing data.
+ * If there is not existing file, it will create one.
+ * It takes in user input and create the task object.
+ * It will save the final list into the file when programmes end.
+ */
+public class Duke {
+    static void drawLine() {
+        System.out.println("___________________________________________");
+    }
```
tanhuakun: Is this function redundant now that UI can drawLine() too?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public class Parser {
+    String input;
+    public Parser(String input) {
+        this.input = input;
+    }
+
+    public void parse() {
+
+    }
```
tanhuakun: What is the purpose of this method? Perhaps methods with no definition should be in abstract classes.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Task class that is used by duke class to create tasks.
+ * Stores information like description, task type and whether if it is done.
+ * Stores date if the task is a deadline or event.
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected String taskType;
```
tanhuakun, line 14: Do you need to store a different instance of the same String in every object for task type? It could be better to store it as a static field and override getTaskType() in each subclass.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static void main(String[] args) throws IOException {
```
tanhuakun: Should an IOException be allowed to crash the whole program? It could be better to handle it in the functions that could cause the exception.

tanhuakun (commented):
LGTM. There are two redundant methods but otherwise the code and logic works well enough.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class Deadline extends Task{
+    protected LocalDate date;
+
+    public Deadline(String description, String by) {
+        super(description, "Deadline", by);
+        if (by.contains("/")) {
+            date = LocalDate.parse(by.split("/")[2] + "-" + by.split("/")[1] + "-"
+                    + (Integer.parseInt(by.split("/")[0]) < 10 ? "0" + by.split("/")[0] :
+                    by.split("/")[0]));
```
ganhongyao, line 19: Would it be better to parse the `LocalDate` the overloaded method that allows you to use a self-defined `DateTimeFormatter`?

You can find the method documentation [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html#parse(java.lang.CharSequence,java.time.format.DateTimeFormatter)).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,83 @@
...
+        String s;
```
ganhongyao: Should the variable name be something more descriptive, especially since the variable is referenced rather extensively? It may not be clear what `s` represents.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,83 @@
...
+        int taskNumber;
```
ganhongyao: Should the task number be declared and initialised in the respective blocks to keep the scope as small as possible? It seems like it's not being used outside of the `delete` and `done` blocks.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+import java.io.IOException;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+
+import java.util.ArrayList;
+
+import java.util.List;
```
ganhongyao, line 12: Should the line break between these 2 import statements be removed? Especially since they are from the same package.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Loads the file and produce it as list.
+     *
+     * @param p Path of the file.
+     * @return List of task.
+     * @throw IOException.
+     */
+    public List<Task> load(Path p) throws IOException {
+        String currContent;
+        String[] parts;
```
ganhongyao, line 58: Would it be better to declare and initialise the variables within the `for` loop?

You can refer to the coding standard [here](https://se-education.org/guides/conventions/java/intermediate.html#variables).

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+        ArrayList<Task> taskList = new ArrayList<>();
+        List<String> fileContents = Files.readAllLines(p);
+        for (int i = 0; i< fileContents.size(); i++) {
```
ganhongyao: Should there be a whitespace before the operator?
```suggestion
        for (int i = 0; i < fileContents.size(); i++) {
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+            currContent = fileContents.get(i);
+            parts = currContent.split("\\|");
+            if (parts[0].equals("T")) {
+                Todo todo = new Todo(parts[2]);
+                if (parts[1].equals("1")) {
+                    todo.markAsDone();
+                }
+                taskList.add(todo);
+            } else if (parts[0].equals("D")) {
+                Deadline deadline = new Deadline(parts[2],parts[3]);
```
ganhongyao:
```suggestion
                Deadline deadline = new Deadline(parts[2], parts[3]);
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+                if (parts[1].equals("1")) {
+                    deadline.markAsDone();
+                }
+                taskList.add(deadline);
+            } else if (parts[0].equals("E")) {
+                Event event = new Event(parts[2],parts[3]);
```
ganhongyao:
```suggestion
                Event event = new Event(parts[2], parts[3]);
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Loads the file and produce it as list.
+     *
+     * @param tasks List of tasks.
+     * @param path Path of the file.
+     * @throw IOException.
+     */
+    public void write(List<Task> tasks, Path path) throws IOException {
+        String stringToInsert;
+        Task t;
```
ganhongyao, line 100: Should these variables be declared at the start of the loop iteration to keep the scope small?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public Boolean getIsDone() {
```
ganhongyao, line 48: Would it be better to use the primitive type instead? I may have missed out on the reason if this was intended (sorry in advance!)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,63 @@
+package duke;
+
+import java.util.List;
+
+/**
+ * TaskList class that is used by duke class to deal with the list of task.
+ * It allows adding and deleting of tasks from the list.
+ */
+public class TaskList {
+    private List<Task> taskList;
```
ganhongyao, line 10: Would it be better to name this `tasks` instead? It may be confusing for the class name and the field name to be the same.
```suggestion
    private List<Task> tasks;
```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public TaskList(List<Task> taskList) {
+        this.taskList = taskList;
+    }
+
+    /**
+     * Add tasks to the list.
```
ganhongyao: Should it be "Adds" instead? Same for the other javadoc comments :)
```suggestion
     * Adds tasks to the list.
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Class deals with the Ui.
+ * Does the different action depend on what the use types.
+ * It deals with the design (e.g. the logo and the divider line).
+ */
+public class Ui {
+    String logo;
+    Storage storage;
+    TaskList list;
```
ganhongyao: Should these variables be declared as `private` instead?

ganhongyao (commented):
I added a few comments with regards to coding standards. Feel free to ping me if there were parts that I misunderstood! 😄

PR #459, opened 2021-08-25
[Lee Jia Yi] iP

jiayi1129:
# QuestPro

QuestPro frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://github.com/jiayi1129/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.
5. And it is **FREE**!

Features:

- [x]  Managing tasks
- [x]  Managing deadlines
- [x]  Managing events

This is how QuestPro is ran 🏃 
```Java
public class Launcher {
    }
}
```


src/main/java/helpers/Helpers.java
```diff
@@ -0,0 +1,20 @@
+package helpers;
+
+public class Helpers {
```
Py0000: Although it may not be necessary, but perhaps can consider to include comment/brief description of `Helpers` class?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+     * todo read book
+     * deadline return book /by 2021-10-15
+     * deadline return book /by 2021-10-15 18:00
+     * event book fest /at 2021-10-15
+     * event book fest /at 2021-10-15 18:00
+     * done 2
+     *
+     * @param fullCommand Full String command input by user.
+     * @return Corresponding command from the input.
+     */
```
Py0000: I like how the comments are so detailed. 👍

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+import command.Command;
+import command.DeleteCommand;
+import command.DoneCommand;
+import command.ExitCommand;
+import command.FindCommand;
+import command.ListCommand;
+
+import exceptions.DukeException;
+
+public class Parser {
```
Py0000: Similar to previous suggestion, although it may not be necessary, but perhaps can consider to include comment/brief description of `Parser` class?

src/main/java/task/Task.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param string Input String.
+     * @return true if description contains input String, else false.
+     */
+    public boolean checkDescriptionContains(String string) {
+        return description.contains(string);
+    }
+
+    /**
+     * Get task string to be inputted into file.
```
Py0000: Not really sure on this, but perhaps, can replace `Get` with `Gets`? Please do check if I am right.

src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,71 @@
+package parser;
+
+import command.*;
```
Py0000: Perhaps can list out all the import statements explicitly?

src/test/java/task/TodoTaskTest.java
```diff
@@ -0,0 +1,42 @@
+package task;
+
+import exceptions.DukeException;
+import org.junit.jupiter.api.Test;
```
Py0000: Not sure if it is necessary, maybe you can consider to separate the import statements from different packages? Please do check if this is correct.
I have also noticed this in the other test files.

Py0000 (approved):
I like your program! Its it well coded and well commented! It is easy to read and understand the flow and logic of the codes. 👍  

Other than a few minor nitpicks, it LGTM 😄

src/test/java/task/TodoTaskTest.java
jiayi1129: Great Point! Will push this out in the next commits!

src/test/java/parser/ParserTest.java
jiayi1129: Great Point! Will push this out in the next commits!

src/main/java/task/Task.java
jiayi1129: That's a good catch. Will fix it, thanks for pointing out!

src/main/java/parser/Parser.java
jiayi1129: Yes I think it would be good to have a description too. Will add in future commits.

src/main/java/helpers/Helpers.java
jiayi1129: Yes I think it would be good to have a description too. Will add in future commits.

src/main/java/parser/Parser.java
jiayi1129: Thank you! 😄

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,41 @@
...
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+import javafx.application.Application;
+import javafx.stage.Stage;
+
+public class Duke extends Application {
+    private final String DEFAULT_FILEPATH = "data/tasks.txt";
+
+    private final Storage storage;
```
KT27Learn: Perhaps your final variable should be in all uppercase? I've noticed this in other areas as well.
Do correct me if I'm wrong :)

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+     * List of Command Strings.
+     */
+    private static final String COMMAND_EXIT = "bye";
+    private static final String COMMAND_LIST = "list";
+    public static final String COMMAND_TODO = "todo";
+    public static final String COMMAND_DEADLINE = "deadline";
+    public static final String COMMAND_EVENT = "event";
+    public static final String COMMAND_DONE = "done";
+    public static final String COMMAND_DELETE = "delete";
+    public static final String COMMAND_FIND = "find";
```
KT27Learn, line 27: Perhaps an alternative would be to declare the command strings as enum variables instead?

KT27Learn (commented):
Everything looks great, and I think that it is an overall job well done! There are no major issues, and the code base was easy to understand. There are just some small styling suggestions that I have both brought up.

src/main/java/duke/parser/Parser.java
jiayi1129: How would you related the enum variables to the command strings then? Would it be through a Map?

src/main/java/duke/Duke.java
jiayi1129: Good point! I'll work on this, thanks!

PR #460, opened 2021-08-25
[Koh Hong Po] iP

hpkoh:
# WingerIO
> Denny's is for winners

Turns out you are who winger's **always** on the phone with, he helps you
- [x] Manage Tasks
- [ ] Give reminders

Types of tasks currently  supported:
* Todo
* Deadline
* Event

Running it on a terminal
1. Download the jar file [here](https://github.com/hpkoh/ip/releases/tag/A-Jar)
2. Go to the directory where you saved the file
3. Run the command ```java -jar {filename}.jar```

This is how *Duke* class is ran 🏃 
```Java
public class Main {
        new Duke("path/to/save/file").run();
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
...
+        new Duke("./data/duke.txt").run();
+    }
+
+    /**
+     * Run Duke ChatBot
+     */
+    public void run() {
+        ui.printWelcomeMessage();
+
+        boolean notExitCommand = true;
```
kwokieee:
```suggestion
        boolean isExitCommand = false;
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
...
+                String userCommand = ui.readCommand();
+                Command command = Parser.parse(userCommand);
+                notExitCommand = command.execute(this.taskList, this.ui, this.storage);
+                this.storage.save(this.taskList.getFormattedData());
+
+            } catch (InvalidInputException e) {
+                ui.printError(e.getMessage());
+            }          
+        }
+        ui.printExitMessage();
```
kwokieee: Should this be encapsulated within `ExitCommand::execute`?

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,31 @@
+package tasks;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
+
+    private LocalDate time;
```
kwokieee:
```suggestion
    private LocalDate date;
```

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,38 @@
+package tasks;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task{
+
+    private LocalDate startTime;
+    private LocalDate endTime;
```
kwokieee:
```suggestion
    private LocalDate startDate;
    private LocalDate endDate;
```

src/main/java/utils/Parser.java
```diff
@@ -0,0 +1,79 @@
+package utils;
+
+import java.time.LocalDate;
+
+import commands.*;
```
kwokieee: Be careful not to use wildcard imports!

src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,27 @@
+package commands;
```
kwokieee: Would it be better if the root name of the package was the project name?
```suggestion
package duke.commands;
```

kwokieee (commented):
Overall, I found that you followed the coding standard well! 😊 I noted a few minor violations to fix, and I have also made some suggestions to some names that may have been a bit misleading.

src/main/java/Duke.java
```diff
@@ -1,10 +1,47 @@
+import commands.Command;
+
+import exceptions.InvalidInputException;
+import tasks.TaskList;
+import utils.Parser;
+import utils.Storage;
+import utils.Ui;
+
```
wyrchris: Missing headers comments for public classes

src/main/java/commands/AddCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private static final String NEW_TASK_MSG = "New task added:";
+
+    private Task task;
+    
+    public AddCommand(Task task) {
+        this.task = task;
+    }
+
+    public boolean execute(TaskList tasks, Ui ui, Storage storage) {
```
wyrchris: Could include the annotation to alert reader that it is overriding the abstract method
```suggestion
    @Override
    public boolean execute(TaskList tasks, Ui ui, Storage storage) {
```

src/main/java/utils/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Return command according to user input
+     */
+    public static Command parse(String command) throws InvalidInputException {
+        String[] commands = command.split(" ", 2);
+        command = commands[0];
+
+        String description = commands.length > 1 ? commands[1] : "";
+        
+        if (command.equals("bye")) {
```
wyrchris: Could consider using a switch-case statement for greater clarity

src/main/java/utils/Storage.java
```diff
@@ -0,0 +1,77 @@
...
+    public ArrayList<Task> load() {        
+        ArrayList<Task> tasks = new ArrayList<>();
+        
+        try {
+            Scanner dataReader = new Scanner(this.file);
+
+            while (dataReader.hasNextLine()) {
+                String[] data = dataReader.nextLine().split("\\|");
+                String type = data[0];
+                boolean isDone = data[1] == "1" ? true : false;
```
wyrchris: Need not explictly return true/false
```suggestion
                boolean isDone = data[1] == "1";
```

wyrchris (commented):
Other than including comments/Javadocs to the relevant class/method, overall the code looks structured well and is easy to understand!

hpkoh:
# WingerIO
> Denny's is for winners

Turns out you are who winger's **always** on the phone with, he helps you
- [x] Manage Tasks
- [ ] Give reminders


Types of tasks currently  supported:
* Todo
* Deadline
* Event

Running it on a terminal
1. Download the jar file [here](https://github.com/hpkoh/ip/releases/tag/A-Jar)
2. Go to the directory where you saved the file
3. Run the command ```java -jar {filename}.jar```

This is how *Duke* class is ran 🏃 
```Java
public class Main {
        new Duke("path/to/save/file").run();
    }
}
```

src/main/java/Duke.java
hpkoh: Nice catch! Execute handling the message does make more sense!

src/main/java/utils/Parser.java
hpkoh: Thanks for the heads up!

src/main/java/commands/AddCommand.java
hpkoh: Absolutely, great catch again!

src/main/java/Duke.java
hpkoh: isExitCommand does make it clearer! Thanks

src/main/java/tasks/Deadline.java
hpkoh: Since I'm using dates, naming it date does make more sense. Thanks!

src/main/java/utils/Parser.java
hpkoh: I agree! Will implement alongside enums.

src/main/java/utils/Storage.java
hpkoh: Nice catch! Thanks

src/main/java/commands/AddCommand.java
hpkoh: Nice catch! thanks

src/main/java/Duke.java
hpkoh: Header comments will definitely aid in clarity. Will add headers!

PR #461, opened 2021-08-25
[Bernard Wan De Yuan] iP

bernardwan:
### DUKE 

Hi this is my Duke program! It's

- _Fun_ 👍 
- Easy to use
- text-based

Steps to use:

1. Download from [here](https://github.com/nus-cs2103-AY2122S1/ip/pull/461)
2. Run
3. **Use!**

Features:

- [x] Managing tasks
- [ ] GUI

Sample code `main` :
```java
        new Duke().run();
    }
```


src/test/java/dukeTest/TaskListTest.java
```diff
@@ -0,0 +1,31 @@
+package dukeTest;
+
+import duke.DukeException;
+import duke.Task;
+import duke.TaskList;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
```
david-eom: Should `java` related imports come before `duke`?

src/main/java/duke/Command/AddCommand.java
```diff
@@ -0,0 +1,42 @@
+package duke.Command;
+
+import duke.*;
```
david-eom: I think perhaps can consider importing individual classes instead of using wildcard `import`!

src/main/java/duke/Command/AddCommand.java
```diff
@@ -0,0 +1,42 @@
...
+        if(type.equals("todo")) {
+            task = new Task.Todo(description, false);
+
+        } else if(type.equals("deadline")) {
+            String[] temp = description.split("by ", 2);
+            task = new Task.Deadline(temp[0], false, temp[1]);
+
+        } else if(type.equals("event")) {
+            String[] temp = description.split("at ", 2);
+            task = new Task.Event(temp[0], false, temp[1]);
```
david-eom: Maybe can consider using `enum` for this!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public Storage(String filepath) {
+        this.filepath = filepath;
+    }
+
+    /**
+     * Loads task data from storage
+     *
+     * @return ArrayList of tasks
+     * @throws DukeException
```
david-eom: Feel like the condition when the `DukeException` is thrown should be specified here :))

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public String getDescription() {
+        return this.description;
+    }
+
+    public void setStatus(boolean status) {
+        this.status= status;
+    }
+
+    public static class Todo extends Task{
```
david-eom: Should there be a whitespace before the curly braces?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+            return new FindCommand(remaining);
+        } else {
+            String remaining;
+            try {
+                remaining = input.split(" ", 2)[1];
+            } catch (ArrayIndexOutOfBoundsException e) {
+                throw new DukeException("Sorry, tasks must include descriptions.");
+            }
+            return new AddCommand(first_word, remaining);
+        }
```
david-eom: Perhaps we can use `enum` here to simplify the logic here! Also, personally I fell that all the `try-catch` statement here can be abstracted to other class method e.g. `createDone()` etc. to prevent arrowhead style code, and that one single `catch` for `ArrayIndexOutOfBoundsException` could be used for the entire method. Don't know if I articulate my idea in a very accurate way 😂

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,71 @@
...
+            reader = new Scanner(list_data);
+        } catch (FileNotFoundException e) {
+            throw new DukeException("File not found");
+        }
+
+        while (reader.hasNextLine()) {
+            String line = reader.nextLine();
+            String[] data = line.split(" \\| ");
+            switch (data[0]) {
+                case "todo":
```
david-eom: Should `case` be on the same indent level as `switch`?

david-eom (commented):
LGTM! The code looks clean and easy to read, with most of the comments being nit-picky 😂 Think that the logic can be a bit more simplified in `Parser`, and that some javadoc is missing. Excellent code, keep up the good work!! 👍

src/main/java/duke/Command/FindCommand.java
```diff
@@ -0,0 +1,30 @@
+package duke.Command;
+
+import duke.*;
```
chengseong: I think you should import package by package, for example `import duke.TaskList` rather than using `import duke.*`

src/main/java/Command.java
```diff
@@ -0,0 +1,5 @@
+public abstract class Command {
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    public abstract boolean isExit();
```
chengseong: You might want to consider setting `isExit` to `False` in the abstract, then only changing it to `True` when it is an ExitCommand. This way you do not have to constantly check what Command it is

chengseong (commented):
LGTM! I think the code is very clean and easy to read and there is very clear abstraction between classes. Well done!

PR #465, opened 2021-08-26
[Quek Sze Long] iP

szelongq:
# Tipsy

Tipsy frees your drunk mind of having to remember things you need to do. It's,

* text-based
* easy to learn
* ~FAST~ *SUPER FAST* to use

All you need to do is,

1. download it from [here](https://github.com/szelongq/ip/releases).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)
```java
public class Main {
    }
}
```


src/main/java/duke/Event.java
```diff
@@ -0,0 +1,28 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task{
```
twothicc: Perhaps  adding a space after Task would be more consistent with your other class declarations.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,71 @@
...
+        else {
+            this.tasks = new ArrayList<>();
+        }
+    }
+
+    public int getSize() {
+        return tasks.size();
+    }
+
+    public Task getTask(int taskNum) {
```
twothicc: Perhaps it would be good to add a header comment here to tell users that taskNum is using 1-based indexing.
This applies to other methods of TaskList too.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param newTask the new task to be added.
+     */
+    public void addTask(Task newTask) {
+        tasks.add(newTask);
+    }
+
+    /**
+     * Marks a task in the list as done.
+     *
+     * @param taskNum the number of the task to be marked as done.
```
twothicc: Rather than number, wouldn't index be a more suitable term?

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,50 @@
+import duke.*;
```
twothicc: Perhaps it would be better to list imported classes explicitly. 

1. This could keep the import list minimal and up to date. 
2. The class would also be easier to comprehend and maintain.

twothicc (commented):
LGTM :thumbsup:
Only some recommended changes related to coding standard.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,93 @@
...
+        do {
+            ui.printWaitingUserInput();
+            userInput = scanner.nextLine();
+
+            try {
+                executeCommand(userInput);
+            } catch (DukeException e) {
+                ui.printErrorMessage(e.getMessage());
+            }
+        } while (true);
```
cookiedan42: Consider making it obvious here what the loop exit condition is, either as a comment or by code structure. Without tracing executeCommand method, this initially looks like an infinite loop

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,93 @@
...
+        case COMPLETE_TASK:
+            completeTask(Parser.parseTaskNum(userInput));
+            break;
+        case DELETE_TASK:
+            deleteTask(Parser.parseTaskNum(userInput));
+            break;
+        case FIND_TASK:
+            findTask(Parser.parseSearchSubject(userInput));
+            break;
+        }
```
cookiedan42: Consider explaining what happens when invalid input is passed, perhaps as a comment. Not immediately obvious that parseCommand throws exception and control will never flow to switch section

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,110 @@
...
+        case "todo":
+            taskType = TaskType.TODO;
+            break;
+        case "deadline":
+            taskType = TaskType.DEADLINE;
+            break;
+        case "event":
+            taskType = TaskType.EVENT;
+            break;
+        }
```
cookiedan42: Consider explicitly marking out the invalid case using default

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        else {
+            try {
+                Files.createDirectories(savePath.getParent());
+                Files.createFile(savePath);
+            } catch (IOException e) {
+                e.printStackTrace();
+                throw new LoadingException();
+            }
+            return null;
```
cookiedan42: Do all callers of this function have to handle null case?
Consider returning empty ArrayList instead or documenting that the method can return null

src/main/java/duke/Event.java
szelongq: Thanks, I have fixed it!

src/main/java/duke/TaskList.java
szelongq: That's a great point, I have fixed it. Thank you!

src/main/java/duke/TaskList.java
szelongq: I agree! I have fixed it as well as for the other methods as well, thank you!

src/test/java/DukeTest.java
szelongq: Yes, I agree, I had missed this line when checking for code standard violations. Fixed it, thank you!

src/main/java/duke/Duke.java
szelongq: Very true, I agree. I have added a comment to make clear what the loop exit condition is. Thank you!

src/main/java/duke/Duke.java
szelongq: I agree. I have added more comments to elaborate what exceptions will be thrown when invalid inputs are entered. Thank you!

src/main/java/duke/Parser.java
szelongq: Fixed it, thank you!

src/main/java/duke/Storage.java
szelongq: That's a better alternative I had not thought of. I have updated it to return an empty ArrayList instead. Thank you!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+     */
+    public void printLoadTasks(TaskList tasks) {
+        System.out.println(DUKE + "I have loaded your past tasks list!");
+        printTaskList(tasks);
+    }
+
+    /**
+     * Prints the error message when the program encountered an error.
+     *
+     * @param message the error message from the exception.
```
eezj35: Remember to capitalize the first letter in your description!

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,56 @@
+import duke.CommandType;
+import duke.DukeException;
+import duke.MissingInputException;
+import duke.Parser;
+import duke.Task;
+import duke.TaskList;
+import duke.ToDo;
+import duke.UnsupportedOperationException;
+import org.junit.jupiter.api.Test;
```
eezj35: I think maybe you could group the duke package imports and the org.junit imports in separate groups?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,105 @@
...
+     * @param tasks  TaskList containing the tasks to be saved.
+     */
+    public void saveTasks(TaskList tasks) {
+        try {
+            BufferedWriter writer = Files.newBufferedWriter(savePath);
+            for (int i = 1; i <= tasks.getSize(); i++) {
+                writer.write(tasks.getTask(i).toSaveFormat() + "\n");
+            }
+            writer.close();
+        } catch(IOException ex){
```
eezj35: remember the whitespace before the curly brace!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Prints a list of tasks whose names contain the subject.
+     *
+     * @param tasks the TaskList containing all the current tasks.
+     * @param subject the keyword that is being searched for.
+     */
+    public void printTasksWithSubject(TaskList tasks, String subject) {
+        int nResults = 0;
```
eezj35: A very minor point, but maybe you could rename "nResults" to "numOfResults"? Just to improve overall readability

src/main/java/duke/CommandType.java
```diff
@@ -0,0 +1,5 @@
+package duke;
+
+public enum CommandType {
+    EXIT, LIST, ADD_TASK, COMPLETE_TASK, DELETE_TASK, FIND_TASK
```
eezj35: I like your use of enums for this case!

eezj35 (commented):
Great job on the code! It was actually quite hard for me to find much stuff to comment about, and I really liked that your methods were quite short and cleanly written!

szelongq:
> Great job on the code! It was actually quite hard for me to find much stuff to comment about, and I really liked that your methods were quite short and cleanly written!

Thanks for the feedback! I have edited my code accordingly. :)

PR #466, opened 2021-08-26
[Mengzhe] iP

charliemoweng:
# DukeAzure
> “Focus on being productive instead of busy.” - Tim Ferriss, American podcaster, author and entrepreneur

Here at DukeAzure, we help you be productive so you don't have to be busy taking down your tasks everywhere, ever.

### Some selling points include:
- easy to use
- small-sized
- ~~fast~~ *VERY* fast
- Permanently **FREE**

Get yours from [here](https://github.com/charliemoweng/ip/releases/tag/v0.1.1) today!

### Some simple steps to follow:
1. Follow the link above to the webpage.
2. Download it!
3. Run it in your Com :computer: !
4. Start adding your tasks and manage them as you wish.

### Some features (many more are on the way)
- [X] Add and store a task
- [X] Mark a task as done
- [X] Show all your tasks
- [X] Find a specific task
- [ ] Send reminders before a task is due

### For those avid coders out there
Here is a snippet of the code. Feel free to give suggestions!
The `main` method of Duke:

```java
       new Duke(FILE_PATH).run();
} 
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,20 @@
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
+
```
jackgugz: We need to add some JavaDoc comments for all public classes and methods. Noticed the lack of javadoc in some other places too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.FileNotFoundException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
```
jackgugz: Should an empty line be added between imports from different packages? Also shall we order imports from same package by their alphabetical order?

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+
 public class Duke {
+
+    private static void appendToFile(String filePath, String textToAppend) throws IOException {
+        FileWriter fw = new FileWriter(filePath, true); // create a FileWriter in append mode
```
jackgugz: Should comments be indented relative to their position in the code?
```suggestion
        // create a FileWriter in append mode
        FileWriter fw = new FileWriter(filePath, true); 
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+    }
+
+//    private static void printFileContents(File file) throws FileNotFoundException {
+//        //File f = new File(filePath); // create a File for the given file path
+//        Scanner s = new Scanner(file); // create a Scanner using the File as the source
+//        while (s.hasNext()) {
+//            System.out.println(s.nextLine());
+//        }
+//    }
+
```
jackgugz: Shall we avoid such commented out codes?

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+        // if data folder does not exist, create folder and file
+        if (!data.exists()) {
+            data.mkdir();
+            try {
+                File dataDirectory = new File("./data/data.txt");
+                if (dataDirectory.createNewFile()) {
+                    System.out.println("Data file has been created.");
+                } else {
+                    // System.out.println("Data folder is already present.");
+                }
```
jackgugz: Is it possible that we take out the else clause then?

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+
+        // if data.txt does not exist, create file
+        if (!dataFile.exists()) {
+            try {
+                File dataDirectory = new File("./data/data.txt");
+                if (dataDirectory.createNewFile()) {
+                    System.out.println("Data file has been created.");
+                } else {
+                    // System.out.println("Data folder is already present.");
+                }
```
jackgugz: Same as above

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+                    inputs.add(eveToAdd);
+                }
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("data.txt is not found.");
+        }
+
+//        for (int i = 0; i < inputs.size(); i++) {
+//            System.out.println("test read todo and ddl\n" + dataRead.get(i));
+//        }
```
jackgugz: We may delete this as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+
+        // String input
+        while(true) {
+            String input = myObj.nextLine();
+            Task newTask = new Task(input);
+
+            // mark a task as Done
+            Pattern pattern = Pattern.compile("done\\s\\d+");
+            Matcher matcher = pattern.matcher(input);
+            boolean matchFound = matcher.find();
```
jackgugz:
```suggestion
            boolean ifMatchFound = matcher.find();
```
so that it sounds more like a boolean

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+                    System.out.println(inputs.get(index).toString());
+                    System.out.println(lineBreak + "\n");
+                }
+                continue;
+            }
+
+            // delete a task
+            Pattern patternDelete = Pattern.compile("delete\\s\\d+");
+            Matcher matcherDelete = patternDelete.matcher(input);
+            boolean matchFoundDelete = matcherDelete.find();
```
jackgugz: Same to the boolean name here as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+                    System.out.println("    " + newDeadline.toString());
+                    System.out.println("Now you have " + inputs.size() + " tasks in your list.");
+                    System.out.println(lineBreak + "\n");
+                    continue;
+                }
+
+                // Case Event
+                if (group[0].equals(taskEve) && input.contains(" /at ")) {
+                    String[] eveGroup = input.split(" /at ");
+                    String eveToAdd = eveGroup[0].substring(6); // name of the task is after "event" and space
```
jackgugz:
```suggestion
                    // name of the task is after "event" and space
                    String eveToAdd = eveGroup[0].substring(6); 
```

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
```
jackgugz: The ordering of the imports can be better!

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+         this.taskName = taskName;
+         this.isDone = false;
+     }
+
+     public void setDone() {
+         isDone = true;
+     }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
jackgugz:
```suggestion
        // mark done task with X
        return (isDone ? "X" : " "); 
```

jackgugz (commented):
Overall, I see consistent good coding standards being especially for the `for` loops and conditional statements! The spacing and indentation made it super easy to read. However, I do observe some common places for improvements: 
1. ordering of the imports
2. comments indentation and lack of javadoc comments
3. naming of variables. 

Nonetheless still good job! Looking forward to the more updated version as I think it may be a few versions behind the schedule.

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+                if (dataDirectory.createNewFile()) {
+                    System.out.println("Data file has been created.");
+                } else {
+                    // System.out.println("Data folder is already present.");
+                }
+            } catch (IOException e) {
+                System.out.println("An error occurred.");
+                e.printStackTrace();
+            }
+        }
```
1waykiat: Could be extracted out to the Storage class to handle instead

src/main/java/Duke.java
```diff
@@ -1,10 +1,350 @@
...
+        // String input
+        while(true) {
+            String input = myObj.nextLine();
+            Task newTask = new Task(input);
+
+            // mark a task as Done
+            Pattern pattern = Pattern.compile("done\\s\\d+");
+            Matcher matcher = pattern.matcher(input);
+            boolean matchFound = matcher.find();
+            if (matchFound) {
```
1waykiat: Perhaps let Parser class handle the match finding instead, and extract the code there.

1waykiat (commented):
Remember to complete the tasks from last week! 
Mostly just need to extract away the Storage and Parser class so that the parsing and loading and storing of data can be abstracted away.

src/main/java/Deadline.java
charliemoweng: Yes indeed. I will add that in soon. Thank you for pointing it out.

src/main/java/Duke.java
charliemoweng: Yes, you are right. I will reorder them to make them look nicer.

src/main/java/Duke.java
charliemoweng: you are right.

src/main/java/Duke.java
charliemoweng: yes thank you for pointing it out. i have removed it since.

src/main/java/Event.java
charliemoweng: yes i will improve upon it.

PR #471, opened 2021-08-26
[Zhang Yubin] ip

Kimowarui:
**DukePro**

- text-based
- easy to learn
- SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/Kimowarui/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks

```
public class Launcher {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+
+    protected String by;
```
justintanyf: Should this be more specific as to what it refers to?

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+        ArrayList<Task> task = new ArrayList<>();
+        System.out.println("Hello from\n" + logo + line + "\nWhat can I do for you?\n" + line);
+
+        String input;
+        Scanner s = new Scanner(System.in);
+        input = s.nextLine();
+
+        while(!input.equals("bye")) {
+            String[] parts = input.split(" ");
+            if(input.equals("list")) {
```
justintanyf: Might be cleaner to change to switch statements

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task {
+
+    protected String at;
```
justintanyf: Might be good to use a more desceiptive name for the variable

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
justintanyf: Ternary operators might look cleaner when seperated into multiple lines, and there is no need for the brackets.

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+                System.out.println(line);
+            } else if (parts[0].equals("event")){
+                String[] part2 = input.split("/at ");
+                String description = part2[0].split("event ")[1];
+                Task event = new Event(description, part2[1]);
+                System.out.println(line + "\n" + "added: " + event.toString());
+                task.add(event);
+                System.out.println("Now you have " + task.size() + " tasks in your list");
+                System.out.println(line);
+            } else if (parts[0].equals("todo")){
```
justintanyf: I think there should be a space between ")" and "{", so ) {

justintanyf (commented):
Overall, a really good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+                    i++;
+                }
+                System.out.println(line);
+            } else if (parts[0].equals("done")) {
+                int index = Integer.valueOf(parts[1]);
+                Task thisTask = task.get(index - 1);
+                thisTask.markDone();
+                System.out.println(line);
+                System.out.println("Nice! I've marked this task as done:\n" + thisTask.toString());
+                System.out.println(line);
```
weichang18: Maybe for this portion can create an additional class(Example UI) to do all this line of code for you, and just need to call that function from the class so that it looks much neater. Can also be done so on your other if-else statements

weichang18 (commented):
Overall pretty good,  but it seems like you have not done week 2's task. If you have done so, you should commit it so we can help you review your code.

PR #473, opened 2021-08-26
[Lulu Y] iP

luluyousef:
### **DukePro**

- text-based
- easy to learn
- ~~FAST~~ _SUPER_ FAST to use

All you need to do is,

1. download it from [here](https://nus-cs2103-ay2122s1.github.io/website/admin/ip-w4.html).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x]  Managing tasks
____
```java
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+
+    public static void main(String[] args) throws DukeException, IOException{
+        String pathName = "./data/duke.txt";
+        File f = new File(pathName);
+        try {
+            if(!f.exists()){
+                f.createNewFile();
+            }
+            Scanner sc = new Scanner(f);
+            ArrayList<Task> t = new ArrayList<>();
```
Joel-Sung: I am not sure I like the name of the ArrayList. I think giving it a clearer name would improve the readability of your code. For example, 'taskList'.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+                    if (taskArr[1].equals("1")) {
+                        e.markAsDone();
+                    }
+                    t.add(e);
+                }
+            }
+            int ctr = 0;
+            Scanner scanner = new Scanner(System.in);
+            System.out.println("_______________________________________________");
+            System.out.println("Hello! I'm Duke\n" + "What can I do for you?");
```
Joel-Sung: I think placing a line break before the '+' would improve improve the readability of your code. This is because you add a next line character before the '+', so presenting your code in the same way the output will look would be nice.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+                        System.out.println("_______________________________________________");
+                    } else {
+                        throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-( " +
+                                "Try todo, event, or deadline");
+                    }
+                } catch (DukeException d) {
+                    System.out.println(d.getMessage());
+                }
+            }
+        }catch(IOException e){
```
Joel-Sung: I think you may have missed this. Add a space between the } and catch.

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+            }
+            if(t.label.equals("E")){
+                text =  text.concat("E|");
+                if(t.isDone){
+                    text = text.concat("1|");
+                }else{
+                    text = text.concat("0|");
+                }
+                text = text.concat(t.description + "|" + ((Event) t).by + "\n");
+            }
```
Joel-Sung: I think it is better to add a space around every 'if', 'else' and 'for' as this helps the readability of your code.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+public class DukeException extends Exception{
```
Joel-Sung: I think you should add a space after Exception. I noticed that this is a recurring practice for your classes. Adding a space between the class declaration and the { will help to add readability to your code. I suggest you read the part on "White space within a statement" in the Java coding standard 😄 .

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task{
+
+    protected String by;
```
Joel-Sung: I think it would be better to keep your class variables as private. Do you have a reason to declare them as protected?

Joel-Sung (commented):
Overall, your code quality is pretty good in my opinion. Just some things to take note, I think you should start adding spaces within your statements as it is part of the Java coding standard. Maybe also consider declaring class variables as private, unless you have a reason to do otherwise.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
```
tchiong: Could perhaps add header docs to the class!

src/main/java/Duke.java
```diff
@@ -1,10 +1,172 @@
...
+                    Event e = new Event(taskArr[2], taskArr[3]);
+                    if (taskArr[1].equals("1")) {
+                        e.markAsDone();
+                    }
+                    t.add(e);
+                }
+            }
+            int ctr = 0;
+            Scanner scanner = new Scanner(System.in);
+            System.out.println("_______________________________________________");
```
tchiong: Can perhaps use a function to print the dividers to make the code look more readable!

tchiong (commented):
Other than the comments from the other reviewers and myself, LGTM!

PR #477, opened 2021-08-26
[Pranav Ganesh] iP

pranav-ganesh:
### :fire:  DUKE :fire: 

> “Your mind is for having ideas, not holding them.” – David Allen [source](https://dansilvestre.com/productivity-quotes/)

- text-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use

All you need to do is,
1. download it from [here.](https://github.com/pranav-ganesh/ip/releases/tag/A-Jar)
2. double-click it.
3. add your tasks
4. let it manage your tasks for you :smiley: :blush:

And it is **FREE!**

Duke's features:

- [x] Managing features
- [ ] Reminders (coming soon)
#

If you Java programmer, you can use it to practice Java too. Here's the  ``` main``` method:
```java
    new Duke("./duke.txt").run();
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
+package duke;
+
+import duke.command.*;
```
yongxiangng: You might want to avoid wild card imports even if you have to import everything 
https://se-education.org/guides/conventions/java/intermediate.html#package-and-import-statements

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            if (!file.exists()) {
+                file.createNewFile();
+            }
+            this.file = file;
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+    
+    public void loadTaskListData(TaskList taskList) {
```
yongxiangng, line 47: You may want to document public methods with javadocs
https://se-education.org/guides/conventions/java/intermediate.html#comments

The method here is also quite lengthy. It might be a good idea to abstract the other logic into various methods
https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            if (!s.hasNext()) {
+                Ui.printMessage("There are no items in your task list!");
+            } else {
+                Ui.printMessage("Here is your current task list: ");
+                while (s.hasNext()) {
+                    String str = s.nextLine();
+                    System.out.println(str);
+                    String[] parts = str.split("\\|", 4);
+                    String subStr = parts[0].substring(3).trim();
+                    if (subStr.equals("duke.task.Todo") || subStr.equals("duke.task.Todo")) {
```
yongxiangng: I might be wrong here but doesn't `subStr.equals("duke.task.Todo") || subStr.equals("duke.task.Todo")` do the same as `subStr.equals("duke.task.Todo")`

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            this.file = file;
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+    
+    public void loadTaskListData(TaskList taskList) {
+        try {
+            Scanner s = new Scanner(this.file); // create a Scanner using the File as the source
+            if (!s.hasNext()) {
```
yongxiangng: The logic here is quite deeply nested, perhaps it might be good if you handled the unusual case with a guard clause
https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent
```java
if (!s.hasNext()) {
    // code here...
    return;
}
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+                        taskList.addTask(task);
+                    } else if (subStr.equals("duke.task.Deadline") || subStr.equals("duke.task.Deadline")) {
+                        DateTimeFormatter dtf = DateTimeFormatter.ofPattern("MMM d yyyy, h a");
+                        LocalDateTime dateTime = LocalDateTime.parse(parts[3].substring(5).trim(), dtf);
+                        Task task = new Deadline(parts[2].trim(), dateTime);
+                        if (parts[1].trim().equals("X")) {
+                            task.markAsDone();
+                        }
+                        taskList.addTask(task);
+                    } else {
```
yongxiangng: You may want to use a switch statement for this long if else. It may be good because you're comparing strings and it can directly be used for your cases.

Additionally the else statement here captures all other cases of subStr, not necessarily when it is an `Event`. If new requirements come in and there are new type of `Task`, the developer might forget to update the logic here and the new type of `Task` will automatically fall into the else case and treated like an `Event`.

The default case in a switch statement could be used to catch invalid cases. I believe the default case is optional in Java
https://www.geeksforgeeks.org/switch-statement-in-java/

The style guide suggests that
```
Always include a default branch in case statements.

Furthermore, use it for the intended default action and not just to execute the last option. If there is no default action, you can use the default branch to detect errors (i.e. if execution reached the default branch, raise a suitable error). This also applies to the final else of an if-else construct. That is, the final else should mean 'everything else', and not the final option. Do not use else when an if condition can be explicitly specified, unless there is absolutely no other possibility.
```

yongxiangng (commented):
There are many Javadocs missing, it would be good to write header comments for all public classes and methods see [here](https://se-education.org/guides/conventions/java/intermediate.html#comments)

It might also be good to be consistent with the number of new lines at the end of your files rather than no or dangling new lines. I believe it is good to have a single new line at the end.

There are several places where `System.out.println` was directly used, possibly violating the abstraction from `Ui.printMessage` (I guess this might not really matter once you start implementing the GUI, you wouldn't be able to use `System.out.println` to directly print to your UI)

src/main/java/duke/Parser.java
yongxiangng: For the `parse` method below

Perhaps a switch statement here would help to improve readability or you could modify the code by splitting the string by the empty character delimiter twice so you don't have to do another `if` inside the external `if`

Maybe your InvalidCommand can be an exception, and your doneCommand can throw the exception so you don't have to do all the checking within this method which could make the code easier to read

https://www.geeksforgeeks.org/split-string-java-examples/

https://nus-cs2103-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#guideline-maximise-readability

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+
+public class Parser {
+    private String command;
+    
+    public Parser(String command) {
+        this.command = command;
+    }
+    
+    public Command parse() {
```
Linxcathyyy, line 35: Perhaps try not to write deeply nested loops in your function? You might consider using 'switch' for this function?☺️

src/main/java/duke/Storage.java
Linxcathyyy, line 47: You might want to break it down into smaller functions here? This function seems to be too deeply nested.

Linxcathyyy (commented):
Good job overall! Just one thing to take note of: try to use small and simple functions to handle your complex logic, don't put everything into one function. It makes readers confusing and it is also hard to do unit testing~

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+            if (command.startsWith("todo")) {
+                return new todoCommand(command);
+            } else if (command.startsWith("deadline")) {
+                return new deadlineCommand(command);
+            } else if (command.startsWith("event")) {
+                return new eventCommand(command);
+            } else if (command.startsWith("delete")) {
+                return new deleteCommand(command);
+            } else {
+                return new InvalidCommand(command);
```
SKCheongBrian: You can consider using a switch statement here 😄

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,74 @@
...
+        System.out.println("Got it. I've added this task:");
+        System.out.println(task);
+        System.out.println("Now you have " + taskList.size() + " tasks in the list.");
+    }
+    
+    public static void doneResponse(Task task) {
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println(task);
+        storage.writeToFile("./duke.txt", taskList);
+    }
```
SKCheongBrian: Good idea to write to file everytime a task is edited. This way the state of the tasks are not lost incase the app crashes.

src/main/java/Duke.java
```diff
@@ -1,10 +1,37 @@
...
+        ui.showWelcome();
+        ui.showLine();
+        boolean isExit = false;
+        while (!isExit) {
+            String command = ui.command();
+            Parser parser = new Parser(command);
+            if (parser.isExit()) {
+                break;
+            }
+            Command c = parser.parse();
```
SKCheongBrian: You can consider making `parse()` a static method, that way you would not have to create a new parser object everytime a command is given.

SKCheongBrian (commented):
Good job! Your code seems very well done.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,69 @@
...
+     * Returns the instance of the type of command entered by the user.
+     *
+     * @return Command object.
+     */
+    public Command parse() {
+        if (command.equals("list")) {
+            return new listCommand(command);
+        } else if (command.startsWith("done") && Character.isDigit(command.charAt(command.length() - 1)) && command.length() <= 8 
+                && !Character.isAlphabetic(command.charAt(command.length() - 2))  && Character.isDigit(command.charAt(5))) {
+            return new doneCommand(command);
```
Smanmos: This if block's condition could be simplified, with only the command type being checked in the main condition, and the other checks could be further done inside the body.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,69 @@
...
+            } else {
+                return new InvalidCommand(command);
+            }
+        }
+    }
+
+    /**
+     * Checks if the user types the command 'bye' or the user clicks enter without typing any command.
+     *
+     * @return True or False.
```
Smanmos: The return value could be further elaborated.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,69 @@
...
+import duke.command.InvalidCommand;
+import duke.command.deleteCommand;
+import duke.command.todoCommand;
+import duke.command.findCommand;
+
+/**
+ * Deals with making sense of the user command
+ */
+public class Parser {
+    private String command;
```
Smanmos, line 19: The string name `command` can be confusing, because a class called Command is in this project, but command is not a Command object. 

Maybe change it to `commandString`?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,107 @@
...
+import duke.task.Task;
+import duke.task.TaskList;
+
+/**
+ * Represents the user interface which deals with interactions with the user
+ */
+public class Ui {
+    Scanner sc;
+    private static TaskList taskList;
+    private static Storage storage;
```
Smanmos, line 14: Why are taskList and storage static?

src/main/java/duke/command/deadlineCommand.java
```diff
@@ -0,0 +1,71 @@
...
+import duke.Ui;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents the user command when the user enters a deadline.
+ */
+public class deadlineCommand extends Command {
```
Smanmos: Class name should be PascalCase: `DeadlineCommand`

src/main/java/duke/command/InvalidCommand.java
```diff
@@ -0,0 +1,45 @@
...
+        return "This is an invalid command";
+    }
+
+    /**
+     * Prints an error message when the user enters an invalid command.
+     *
+     * @param taskList TaskList that stores the tasks.
+     * @param storage Storage that deals with loading tasks from the file and saving tasks in the file.            
+     */
+    public void execute(TaskList taskList, Storage storage) {
```
Smanmos: Should have `@Override` annotation

src/main/java/duke/command/deleteCommand.java
```diff
@@ -0,0 +1,49 @@
...
+import duke.task.TaskList;
+
+import duke.Storage;
+
+import duke.Ui;
+
+/**
+ * Represents the user command when the user is deletes a task.
+ */
+public class deleteCommand extends Command {
```
Smanmos: Class name should be PascalCase: `DeleteCommand`

src/main/java/duke/command/doneCommand.java
```diff
@@ -0,0 +1,53 @@
...
+import duke.task.TaskList;
+
+import duke.Storage;
+
+import duke.Ui;
+
+/**
+ * Represents the user command when the user is done with a task.
+ */
+public class doneCommand extends Command {
```
Smanmos: Class name should be PascalCase: `DoneCommand`

src/main/java/duke/command/eventCommand.java
```diff
@@ -0,0 +1,71 @@
...
+import duke.Ui;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents the user command when the user enters an event.
+ */
+public class eventCommand extends Command {
```
Smanmos: Class name should be PascalCase: `EventCommand`

src/main/java/duke/command/findCommand.java
```diff
@@ -0,0 +1,50 @@
+package duke.command;
+
+import duke.task.TaskList;
+import duke.task.Task;
+
+import duke.Storage;
+
+import duke.Ui;
+
+public class findCommand extends Command {
```
Smanmos: Class name should be PascalCase: `FindCommand`

src/main/java/duke/command/findCommand.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+        return false;
+    }
+
+    public void execute(TaskList taskList, Storage storage) {
+        Ui.printMessage("Here are the matching tasks in your list:");
+        String[] parts = this.command.split(" ", 2);
+        String wordToFind = parts[1];
+        int count = 1;
```
Smanmos: count should start at 0 instead of 1, to make things more intuitive

src/main/java/duke/command/listCommand.java
```diff
@@ -0,0 +1,41 @@
+package duke.command;
+
+import duke.task.TaskList;
+
+import duke.Storage;
+
+/**
+ * Represents the user command when the user wants to view the task list.
+ */
+public class listCommand extends Command {
```
Smanmos: Class name should be PascalCase: `ListCommand`

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,28 @@
...
+
+import duke.Storage;
+
+import duke.task.TaskList;
+
+/**
+ * An abstract class that represents the various types of commands that can be entered by the user.
+ */
+public abstract class Command {
+    private String command;
```
Smanmos, line 11: Perhaps consider making `command` `protected` instead of `private`? It would make the code much cleaner since subclasses make use of this variable a lot

Smanmos (commented):
There were some violations of code (in particular class name capitalization). There was also a questionable design decision of writing to the file every time a new task was inserted or deleted, which can case performance issues. 

But overall, it's a good program design.

PR #484, opened 2021-08-27
[Saketh] iP

loose-bus-change:
# Duke
*Your own personal task recording system*

> “Your mind is for having ideas, not holding them.” – David Allen

* text-based
* easy to learn
* ~~FAST~~ **EXTREMELY FAST** to use

In order to access it:

1. download it from [here](https://github.com/loose-bus-change/ip/releases).
1. double-click it.
1. add your tasks.

And it is **FREE!**

Features:

 Managing tasks
 Managing deadlines (coming soon)
 Reminders (coming soon)

This is a sample task list that you can view if you type the `list` command: 
[D][ ] return book    by: 2019-12-02
[E][ ] meeting    at: 2015-10-15
[E][X] project meeting   at: 2015-10-15

Here is the main method of the application:

```java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+            int value = Integer.parseInt(input.replaceAll("[^0-9]", ""));
+            Task removedTask = t.get(value - 1);
+            t.delete(value - 1);
+            System.out.println("Noted. I've removed this task:");
+            System.out.println(removedTask.toString());
+            System.out.println("Now you have " + t.size() + " tasks in the list.");
+            s.appendListToFile(t);
+        } else if (input.startsWith("find")) {
+            String keyword = input.substring(5);
+            ArrayList<Task> output = new ArrayList<>();
```
gordonlzy: Perhaps the usage of plural form for output will work better?

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
+import duke.*;
+import java.io.File;
+
 public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private File
+            file;
```
gordonlzy: Any reason why you put file on the next line?

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
+import duke.*;
```
gordonlzy: Would it be better to list all the imported classes?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Method that adds a Deadline task onto TaskList
+     *
+     * @param input The string that contains the user input
+     * @param t The TaskList that contains the tasks to be added
+     * @param s The Storage that handles the reading and writing to a text file
+     * @param file The file that gets written to and read from
+     *
+     * @return void
```
gordonlzy: Is it necessary to add @return for void?

gordonlzy (commented):
LGTM, just a few nitpicks

src/main/java/Duke.java
wilburrito: I concur with @gordonlzy. I think this violates one of the [coding standards.](https://se-education.org/guides/conventions/java/intermediate.html#package-and-import-statements)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+
+public class Parser {
+
+    /**
+     * Method that adds a Deadline task onto TaskList
+     *
+     * @param input The string that contains the user input
+     * @param t The TaskList that contains the tasks to be added
```
wilburrito: perhaps the naming of some of the parameters could be more verbose; however this is just nitpicky, since your javadocs already explain what the parameters mean. Good work! 👍🏻

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Method that adds an Event onto TaskList
+     *
+     * @param description The string that contains the description of the task.
+     * @param date The date mentioned in the task initialization.
+     *
+     *
+     * @return void
```
wilburrito: I'm not sure if this is necessary for a `void` method.

wilburrito (approved):
LGTM, but just a few nits to fix, that's all.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    protected LocalDate by;
```
JasonC01: In my opinion, deadline seems to be a better name compared to by

JasonC01 (commented):
LGTM:)!! Great work overall!

src/main/java/duke/Parser.java
Smanmos: I think the variable name should be completely renamed. The queried tasks aren't an "output"; they are the result of a query, and are then reformatted into a text result that gets printed.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public String getMsg() {
+        return "";
+    }
+}
+
+class NullTaskError extends DukeException {
+    public NullTaskError() {
+        super("OOPS!!! The description of a todo cannot be empty.");
+    }
```
Smanmos, line 22: I think it would be better for type to be in the constructor instead of the getMsg method.

src/main/java/Duke.java
Smanmos: I agree as well. Imports should be one at a time.

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+import java.io.File;
+
 public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private File
+            file;
+    private Parser p;
```
Smanmos: Avoid one letter names; `parser` would be a better name

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,125 @@
...
+                String[] breakingDate = date.split("/");
+                String year = breakingDate[2];
+                String month = breakingDate[1];
+                String currentDate = breakingDate[0];
+                int i = Integer.parseInt(currentDate);
+                if (i < 10) {
+                    currentDate = "0" + currentDate;
+                }
+                String finalDateFormat = year + "-" + month + "-" + currentDate;
+                LocalDate date1 = LocalDate.parse(finalDateFormat);
```
Smanmos: Can be simplified using method LocalDate.of

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,38 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Ui {
+    Scanner scan;
+    private String str;
```
Smanmos: Could be more descriptive; what String is str holding?

Smanmos (changes requested):
Quite good overall, but many parts could be improved

PR #634, opened 2021-10-12
[Lim Yu Long] ip 

yulonglim:


src/main/java/duke/Controller.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * Saves tasklist to storage when any actions are carried out
+     */
+    public void save() {
+        try {
+            storage.save(tasks);
+        } catch (IOException e) {
+            e.printStackTrace();
```
damithc: Doesn't look like a good way to handle this exception. 🤔

src/main/java/duke/Controller.java
```diff
@@ -0,0 +1,215 @@
...
+                updateError.setText("Please enter a number in the list");
+            } else {
+                tasks.updateTask(index, updateDescription.getText(), updateDate.getValue());
+                this.save();
+                assert updateError.getText().equals("");
+            }
+        } catch (NumberFormatException e) {
+            updateError.setText("Please enter integer");
+        }
+    }
```
damithc, line 203: These three methods seem to be very similar. See if you can reduce duplication.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public Event(String description, String at) {
+        super(description);
+        if (at.contains("/")) {
+            date = LocalDate.parse(at.split("/")[2] + "-" + at.split("/")[1] + "-" +
+                    (Integer.parseInt(at.split("/")[0]) < 10 ? "0" + at.split("/")[0] :
+                            at.split("/")[0]));
+        } else {
+            date = LocalDate.parse(at);
+        }
```
damithc: There are some similar code elsewhere. Minimize code duplication.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Parses input given by storage during initialization of duke
+     *
+     * @param storageData string passed in after reading from storage file
+     * @return Arraylist of task to be passed into a task list during initialization of duke
+     */
+    public static ArrayList<Task> storageParse(String storageData) {
+        ArrayList<Task> tasks = new ArrayList<>();
+        if (!storageData.equals("")) {
```
damithc: You can use a guard clause to reduce nesting. e.g.,
```java
if (storageData.equals("")) {
    return tasks;
}
...
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+            for (String i :
+                    storageData.split("\n")) {
```
damithc: Is this line break needed?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+            return listContent;
+        }
+    }
+
+    @Override
+    public String toString() {
+        if (this.list.isEmpty()) {
+            return "There are no tasks in your list!\n";
+        } else {
```
damithc: No need to put this inside `else` as line 116 is a return statement.

damithc (commented):
Had a quick look (not a full review) and added some comments. Optional to fix. FYI only.

